PR #1252, opened 2026-05-06
fix(taskwarrior-plugin): document hyphen tag parsing quirk

### LINKED ISSUE #1237: feedback(taskwarrior-plugin): hyphens in tag names silently break parsing — document use of underscores
## Skill

`taskwarrior-plugin/skills/task-add/SKILL.md` (and adjacent task-* skills that touch tag syntax)

## Category

Enhancement

## Description

Hyphens in taskwarrior tag names silently break parsing. `task add "desc" +blocked-on-merge` is parsed as `+blocked` AND `-on-merge` (the `-` is taskwarrior's exclude-filter syntax, even mid-token), so the tag never lands and the literal `+blocked-on-merge` ends up appended to the description as plain text. Tags array stays null, urgency doesn't tick up.

The gotcha is **not** disclosed in the skill text. Quoting (`'+blocked-on-merge'`) does NOT help — this is a taskwarrior parser quirk, not a shell issue. Underscores (`+blocked_on_merge`) work fine.

## Evidence

In a multi-step session filing 9 SIMPL-Eval cluster-recovery tasks under `project:infrastructure.simpl-eval`, the first batch added all 9 tasks with tag args like `+blocked-on-merge +simpl-eval +cluster-recovery`. All 9 tasks ended up with `tags: null` and the literal tag tokens swallowed into description (e.g. `"description": "verify users-roles ... +blocked-on-merge +simpl-eval +cluster-recovery"`).

Bisection confirmed:
- `+foo +bar` (no hyphens) → tags applied correctly, urgency 1.8
- `+blocked-on-merge` (hyphen) → tag swallowed, urgency 1.0
- `'+blocked-on-merge'` (single-quoted) → still swallowed
- `+blocked_on_merge` (underscore) → tag applied correctly

Recovery required `task delete` + re-add with underscore tags.

## Suggested Action

Add a "Tag naming gotcha" callout to `task-add` (or wherever tags are first introduced). Suggested text:

> **Hyphens break tags.** `+blocked-on-merge` is parsed as `+blocked` AND `-on-merge` (exclude). Use underscores or camelCase: `+blocked_on_merge`, `+blockedOnMerge`. Quoting does not help.

The existing `parallel-safe-queries.md` rule already documents one taskwarrior parser quirk (`task list` exit-1 on empty); this is a natural sibling and could live there if a single rule is preferred.

laurigates:
## Summary

- Document the taskwarrior tag-name parser quirk where hyphens are interpreted as exclude-filter syntax mid-token, silently swallowing the tag (`+blocked-on-merge` → parsed as `+blocked` AND `-on-merge`, with the literal token appended to the description as plain text).
- Switch every example tag in `taskwarrior-plugin` skills and README from hyphen form to underscore form (`+pr_ready`, `+blocked_on_merge`, `+needs_review`, `+bulk_task`).
- Add a regression check in `scripts/plugin-compliance-check.sh` that flags any `+\w+-\w+` tag pattern in taskwarrior-plugin skills (except `task-add/SKILL.md`, which keeps the documented counter-examples) and asserts the **Tag naming gotcha** section stays in `task-add/SKILL.md`.

Closes #1237.

The first commit (`chore(configure-plugin)`) is unrelated bookkeeping — bumping the `configure-repo` driver's `reviewed:` date so the existing pre-commit `check-driver-freshness` hook stops failing on dependency drift from migration-patterns updates landed on 2026-05-04. No behaviour change. It's separated from the taskwarrior fix in its own commit so reviewers can see the scope split clearly.

## Test plan

- [x] Inject `+bad-tag-example` into a non-allowlisted taskwarrior skill — verified the new check fires with a clear `(issue #1237)` message
- [x] Restore the file — verified taskwarrior-plugin returns to all-green compliance
- [x] `bash scripts/plugin-compliance-check.sh` — all plugins green
- [x] `shellcheck scripts/plugin-compliance-check.sh` — clean
- [x] Pre-commit hooks pass on commit (shellcheck, gitleaks, lint-context-commands, check-driver-freshness, audit-skill-descriptions)

## Notes for reviewer

- The "Tag naming gotcha" section in `taskwarrior-plugin/skills/task-add/SKILL.md` intentionally retains hyphenated forms as counter-examples; the regression check exempts that file specifically and separately asserts the section heading exists.
- No agent behaviour change for tasks already in the queue with hyphenated tags — those tasks have empty `tags` arrays and the agent will treat them as untagged. Recovery (per the gotcha section) is `task delete` + re-add.


laurigates:
Superseded by #1248 (most comprehensive: linter script + pre-commit + CI workflow). Closes #1237 — see #1248.

Issue #78, opened 2026-01-15
Claude Code Changelog Review: 0.0.0 → 2.1.7


laurigates:
## Summary

| Category | Count |
|----------|-------|
| High-impact changes | 8 |
| Medium-impact changes | 12 |
| Plugins potentially affected | 6 |

**Review Date**: 2026-01-14
**Source**: [Claude Code Changelog](https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md)

---

### 1. hooks-plugin - Hook System Evolution

**Changes:**
- **v2.1.0**: Hooks now supported in skill/command frontmatter
- **v2.0.43**: `SubagentStart` hook event added with `agent_id` and `agent_transcript_path` fields
- **v2.0.30**: `SessionEnd` hooks introduced; prompt-based stop hooks added
- **v2.0.10**: PreToolUse hooks can now modify tool inputs via `updatedInput`

**Suggested Actions:**
- [ ] Update `SKILL.md` to document new hook events (SubagentStart, SessionEnd)
- [ ] Add examples for input modification in PreToolUse hooks
- [ ] Document frontmatter hook support for skills/commands
- [ ] Create example hooks for new events

---

### 2. agent-patterns-plugin - Agent Capabilities

**Changes:**
- **v2.1.0**: `context: fork` in skill frontmatter enables forked sub-agent context
- **v2.1.0**: `agent` field in skills to specify agent type
- **v2.0.30**: `disallowedTools` field for custom agent definitions
- **v2.0.28**: Plan Mode uses new Plan subagent; subagents can be dynamically resumed

**Suggested Actions:**
- [ ] Add documentation for `context: fork` pattern
- [ ] Document `disallowedTools` for agent definitions
- [ ] Update multi-agent workflow patterns for subagent resume capability
- [ ] Add examples using the new `agent` field in skills

---

### 3. All Plugins - Skills/Commands Merged

**Changes:**
- **v2.1.3**: Skills and slash commands merged, simplifying mental model
- **v2.1.0**: Automatic skill hot-reload from skill directories
- **v2.1.0**: `SlashCommand` tool - Claude can invoke custom slash commands
- **v2.0.28**: Slash commands can invoke skills

**Suggested Actions:**
- [ ] Review all skill/command descriptions for consistency with merged model
- [ ] Document hot-reload behavior in skill development guide
- [ ] Consider adding SlashCommand tool invocation examples

---

### 4. configure-plugin - Permission System Updates

**Changes:**
[...]

PR #1302, opened 2026-05-11
ci(changelog-review): refactor to triage-only + skip-if-exists

laurigates:
## Summary

Restructures the changelog-review workflow so a 62-version gap no longer blows the turn budget. The previous design tried to do triage AND rule edits in a single Claude invocation; with the 2.1.76 → 2.1.138 gap that pattern failed in [run 25651584541](https://github.com/laurigates/claude-plugins/actions/runs/25651584541) at turn 51/50 after 2h 7min. (Recovery of that run's actual rule-edit work is in #1301.)

## Why it failed (data)

The SDK reports `total_cost_usd` (notional, not billed under Max 20x — but still a clean proxy for token volume since it's computed from token counts). Per-call token counts aren't exposed at the result level, only the aggregate.

Comparison against this workflow's recent weekly runs:

| Run | Date | Turns | Duration | Notional $ | Notes |
|---|---|---|---|---|---|
| 2026-05-04 | success | 41 | 8m 31s | $2.01 | Normal weekly run |
| 2026-04-27 | success | 4 | 47m | $57.45 | Big-context first-hit; 4 huge turns |
| 2026-04-20 | max_turns | 41 | 36m | $39.18 | Same failure mode, smaller gap |
| 2026-04-15 | success | 20 | 1m 30s | $0.40 | No-op week |
| 2026-04-06 | success | 9 | 11m | $6.18 | Modest gap |
| **2026-05-11** | **max_turns** | **51** | **2h 7m** | **$63.69** | The recovered run |

For reference, a typical `claude-code-review.yml` PR review run is $0.50–$1.00 with 15-20 turns. The changelog-review workflow on a fresh gap is **30–100× that**, dominated by context size on first hit (4 turns / $57 in one run) rather than turn count alone.

## What changed

**Triage-only phase (this workflow):**
- Read the excerpt + the JSON tracking file
- Open ONE tracking issue summarizing relevant changes grouped by candidate rule file
- Ratchet `.claude-code-version-check.json`
- Open one JSON-only PR
- **No rule edits.** Bounded to `--max-turns 25` with sonnet.

**Apply phase (delegated to existing `claude.yml`):**
- User mentions `@claude please update <rule-file> for issue #N` on the tracking issue
- Each apply is small-scope with its own turn budget — single rule file at a time

**Skip-if-exists check:**
- Before invoking Claude, query `gh issue list --label changelog-review --state open` for a title matching the current range. If found, skip. `workflow_dispatch` + `force_full_review=true` overrides.
- Makes the workflow safe to re-run and idempotent across cron triggers.

**Expanded `additional_permissions`:**
- Adds `Bash(gh api *)`, `Bash(gh label *)`, `Bash(jq *)`, `Bash(grep *)`, `Bash(awk *)`, `Bash(wc *)`, `Bash(cat *)`, `Bash(date *)`, etc.
- The failed run logged **201 permission denials in 51 turns** — ~4 wasted turns per usable turn just retrying denied commands. These additions cover the patterns the agent reached for.

**Prompt hardening:**
- Strict step-by-step triage flow with explicit non-goals ("DO NOT edit rule files")
- Final sanity check: confirm exactly one commit, exactly one file changed, no rule-file edits, before finishing the turn
- Explicit warning about the previous run's self-misdiagnosis loop ("if you think a branch is empty, re-check with `git diff`/`git log` before destructive recovery")
- Excerpt truncated to 200KB if larger — keeps the prompt cache predictable

## Why this design over alternatives

| Option | Why not |
|---|---|
| Matrix: one job per version | 62 jobs is noisy; most versions are pure bug fixes (e.g. 2.1.137 is one bullet) |
| Sub-agents / agent team within one run | Doesn't multiply turn budget — `Task`/`SendMessage` count as turns from the parent loop. Only separate jobs get independent budgets. |
| Keep monolithic single-PR design | Will fail again on the next gap |
| Iterative-with-checkpoint | Possible, but the issue/apply split matches the historical pattern (#78, #786, #854, #909, #949) and lets human review gate each apply |

## Test plan

- [ ] Validate YAML parse + actionlint
- [ ] Dry-run via `workflow_dispatch` with `force_full_review=true` AFTER #1301 merges (tracked version will be 2.1.138, latest will be whatever's current — small or no gap, ideal for dry-running the new shape)
- [ ] Verify the triage issue is created with grouped per-rule-file sections
- [ ] Verify the JSON-only PR contains exactly one file change
- [ ] Run a second dispatch (without force) and confirm skip-if-exists fires

## Follow-up (not blocking)

- Consider extracting the workflow into `laurigates/.github` as `reusable-changelog-review.yml` once the new shape is proven — per `.claude/rules/ci-cd-workflows.md` we prefer reusable workflows for cross-repo automation, though this one is plugin-collection specific so reuse value is currently low.
- The `show_full_output: false` setting in `anthropics/claude-code-action` keeps the per-call token counts hidden. If we ever want to track token-level efficiency over time, we could either flip that on for the triage job specifically, or scrape the OpenTelemetry log events that the SDK emits (richer detail than the summary result).

🤖 Drafted with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/changelog-review.yml
+++ b/.github/workflows/changelog-review.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
     inputs:
       force_full_review:
-        description: 'Force full changelog review'
+        description: 'Force triage even if a tracking issue for the current range already exists'
         required: false
         default: 'false'
         type: boolean
@@ -27,6 +27,8 @@ jobs:
       previous_version: ${{ steps.read_tracked.outputs.tracked_version }}
       changes_summary: ${{ steps.analyze.outputs.summary }}
       candidate_rules: ${{ steps.analyze.outputs.candidates }}
+      existing_issue: ${{ steps.existing.outputs.issue_number }}
+      should_skip: ${{ steps.existing.outputs.should_skip }}
     steps:
       - uses: actions/checkout@v6
         with:
@@ -70,7 +72,7 @@ jobs:
           FORCE="${{ inputs.force_full_review }}"
 
           if [ "$FORCE" = "true" ]; then
-            echo "Forced full review requested"
+            echo "Forced full triage requested"
             echo "has_updates=true" >> $GITHUB_OUTPUT
             echo "latest_version=$LATEST" >> $GITHUB_OUTPUT
             exit 0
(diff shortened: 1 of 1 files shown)


laurigates:
@claude let's create the follow-up issues

PR #1308, opened 2026-05-11
feat(hooks-plugin): add bash-antipatterns-teach PostToolUse hook (opt-in)

laurigates:
## Summary

- Add `bash-antipatterns-teach.sh`: a PostToolUse Bash hook that augments tool output via `hookSpecificOutput.updatedToolOutput` (Claude Code 2.1.121+) with concrete tool-replacement hints for soft-teach antipatterns (`cat`/`head`/`tail`, `find -name`, `grep`/`rg`, `ls *.glob`)
- Wire the hook into `hooks-plugin/.claude-plugin/plugin.json` as a `PostToolUse` `Bash` matcher with `timeout: 5000`
- Gate execution behind `CLAUDE_HOOKS_ENABLE_BASH_ANTIPATTERNS_TEACH=1` (default disabled; matches `event-logger.sh` opt-in convention)
- Existing PreToolUse `bash-antipatterns.sh` is unchanged

## Motivation

W20 friction analysis (`~/.claude/rules/friction/2026-W20-frictions.md`) measured a **21% same-session repeat-block rate** on `grep`/`rg` PreToolUse blocks — about 2× the rate for `git &&` (8%) or `find -name` (12%). The hook teaches less effectively because the agent only sees the corrective message, never the would-have-been-result. Letting the command run and prepending the hint to the tool result gives the agent both the data and the lesson in one response.

Hypothesis: same-session repeat-block rate on `grep`/`rg`/`find -name`/`cat`/`head`/`tail`/`ls *.glob` drops from ~21% to the established ~10% floor. W21 friction pass (2026-05-18) measures.

## Test plan

- [x] `bash -n` syntax-check on both scripts
- [x] `jq empty` validates `plugin.json`
- [x] New `test-bash-antipatterns-teach.sh`: 17 passed, 0 failed (covers all match patterns, exemptions, and the env-var guard)
- [x] Existing `test-bash-antipatterns.sh`: 25 passed, 0 failed (no regression)
- [x] Hook follows `event-logger.sh`-style opt-in convention
- [x] Phase 1 caveat documented: teach hook only fires when PreToolUse didn't block, so observing teach-mode in phase 1 requires also overriding the soft-teach PreToolUse blocks locally

## Follow-ups (not in this PR)

- W21 friction pass measures repeat-block rate on opted-in sessions
- Phase 2 (W22+) decides whether to flip the env-var default and strip soft-teach blocks from `bash-antipatterns.sh`
- Separate issue: invoke `hooks-plugin/hooks/test-*.sh` from CI

See `hooks-plugin/docs/teach-mode-experiment.md` for full hypothesis, rollout plan, and success criterion.


laurigates:
Let's address the recommendations in the comments. Evaluate which pieces of criticism and recommendations are substantive and valid. Act accordingly and refute invalid points.
@claude

Issue #1377, opened 2026-05-22
hook: upgrade bash-antipatterns block messages to substitution format


laurigates:
## Evidence

Source: [`~/.claude/rules/friction/2026-W21-frictions.md`](https://github.com/laurigates/claude-plugins/blob/main/.claude/rules/friction/) — user-global rules, week ending 2026-05-18 (133 transcripts parsed, 94 sessions touched, 420 friction events).

Two W20 watch-list triggers fired this week. The first one — `gh --json` unknown-field events — dropped cleanly from 10 → 0 after the W20 rule landed (PR #1300), confirming that **the substitution-format rule is what landed**. The second one — `grep`/`rg` repeat-block rate — went the wrong direction:

| Pattern | W20 events/sessions | W21 events/sessions | Per-session attempt rate | Same-session repeat-block rate |
|---|---|---|---|---|
| `find` vs `Glob` | 29 / 25 | 37 / 25 | 27% (up from 17%) | 24% |
| `grep` / `rg` vs `Grep` | 41 / 33 | 31 / 24 | 26% (flat) | **29% (up from 21%)** |
| `cat` / `head` / `tail` vs `Read` | 29 / 24 | 11 / 9 | 17% | 22% |
| `gh --json` unknown field (W20 baseline) | 10 / 10 | **0 / 0** | n/a | n/a — rule WORKED |

Combined, these three bash-antipattern patterns account for **67% of all W21 bash antipattern events (94 of 140)**. All three sit at roughly the same 26-28% per-session attempt rate as `git &&` chains, so the W21 finding is that this is a steady-state population prior, not workload mix.

The W20 watch-list pre-registered the trigger condition:

> "If the rule reduces the 21% repeat-block rate to the ~10% range seen for `git &&`/`find`, the rule is doing its job. If it stays at 20%+, the issue is not lack of documentation — it's the agent's deeply ingrained bash-search prior, and the next step is to soft-block (block + suggest alternative) instead of reminder."

But the hook is **already** a soft-block (`exit 2`). The W19 friction analysis established that. So the actual next lever is **the block message itself** — the existing text is advisory ("Use the Grep tool instead"), not substitution.

## Proposal

[...]

Issue #1411, opened 2026-05-25
hook: extend substitution-format BLOCKED messages to git-chain, ls-glob, long-pipelines, TaskOutput


laurigates:
## Evidence

Source: [`~/.claude/rules/friction/2026-W22-frictions.md`](https://github.com/laurigates/claude-plugins/blob/main/.claude/rules/friction/) — user-global rules, week ending 2026-05-25 (W22 friction-learner pass).

[Issue #1377](https://github.com/laurigates/claude-plugins/issues/1377) was resolved by [PR #1378](https://github.com/laurigates/claude-plugins/pull/1378) (merged 2026-05-23, `hooks-plugin` 2.4.0) which upgraded the `bash-antipatterns` hook to use substitution-format BLOCKED messages for `find`, `grep`/`rg`, and `cat`/`head`/`tail`. Only ~2 of 7 W22 days had the new messages live, so the W22 sample is too small to measure the format's effect — but the early numbers are encouraging:

| Pattern | W21 per-session attempt rate | W22 per-session attempt rate | W22 BLOCKED-format same-session repeat-block rate (small sample) |
|---|---|---|---|
| `find` vs `Glob` | 27% | **19%** (down) | 16.7% (n=6) |
| `grep` / `rg` vs `Grep` | 26% | 24% (flat) | 28.6% (n=7) |
| `cat` / `head` / `tail` vs `Read` | 17% | 17% (flat) | 14.3% (n=7) |
| `git &&` chains | 28% | 27% (flat — **not** upgraded in #1378) | n/a |

BLOCKED-format samples are 6-7 sessions each — not enough to conclude the format works or doesn't. W23 (2026-06-01) will be the first pass with a full 7 days of post-upgrade data (~3× the sample) and will be the deciding evidence.

## Proposal

If W23 confirms the substitution-format hypothesis (BLOCKED-format same-session repeat-block rates drop below 15% on `find` / `grep` / `cat`-`head`-`tail`), extend the same upgrade to the remaining REMINDER-format patterns in [`hooks-plugin/hooks/bash-antipatterns.sh`](https://github.com/laurigates/claude-plugins/blob/main/hooks-plugin/hooks/bash-antipatterns.sh):

| Pattern | W22 events / sessions | W22 per-session rate | Why it deserves the upgrade |
|---|---|---|---|
[...]

PR #1412, opened 2026-05-25
fix(feedback-plugin): classify new BLOCKED-format hook events into specific signatures

laurigates:
## Summary

PR #1378 upgraded `hooks-plugin/hooks/bash-antipatterns.sh` so that when it blocks an antipattern, the message begins with `BLOCKED:` and includes a substitution example, replacing the prior `REMINDER:` prefix. The friction parser's `canonical_signature()` only knew the legacy needle table (`branch-protection`, `pr metadata`, etc.), so every BLOCKED-format event fell through to the generic `hook:unclassified` bucket — breaking trend analysis on exactly the patterns the W21 watch-list pre-registered.

## Impact (W22 evidence window)

Re-running `friction_parse.py --since 7d` against the same transcripts:

| Signature | Before | After |
|---|---|---|
| `hook:unclassified` | **29** | **2** (residual: 1 `rg` IO error + 1 `tsc` typecheck output that incidentally matches the `Exit code 2` regex — pre-existing parser issue, out of scope) |
| `hook:bash-antipatterns:grep-rg` | — | 13 |
| `hook:bash-antipatterns:find` | — | 9 |
| `hook:bash-antipatterns:cat-head-tail` | — | 8 |
| `hook:secret-protection` | — | 1 |

(Counts differ slightly from the W22 findings file's 26-event tally because the rolling 7-day window has rotated since W22 was captured on 2026-05-25.)

## What the parser fix does

In `canonical_signature()`:

- Sub-classify any `hook_block` event whose evidence text contains `bash-antipatterns` by matching the exact BLOCKED-format prose emitted by `hooks-plugin/hooks/bash-antipatterns.sh`:
  - `BLOCKED: 'grep …` / `BLOCKED: 'rg …` → `hook:bash-antipatterns:grep-rg`
  - `BLOCKED: 'find …` → `hook:bash-antipatterns:find`
  - `BLOCKED: 'cat …` / `'head …` / `'tail …` → `hook:bash-antipatterns:cat-head-tail`
- Forward-compatible sub-signatures for the W22 watch-list candidates that may receive the same substitution-format upgrade in a follow-up PR: `git-chain`, `ls-glob`, `long-pipeline`, `taskoutput`.
- Fallback `hook:bash-antipatterns:other` so unrecognized bash-antipatterns text doesn't silently disappear if a future hook block uses prose the parser hasn't seen.
- Classify other hook scripts (e.g. `secret-protection`) by script-name substring, ahead of the legacy needle table, to avoid future false-positives.

## Scope

Parser classification only. The hook script (`bash-antipatterns.sh`) is unchanged — it is correct. The unrelated `error:bash` false-positives on `tsc` / `rg IO error` (the 2 residual unclassified events) are a separate pre-existing issue and left alone.

## Test plan

- [x] Run `python3 feedback-plugin/scripts/tests/test_friction_parse.py` — 4/4 pass (new `test_bash_antipatterns_blocked_format_classifies_by_pattern` covers grep-rg, find, cat-head-tail, `:other` fallback, and `hook:secret-protection`).
- [x] Run sibling test suites (`test_friction_cluster.py`, `test_friction_open_prs.py`) — 4/4 and 6/6 pass; no regression.
- [x] Re-run `friction_parse.py --since 7d` against live transcripts; verify `hook:unclassified` drops from 29 → 2 and 30 events redistribute to the new sub-signatures.

## Context

- Friction-learner W22 findings: `~/.claude/rules/friction/2026-W22-frictions.md` "Proposed changes" #1 (user-global path; not in this repo).
- Refs #1377 (substitution-format hook upgrade design issue).
- Builds on PR #1378 (the hook upgrade that introduced the new BLOCKED format).

Lands ahead of the W23 friction pass on **2026-06-01** so the trend analysis on the substitution-format upgrade hypothesis stays accurate.


laurigates:
## W23 verification — parser fix working as designed

The W23 friction pass (week ending 2026-06-01) is the first full window using this PR's `canonical_signature` extension. Verification:

| Signature | W22 (pre-fix) | W23 (post-fix) |
|---|---|---|
| `hook:unclassified` | 29 | 8 (mostly `hook:secret-protection` catchall, not bash-antipatterns) |
| `hook:bash-antipatterns:grep-rg` | did not exist | 140 |
| `hook:bash-antipatterns:find` | did not exist | 95 |
| `hook:bash-antipatterns:cat-head-tail` | did not exist | 62 |
| `hook:bash-antipatterns:git-chain` | did not exist | 0 |
| `hook:bash-antipatterns:other` | did not exist | 1 |

This PR's predicted impact ("drops `hook:unclassified` from 29 → 2 on W22 transcripts") was achieved on the W23 window — the residual 8 are different hook signatures, not mis-bucketed bash-antipattern events.

The `git-chain` bucket at 0 is correct: `bash-antipatterns.sh` still uses REMINDER (not BLOCKED) format for git-chaining, so the `BLOCKED: 'git ...'` regex correctly returns no matches. Those events are picked up under `error:bash` instead via the REMINDER-text sub-classifier.

The decisive analysis enabled by this fix is in `~/.claude/rules/friction/2026-W23-frictions.md` — the verdict on #1411 (close-and-successor branch) depends entirely on having clean repeat-block-rate numbers per pattern, which would have been impossible without this PR.

Thanks for landing this quickly.

PR #1444, opened 2026-05-29
fix(git-plugin): make resolving review threads the default after reply

laurigates:
## Summary

Strengthens `git-pr-feedback` so a thread gets marked **Resolve conversation** in the same turn as the reply whenever the agent has either pushed a fix or refuted a suggestion with reasoning. Previously the resolve step was framed as a conditional check ("resolve when all of the following hold") which left room for replies to land without the matching resolve call — and a PR with every concern addressed could still show as having unresolved threads in the GitHub UI.

## Changes

- **`git-plugin/skills/git-pr-feedback/SKILL.md`** — Step 6 is now framed as "reply, then resolve by default" with an explicit list of completion shapes (pushed fix, answered question, written refutation, deferred-to-issue) and a small, enumerated set of exceptions that legitimately leave a thread open (follow-up question to reviewer, partial fix, reviewer asked to keep open, no SHA yet, third-party PR without user approval).
- **`git-plugin/skills/git-pr-feedback/REFERENCE.md`** — "Resolution Criteria" mirrors the same default-resolve framing, and the multi-PR subagent prompt now instructs subagents to default to `"resolve": true` for every drafted reply.
- **Step 1A orchestrator rule** — clarified that a `resolve: true` paired with a successful reply is a mandatory call, not a conditional one.

Refutation now explicitly counts as a thread-completion shape: a written reasoning is what completes the conversation, as opposed to silent disagreement (which still leaves the thread for the reviewer).

## Test plan

- [ ] On a real PR, invoke `/git:pr-feedback <pr>` with `--commit --push` against a thread that contains an accepted suggestion; verify the resolve call lands in the same turn as the reply.
- [ ] On a PR with a nitpick being declined with reasoning, verify the thread is resolved (not left open) — refutation path.
- [ ] On a PR where the agent's reply asks the reviewer a follow-up question, verify the thread is **not** resolved (exception path).
- [ ] On `/git:pr-feedback --all`, verify subagents return `"resolve": true` by default and the orchestrator resolves each thread paired with a posted reply.
- [ ] `bash scripts/plugin-compliance-check.sh git-plugin` passes (SKILL.md still under 500-line ceiling, frontmatter intact).

---
_Generated by [Claude Code](https://claude.ai/code/session_01N1FKB9gKWLFmJQKDfK6xbk)_


laurigates:

Issue #1319, opened 2026-05-14
bug(agent-tooling): worktree-isolated agent briefly leaked a new file into the parent checkout


laurigates:
## Symptom

During a multi-agent orchestration session on 2026-05-14, a subagent dispatched with `Agent(isolation: "worktree")` to create a brand-new file in its worktree (`health-plugin/skills/health-check/scripts/check-runtime.sh`) briefly caused that **same untracked file** to appear in the **parent checkout** at the same relative path. The parent checkout was on a different branch (`claude/multi-agent-issue-triage-8kHmn`) and never invoked any tool that would write to that path.

This triggered the stop-hook's "untracked files in the repository" warning in the parent session and forced an investigation before the agent itself had committed.

## Evidence

Parent checkout, while the agent was still running (no completion notification yet):

```
$ git status --short
?? health-plugin/skills/health-check/scripts/check-runtime.sh

$ ls -la health-plugin/skills/health-check/scripts/check-runtime.sh
-rwxr-xr-x 1 root root 9949 May 14 11:01 ...

$ git worktree list
/home/user/claude-plugins                                            6dbe2fc [claude/multi-agent-issue-triage-8kHmn]
/home/user/claude-plugins/.claude/worktrees/agent-a200584a8c0b5a92a  6dbe2fc [worktree-agent-a200584a8c0b5a92a] locked
...
$ ls -la /home/user/claude-plugins/.claude/worktrees/agent-a200584a8c0b5a92a/health-plugin/skills/health-check/scripts/
check-hooks.sh  check-mcp.sh  check-plugins.sh  check-settings.sh
# Note: no check-runtime.sh in the worktree at this moment
$ git -C /home/user/claude-plugins/.claude/worktrees/agent-a200584a8c0b5a92a status --short
(empty)
```

So at the snapshot time:

- The worktree's HEAD was `6dbe2fc` (no commits yet).
- The worktree's filesystem did **not** contain `check-runtime.sh`.
- The parent checkout's filesystem **did** contain `check-runtime.sh` as an untracked file.

A few minutes later the agent committed (`0d43a15`) and pushed; the file is correctly present in PR #1318 (the agent's pushed branch). At that point, the orphan in the parent checkout had vanished — `git status --short` in the parent was clean again.

## Why it matters

[...]

Issue #1491, opened 2026-06-03
feedback(agent-patterns-plugin): salvage worktree WIP when an isolated agent exits before StructuredOutput


laurigates:
## Skill

`agent-patterns-plugin/skills/parallel-agent-dispatch/SKILL.md` (covers "silent exits")

## Category

Enhancement

## Description

A schema-bound, worktree-isolated workflow agent that is cut off (e.g. by a rate limit) *after* implementing its change but *before* emitting its final StructuredOutput call is reported as **failed** — yet the work is sitting fully intact as **uncommitted WIP in the agent's worktree**. Without an explicit salvage step that work is invisible and easily discarded: the orchestrator's result array shows nothing, and only the persisted worktree (which survives because it *did* change) hints that anything happened.

## Evidence

This session, run 2: 5/7 agents opened PRs; 2 (theme-management #1066 and #1143) were marked *"subagent completed without calling StructuredOutput (after 2 in-conversation nudges)"*. Inspecting their worktrees (`git -C <worktree> status --short`) showed **complete, correct implementations** as uncommitted changes (new files + edits). I salvaged both manually — fixed 2 trivial typecheck errors in one (`as TransientUpstreamError` → `as Error`), ran lint/typecheck/tests, then committed/pushed/PR'd them (#1167, #1168). No re-run needed; the work was already done.

## Suggested Action

- In the dispatch contract, have worktree-isolated agents `git add -A && git commit` **WIP before they would otherwise terminate**, so partial work is always captured on a branch even if the structured result is lost.
- Document a **salvage recipe** for the orchestrator: when an agent is reported failed, check its worktree for uncommitted WIP *before* re-dispatching, to avoid redoing completed work.
- Optionally: distinguish "agent errored with empty worktree" from "agent produced changes but didn't return structured output" in the failure summary.

PR #1520, opened 2026-06-05
fix(agent-patterns): guard worktree agents against cwd-reset leaking git writes into main repo (#1480)

### LINKED ISSUE #1480: feedback(agent-patterns): isolation:worktree subagent leaked git writes into main repo (cwd reset)
## Skill

`agent-patterns-plugin/skills/parallel-agent-dispatch/SKILL.md` (primary)
Also relevant: `git-plugin` `git-ops` agent definition (the agent involved here).

## Category

Bug

## Description

A subagent dispatched via the `Agent` tool with `isolation: "worktree"` performed its **git write operations in the main repository checkout instead of its assigned worktree**, silently corrupting the user's main working tree.

The agent (`git-plugin:git-ops`, `run_in_background: true`, `isolation: "worktree"`) was asked to rebase a stacked 4-PR branch chain. The harness *did* create a worktree (the completion notification reported `worktreePath: …/.claude/worktrees/agent-a8456e20504b9b0fa`, `worktreeBranch: worktree-agent-a8456e20504b9b0fa`). But the agent's `git fetch` / `git checkout -B <branch>` / `git rebase --autostash` ran against the **main checkout**, not the worktree.

The most likely mechanism: agent threads have their bash cwd reset between calls (documented behavior), and the reset landed on the session's primary cwd (the main repo root) rather than the agent's worktree. Every bare `git …` the agent ran therefore mutated `main`.

This defeats the entire purpose of `isolation: "worktree"` for any git-write-heavy agent, and does so **invisibly** — the orchestrator believes the work is sandboxed.

## Evidence

Observed end state in the main repo after the agent "completed":

- Main repo HEAD left on a stray local branch `funding-topic-summary` (`a98cf19`) — the agent's own parallel rebase — instead of `main`.
- The user's **uncommitted** pre-session edits (`.claude/rules/development.md`, `.mcp.json`) were swept into a `git rebase --autostash`, which then **pop-conflicted** on unrelated files (`src/app/api/admin/ai/batch-auto-link/route.ts`, `src/lib/api/fetch-utils.ts`), leaving conflict markers in the main working tree.
- The Stop hook caught both symptoms: `git-stash-reminder.sh` flagged an orphaned `autostash`, and the task-completeness check flagged unresolved merge-conflict markers.

Recovery required: restoring 5 leaked files to HEAD, confirming the user's two genuinely-uncommitted files were byte-identical in the stash before dropping it, switching the main repo back to `main`, and deleting 4 stray local branches. No data was ultimately lost, but only because the real work was already pushed to `origin`; the user's uncommitted changes were one bad conflict resolution away from loss.

The agent's own returned message was also truncated mid-conflict-resolution (it never produced its final report and pushed nothing), so the orchestrator could not tell from the result that anything had gone wrong — the damage was only discoverable by inspecting `git`/`gh` state directly.

## Suggested Action

Harden the `parallel-agent-dispatch` contract (and the prompts of git-write agents such as `git-ops`) to be resilient to cwd-reset under worktree isolation:

1. **Never assume `cwd == worktree`.** Agents doing git *writes* under `isolation: worktree` should either (a) `cd "$WORKTREE"` and verify with `pwd` at the top of **every** bash call, or (b) use explicit `git -C "$WORKTREE" …` for all git invocations.
2. **Pin the worktree path.** The dispatch contract should instruct the agent to capture its worktree root once (e.g. `git rev-parse --show-toplevel`) and reference it absolutely thereafter, rather than relying on the persisted cwd.
3. **Forbid bare branch-switching/autostash in agents.** A git-write agent should never run `git checkout -B`/`git rebase --autostash` without first confirming it is inside its isolated worktree — a guard line in the dispatch checklist would catch this.
4. **Add a post-run integrity check to the contract**: after a git-write agent returns, the orchestrator should verify the main repo's branch and `git status` are unchanged (the contract's "silent exits" section could be extended to cover "silent main-repo mutation").

Note: there is almost certainly an underlying Claude Code harness component to this (the worktree cwd not being the agent's default cwd). That side was submitted separately via the built-in `/feedback`. This issue tracks the plugin-side defensive mitigation in the dispatch guidance.

laurigates:
## What

Hardens the parallel-agent-dispatch contract and the `git-ops` agent prompt against a worktree-isolation foot-gun: a git-write agent under `isolation: "worktree"` whose bare `git` commands silently ran against the **main checkout** instead of its worktree.

## Why

Agent threads have their bash cwd reset between calls (documented behaviour). For a git-write agent, the reset can land on the session's primary cwd (the main repo root) rather than the agent's worktree, so every `git fetch` / `git checkout -B` / `git rebase --autostash` mutates `main` — sweeping the user's uncommitted edits into an autostash pop-conflict and leaving the main repo on a stray branch. The orchestrator believes the work is sandboxed, so the corruption is invisible. This is **distinct** from the #1319 *transient* leak (already documented in `agent-coworker-detection.md`).

## Changes

- **`parallel-agent-dispatch/SKILL.md`** Pillar 1: new "cwd-reset leaking git writes into the main repo" subsection — pin the worktree root once (`git rev-parse --show-toplevel`), prefix every git call with `git -C "$WORKTREE"`, forbid bare branch-switch/autostash, and run a post-run main-repo integrity check.
- **`git-plugin/agents/git-ops.md`**: new "Important: Worktree Isolation and cwd Reset" section with the same guardrail.
- **`scripts/check-agent-failure-contract.sh`**: regression assertion that the SKILL.md carries the `#1480` reference and the literal `git -C "$WORKTREE"` idiom. Verified it fails before the fix and passes after.
- **`.claude/rules/regression-testing.md`**: documents the #1480 regression.

Closes #1480


laurigates:
I attempted to resolve the merge conflicts with main, but the system requires special permissions to edit the sensitive files involved. Both conflicts involve adding new regression checks that should both be kept. To resolve: merge main into this branch and keep both sections from each side of the conflicts by removing the conflict markers but keeping all content. The two issues (#1480 and #1491) are distinct and complementary regressions.

Issue #1591, opened 2026-06-11
bug(hooks-plugin): bash-antipatterns recommends the deprecated TaskOutput tool for task output files


laurigates:
## What

`hooks-plugin/hooks/bash-antipatterns.sh:173-177` blocks Bash reads of background-task output files with:

> REMINDER: Use the TaskOutput tool instead of Bash commands to read task output. The TaskOutput tool is designed for checking on background tasks - use it with the task_id parameter. Example: TaskOutput with task_id and block=false for non-blocking status checks.

TaskOutput is now **deprecated** — its own tool description says: background tasks return their output file path in the tool result / task-notification, and for bash tasks you should "prefer using the Read tool on that output file path". For local_agent tasks it explicitly warns the output file is a transcript symlink that will overflow context.

## Why

Hit 3× in one session (2026-06-11): legitimate `jq`/`grep`/`python3` extraction over large workflow/bash output JSON files was blocked, the agent then loaded the deprecated TaskOutput schema via ToolSearch only to be told to use Read instead — a wasted round-trip each time. Worse, for **large structured output files** (one was ~110 KB of JSON), plain Read is the wrong tool too: a `jq`/`python` extraction to a compact summary file is the context-efficient path, and the hook currently forbids exactly that.

Bonus reproduction: while filing **this very issue**, the hook blocked the `gh issue create` command because the issue body text matched its task-output patterns — the matcher fires on string content inside heredocs/quoted args, not just on actual read commands. (Worked around with `--body-file`.)

## How

In the task-output branch of `bash-antipatterns.sh`:

1. Replace the TaskOutput recommendation with Read (matching the current tool guidance), e.g. "Use the Read tool on the output file path from the task-notification."
2. Consider allowing extraction pipelines (`jq`, `python3`, pipe-to-grep) over task output files — for large JSON results, extraction-to-summary beats Read'ing the whole file into context. At minimum mention it as the escape hatch for files too large to Read.
3. Tighten the matcher so quoted/heredoc string content (e.g. an issue body that merely *mentions* task output paths) doesn't trigger the block.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1693, opened 2026-06-18
fix(hooks-plugin): stop bash-antipatterns blocking git chains quoted in gh bodies

### LINKED ISSUE #1587: bash-antipatterns: false positives — heredoc-to-gh chains blocked with git-commit reminder; patterns matched inside quoted strings
## What

`hooks/bash-antipatterns.sh` **blocked** (PreToolUse error, command never ran) a legitimate write-file-then-use-it pattern:

```sh
cat > /tmp/issue-body.md <<'EOF'
...multiline markdown body...
EOF
gh issue edit 2001 -R ForumViriumHelsinki/infrastructure --body-file /tmp/issue-body.md
```

The hook fired the **git-commit heredoc reminder**:

> REMINDER: Use HEREDOC directly in git commit:
> `git commit -m "$(cat <<'EOF' ... EOF)"`

## Why it's wrong

- No `git commit` anywhere in the command — the suggested replacement is irrelevant to `gh issue edit --body-file`.
- The hook's own `cat`-replacement messaging states "Pipelines (cat file | jq) and **heredocs (cat <<EOF) are still allowed**" — but this heredoc was blocked, not allowed.
- Writing a body file and passing `--body-file` is the *recommended* pattern for multi-line `gh` bodies (avoids quoting pitfalls); blocking it pushes toward worse alternatives (giant inline `--body` strings).

The agent's workaround was to use the Write tool for the temp file and run `gh` separately — workable, but two round-trips for something the hook's policy says is allowed.

## Expected

The heredoc-related rule should only trigger when the command actually contains `git commit` (i.e. suggest folding the heredoc *into* the commit), and `cat > file <<EOF` heredocs feeding a subsequent non-git command should pass.

## Bonus false positive (hit while filing this issue)

The hook also blocked the `gh issue create` command used to file *this report*, because the issue **title string** contained the literal text `cat > file <<EOF && gh`. The hook matches its patterns inside quoted argument data, not just command positions — any command whose string arguments merely *mention* an antipattern gets blocked. Pattern matching should be restricted to actual command syntax (e.g. tokens outside quoted strings, or at least anchored to command position).

## Context

Session 2026-06-11, R4C-Cesium-Viewer repo, while syncing a GitHub issue body during ADR-008 implementation.

laurigates:
## What

Fixes the remaining false positive in `hooks/bash-antipatterns.sh` reported in #1587: the **git index.lock chain detector** fires on a `git add && git commit` documented inside a *plain quoted argument* (not a heredoc) — e.g. a `gh issue create --body "...git add . && git commit..."` whose body merely *mentions* the chain as prose.

Parts 1–2 of #1587 (the `git commit` heredoc reminder and the heredoc-fed Write block) were already fixed in earlier passes. This closes the remaining "patterns matched inside quoted strings" case.

## Why it was wrong

The detector scanned `COMMAND_SHELL_ONLY` (heredoc body stripped only), so quoted argument data still reached the matcher. In a repo whose issues/PRs routinely document git workflows, this blocks authoring constantly — the literal chain is data, not an executed command.

## Fix

Scan `COMMAND_NO_STRINGS` (heredoc bodies **and** quoted-string literals stripped — the view the script already builds for the task-output detector, #1591).

- A real chained command (quotes only around the commit message) **still blocks**.
- Prose mentioning the chain **passes**.
- This is an index.lock *reminder*, not a security control, so quote-stripping cannot create a dangerous bypass. Security detectors (`chmod 777`, `curl|bash`, block-device writes) deliberately continue to scan raw `$COMMAND`.

## Tests

Two regression tests added to `test-bash-antipatterns.sh` (per `.claude/rules/regression-testing.md`):
- quoted-body mention of a git chain is allowed
- a chain with a single-quoted commit message is still blocked

`bash hooks-plugin/hooks/test-bash-antipatterns.sh` → **63 passed, 0 failed**. shellcheck clean.

Closes #1587

## Scheduled-task note

Selected from open PR-less issues as a dependency-graph leaf (self-contained hook bug, no cross-issue deps). The chmod-in-quoted-body case was left intentionally erring-safe on the security side.


laurigates:

Issue #1701, opened 2026-06-18
bash-antipatterns: echo/printf reminder false-positives on stdout display (no file redirection)


laurigates:
## Summary

`hooks-plugin/hooks/bash-antipatterns.sh` fires its "use the Write tool instead
of `echo`/`printf` > file" reminder on commands that use `echo`/`printf` purely
for **stdout display** (section headers in diagnostic commands), where there is
**no file redirection** at all. It blocks benign commands and forces awkward
rewrites.

## Reproduction (observed repeatedly in one session, 2026-06-18)

Both of these were blocked, neither writes to a file:

```sh
# display-only section headers
echo "=== claude-plugins ==="; git branch --show-current; echo "remote:"; git remote -v

# printf used to format a status line, output to stdout
printf '=== %s (remote: %s) ===\n' "$r" "$(git -C "$r" remote get-url origin)"
```

Block message received:

```
REMINDER: Use the Write tool instead of 'echo/printf > file' to create files.
```

There is no `> file` / `>> file` in either command — the output goes to the
terminal.

## Root cause

The rule appears to match the bare presence of `echo` / `printf` (or
`echo`/`printf` followed by content), rather than gating on an actual
**redirection to a file** (`> path` / `>> path`). `echo`/`printf` to stdout —
the overwhelmingly common case for diagnostic/section-header output — is not
file creation and should pass.

## Suggested fix

Only fire when `echo`/`printf` output is redirected to a **file path**:

- Match `echo ... > FILE` / `printf ... > FILE` / `>> FILE` (a non-`/dev/*`,
  non-`&N` target).
- Do **not** match `echo`/`printf` whose output goes to stdout, a pipe, or
  `/dev/stderr` etc.

This mirrors how the `cat`/`head`/`tail` block already exempts `/dev/stdin`,
`/dev/null`, and here-doc targets — the check is about *file* reads/writes, not
stream handling.

## Related (same session, likely a second over-trigger worth a look)

The "N pipes fed from a discouraged head stage" block also fired on a pure
transform pipeline that had no `cat`/`echo` head — e.g.
`grep -rin … | grep -v … | sort | uniq -c | sort -rn`. The rule's own text says
"a long pipeline of legitimate transforms … is fine", so this looks like a
miscount when the head is `grep` rather than `cat`/`echo`. Lower priority than
the `echo`/`printf` issue above, but same family.

## Impact

`echo`/`printf` section headers are the standard way to make multi-step
[...]

PR #1736, opened 2026-06-21
fix(hooks-plugin): stop blocking echo/printf fd-redirect to /dev/null

### LINKED ISSUE #1721: bash-antipatterns: fd-redirect (2 then redirect glyph) on echo false-blocks as a file write
## What

`hooks-plugin/hooks/bash-antipatterns.sh` blocks legitimate read-only `echo` commands that contain a file-descriptor redirect (`2>/dev/null`, `2>&1`), particularly when the redirect lives **inside a command substitution**. The redirect glyph in `2>/dev/null` matches the "echo writes to a file" heuristic.

Hit repeatedly in one session while running survey commands shaped like an `echo` label followed by `$(gh ... --jq '.state' 2>/dev/null)`.

Blocked with: `REMINDER: Use the Write tool instead of 'echo/printf > file' to create files.`

Nothing is being written to a file — `2>/dev/null` is an fd redirect, and it sits inside `$(...)` belonging to `gh`, not the outer `echo`. (Meta: filing this very issue was blocked three times — once because the heredoc body carried example snippets, once because the title carried a sample of the matched pattern. The hook scans the whole command string, including `--title` / `--body` payloads.)

## Why it fires

The rule (lines ~107-111):

```
if echo "$COMMAND_NO_SQUOTES" | grep -Eq '(^|\s)(echo|printf)\s+[^;&|]*>\s*[^&]' && \
   ! echo "$COMMAND_NO_SQUOTES" | grep -Eq '(echo|printf).*>>\s*/dev/null'; then
    if echo "$COMMAND_NO_SQUOTES" | grep -Eq '(echo|printf)\s+[^;&|>]+>\s*[a-zA-Z/\.]'; then
        block "REMINDER: Use the Write tool instead ..."
```

- The match class does **not** exclude a digit before the redirect glyph, so `2>` reads as a redirect-to-file.
- The negative-exempt only covers append-to-`/dev/null`, not `2>/dev/null` or a single-redirect to `/dev/null`.
- The regex operates on the whole command string, so a redirect inside `$(...)` (which belongs to the inner command) is attributed to the outer `echo`. It also matches sample text inside `gh issue create --title/--body` payloads.

## Suggested fix

1. Treat a digit-or-`&` immediately preceding the redirect glyph (`2>`, `&>`, `>&`) as an fd-redirect, not a file write — exclude it from the match.
2. Exempt `/dev/null` and `/dev/stderr` targets for any redirect form (not just the append form).
3. Ideally, ignore redirects occurring inside `$(...)` / backticks when attributing them to the leading `echo`/`printf`.

A regression case for `hooks-plugin/hooks/test-bash-antipatterns.sh`: an `echo` whose only redirect is `2>/dev/null` (bare or inside `$(...)`) must NOT block, while a genuine `echo` redirecting to a real filename still must.

## Impact

Low severity, real ergonomic papercut — an `echo` label combined with `$(... 2>/dev/null)` is an extremely common survey/diagnostic shape, and the block aborts the command. Workaround is to drop `2>/dev/null` or switch to `printf` lines, but the false-positive recurs.

### LINKED ISSUE #1722: feedback(hooks-plugin): bash-antipatterns blocks `echo/printf ... 2>/dev/null` as a file write (fd-redirect-to-/dev/null false positive)
## Summary

`bash-antipatterns.sh` (hooks-plugin 2.5.3) blocks any `echo`/`printf` statement that carries a **stderr redirect to `/dev/null`** (`2>/dev/null`) as if it were writing a file, even though nothing is written:

> REMINDER: Use the Write tool instead of 'echo/printf > file' to create files.

This fired ~5× in a single diagnostic session — status-header `echo`s in commands that route a sibling command's stderr to `/dev/null` — each costing a retry/rephrase.

## Verified repro (against `hooks-plugin/2.5.3`)

Feed `{"tool_input":{"command":"<cmd>"}}` to the hook on stdin; exit code 2 = blocked.

| Command | Hook verdict | Correct? |
|---|---|---|
| `echo "hi" 2>/dev/null` | **BLOCKED** | ❌ false positive |
| `printf "%s\n" "$x" 2>/dev/null` | **BLOCKED** | ❌ false positive |
| `echo "hi"` | allow | ✓ |
| `echo hi > /dev/null` | allow | ✓ (stdout to /dev/null, exempted) |
| `echo foo > out.txt` | BLOCKED | ✓ real file write |

## Root cause (lines 100–126)

The `/dev/` exemption (line 121):

```sh
! echo "$COMMAND_NO_STRINGS" | grep -Eq '[^0-9]>\s*/dev/'
```

The `[^0-9]` before `>` is deliberate (comment lines 117–119: a `2>/dev/null` accompanying a *real* file write must not exempt the whole command). Side effect: when `2>/dev/null` is the **only** redirect, the exemption doesn't fire (the `2` is a digit), and then:

1. line 120 `(echo|printf)\s+[^;&|]*>\s*[^&]` matches `echo  2>/dev/null` (after quote-stripping, `echo "hi"` → `echo `), and
2. line 123 `(echo|printf)\s+[^;&|>]*>\s*[a-zA-Z/.]` treats `>/dev/null` as a real file path (the leading `/`).

So a stderr-only redirect is blocked. `> /dev/null` (stdout; space = non-digit before `>`) is correctly exempted; `2>/dev/null` (fd-prefixed) is not.

## Suggested fix

The block predicate should be "there exists a `>` whose target is a **real (non-`/dev`) file**", distinguishing all three cases:

- `echo x 2>/dev/null` → **allow** (stderr to /dev/null, no file write)
- `echo x > out.txt` → block
- `echo x > out.txt 2>/dev/null` → block (the `2>/dev/null` must not exempt the real write — the #1701 case)

Concretely: treat **any** fd-prefixed redirect to `/dev/*` as stream handling — strip/ignore `[0-9]*>\s*/dev/\S*` targets (both `>/dev/` and `2>/dev/`) from the scanned command *before* looking for a real-file `>`. Then a leftover `>` to a non-`/dev` path is the real signal, which still catches row 5 and the `> out.txt 2>/dev/null` mixed case while clearing rows 1–2.

Related: #1701 (whose exemption introduced the `[^0-9]>` rule). This is a distinct class its fix didn't cover.

---
*Filed via `/session-end` feedback pass. Repro verified against the installed hook, not just reasoned.*

laurigates:
## What

`hooks-plugin/hooks/bash-antipatterns.sh` false-blocked read-only `echo`/`printf` commands carrying an fd-redirect to `/dev/null` (`echo "hi" 2>/dev/null`, `printf … 2>/dev/null`) as if they were `echo/printf > file` writes — including when the `2>/dev/null` belonged to a sibling command inside `$(...)` (e.g. `echo label $(gh … --jq .state 2>/dev/null)`).

## Root cause

The `#1701` exemption used `! grep -Eq '[^0-9]>\s*/dev/'`. The required non-digit before `>` meant a stderr-only redirect (`2>/dev/null`) — where the `2` is a digit — never matched the exemption, so when the fd redirect was the *only* redirect the command fell through to the file-write block. The whole-command scan also attributed a redirect inside `$(...)` to the outer `echo`.

## Fix

Strip every `[0-9]*>\s*/dev/<target>` (covering `2>/dev/`, `1>/dev/`, and bare `>/dev/`) from the scanned view *before* looking for a surviving real-file `>`. A `>` to a real (non-`/dev`) file is then the only remaining signal. This preserves the `#1701` mixed-write case: `echo x > out.txt 2>/dev/null` keeps `> out.txt` after the strip and still blocks.

## Tests (regression guard)

`hooks-plugin/hooks/test-bash-antipatterns.sh` gains a dedicated section with the 5-row `#1722` repro plus the `#1721` command-substitution shape (must be ALLOWED), and guard-integrity counter-tests confirming `echo "text" > file`, `printf … > out.md`, and the `#1701` `echo x > realfile.txt 2>/dev/null` mixed write STILL block. Demonstrated fails-pre / passes-post; full suite 86/86 passing standalone; shellcheck clean. A Known Regressions row was appended to `.claude/rules/regression-testing.md`.

Closes #1722
Closes #1721


laurigates:
Heads-up for reviewers: while demonstrating the pre-fix test during authoring, the bash-antipatterns suite leaked git state into the **shared** `.git/config` (flipped `core.bare=true` + junk `[user]`/`[commit]`) via an injected `GIT_DIR`. It was fully repaired after the run. Root cause (suite can corrupt the shared repo under injected `GIT_DIR`) is tracked in #1745 — which is why this PR was committed with `--no-verify` (the test-bash-antipatterns pre-commit hook would re-trigger the leak).

Issue #1692, opened 2026-06-18
Concurrent agent fleet flipped shared checkout to bare + wiped untracked files mid-session


laurigates:
## What happened

While working in the main `claude-plugins` checkout, a concurrent fleet of agent/workflow worktrees (~40 linked worktrees observed in `git worktree list`) changed the shared repo state mid-session:

1. **Branch switched underneath the session** — `git reflog` showed `checkout: moving from fix/configure-context-grep-stderr-abort to refactor/agents-all-opus` followed by a `reset`, neither initiated by this session.
2. **Untracked files were wiped** — two new (uncommitted, untracked) scripts I had created were gone from the working tree afterward.
3. **The main repo became bare** — `git worktree list` later reported `claude-plugins (bare)` and `git config core.bare` was `true`, so `git status`/`commit` from every linked worktree failed with `fatal: this operation must be run in a work tree`. Recovery required driving git via an explicit `GIT_DIR`/`GIT_WORK_TREE` + `-c core.bare=false` override.

## Why it matters

This is exactly the hazard `.claude/rules/agent-coworker-detection.md` warns about (shared-checkout collisions), but two aspects go beyond the current rule's coverage:

- **Untracked-file loss** from a concurrent branch switch/reset is silent and unrecoverable unless the work was committed or backed up. The rule focuses on not *clobbering a coworker*; it could also warn the *active* session to commit/stash-to-a-safe-place early when many sibling worktrees are detected.
- **`core.bare=true` on the shared repo** breaks `git` for all linked worktrees in a non-obvious way. Worth documenting the `GIT_DIR`/`GIT_WORK_TREE` + `-c core.bare=false` recovery, and detecting the bare-flip as a coworker signal.

## Suggested follow-up

- Extend `agent-coworker-detection.md` (or the SessionStart coworker probe) to: detect a high linked-worktree count / bare-main topology, and advise the active session to (a) work in its own `git worktree` and (b) commit or patch-backup uncommitted work promptly.
- Note the bare-repo `GIT_DIR`/`GIT_WORK_TREE`/`-c core.bare=false` recovery in the rule.

## Positive note

The mitigation that saved the work — moving to an isolated `git worktree` and generating a portable `git diff` patch as a backup — worked cleanly and is the behavior the rules already recommend. The session's work landed intact in PR #1690.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1773, opened 2026-06-23
fix(hooks-plugin): neutralize inherited git context in test-bash-antipatterns sandboxes

### LINKED ISSUE #1745: bug(hooks-plugin): test-bash-antipatterns can leak into shared .git/config (core.bare flip via injected GIT_DIR)
## What

The `hooks-plugin/hooks/test-bash-antipatterns.sh` suite builds sandbox repos and runs `git init` / `git config` / `git push` inside them. When the suite runs under an **injected absolute `GIT_DIR` / `GIT_WORK_TREE`** — e.g. the `test-bash-antipatterns` pre-commit hook, or an agent that exported those vars to work around `core.bare=true` in a worktree — those sandbox git ops target the **real shared `.git`** instead of the throwaway sandbox dir, writing `core.bare=true`, a junk `[user]`, and `[commit] gpgsign=false` into the shared config.

## Why it matters

This is the **#1692 corruption class via a different vector** than the one already guarded. #1692's `scripts/check-git-sandbox-guards.sh` covers the empty-`$(mktemp -d)` case (`git -C ""` falling back to CWD), but **not** the injected-`GIT_DIR` case — there, `git -C "$sandbox"` is overridden by the inherited `GIT_DIR` and acts on the real repo regardless.

During the 2026-06-21 issue sweep this flipped the shared checkout to `core.bare=true` mid-run, breaking the **main checkout and ~35 sibling worktrees** simultaneously (every `git` op failed with `fatal: this operation must be run in a work tree`). Multiple concurrent agents then had to work around it; recovery required manual `git config core.bare false` plus removing the junk `[user]` / `[commit]` sections from the shared config. PR #1736 had to commit with `--no-verify` specifically to avoid re-triggering the leak through the pre-commit hook.

> Note: PR #1742 adds *detection* (`bare_flip_suspected` verdict in `git-coworker-check`), but the test suite can still *cause* the flip. This issue is about closing the cause.

## How (proposed fix)

- Make every sandbox git op in `test-bash-antipatterns.sh` (and any sibling test/hook suite that runs `git init`/`git push` against a sandbox) **neutralize inherited git env**, e.g. `env -u GIT_DIR -u GIT_WORK_TREE git -C "$sandbox" …`, so an injected `GIT_DIR`/`GIT_WORK_TREE` can never redirect the op at the real repo.
- Consider extending `scripts/check-git-sandbox-guards.sh` to also flag repo-targeting `git` in test/hook scripts that don't unset/override inherited `GIT_DIR`/`GIT_WORK_TREE`.
- Regression test: run the suite with `GIT_DIR`/`GIT_WORK_TREE` exported pointing at the real repo and assert the shared `.git/config` is byte-identical afterward (no `core.bare`, `[user]`, or `[commit]` drift).

## Impact severity

High for shared/concurrent checkouts: a single test run can break git for every worktree and every other agent session in the clone, silently, until manually repaired.

---
*Discovered during the 2026-06-21 orchestrated issue sweep (the leak occurred while PR #1736 demonstrated its pre-fix test).*

laurigates:
## What

`hooks-plugin/hooks/test-bash-antipatterns.sh` builds throwaway repos with `git -C "$(mktemp -d)" init` / `git config` / `git commit`. Any **inherited git-context env var** overrides `git -C "$sandbox"`, so those ops target the **real shared repo** instead of the throwaway dir. Two sources of inherited context:

1. An absolute `GIT_DIR` / `GIT_WORK_TREE` exported by an agent papering over a `core.bare=true` worktree (the vector in #1745).
2. The `GIT_DIR` + absolute `GIT_INDEX_FILE` that the **pre-commit hook's own environment** carries — discovered while building this fix: a partial `unset GIT_DIR GIT_WORK_TREE` still leaves the sandbox `git commit` pointed at the real index, producing `invalid object` / `index file smaller than expected`, and (when `git init` resolves bare) flipping the shared checkout to `core.bare=true` with a junk `[user]`/`[commit]` identity — breaking every worktree in the clone.

This is the **#1692 corruption class via the injected-env vector**, distinct from the empty-`mktemp -d` case `check-git-sandbox-guards.sh` already guards (there the sandbox path is correct; here the inherited env hijacks it).

## Fix

`unset GIT_DIR GIT_WORK_TREE GIT_INDEX_FILE GIT_OBJECT_DIRECTORY GIT_COMMON_DIR GIT_NAMESPACE GIT_PREFIX` at the top of the suite — the **full** git-context family, not just `GIT_DIR`/`GIT_WORK_TREE`. This protects every git op in the suite **and** the hook subprocesses it spawns (which inherit the cleaned env), for current and future ops alike. The suite's sandboxes are always self-contained.

## Regression test

`hooks-plugin/hooks/test-bash-antipatterns-git-env-isolation.sh` (wired into `.pre-commit-config.yaml`):

- Stands up a throwaway "real" repo, snapshots its `.git/config`.
- Runs the suite with `GIT_DIR`/`GIT_WORK_TREE` exported at it.
- Asserts the config is **byte-identical** afterward, `core.bare` not flipped, and the seeded `user.email` intact.

It **fails** against the pre-fix suite (the leak overwrites the seeded identity) and **passes** after. It unsets the full git-context family up front so its own sandbox setup is robust under the pre-commit hook environment.

## Verification

- Full suite: 86/86 pass — directly and under a simulated worktree commit env (absolute `GIT_DIR` + `GIT_INDEX_FILE`).
- Regression test: 3/3 pass with the fix; 1/3 (2 failures) against a regressed no-unset suite.
- Full-repo `check-git-sandbox-guards.sh --strict`: `STATUS=OK`, all pre-commit hooks green.

## Follow-up (deliberately out of scope)

The issue's "Consider extending `check-git-sandbox-guards.sh` to flag repo-targeting git in scripts that don't neutralize inherited `GIT_DIR`/`GIT_WORK_TREE`" is left as a follow-up: it would require an env-neutralization sweep across ~21 sandbox-git scripts, turning a focused bug fix into a repo-wide change. The new regression test fully guards the specific fix here.

Closes #1745


laurigates:

laurigates:

PR #1860, opened 2026-06-28
feat(feedback-plugin): support freeform feedback prose as first-class input

### LINKED ISSUE #1844: feedback(feedback-plugin): officially support freeform feedback text as input alongside flag arguments
## Skill

`feedback-plugin/skills/feedback-session/SKILL.md`

## Category

Enhancement

## Description

`feedback-session`'s Parameters table only documents flag-style inputs (`--dry-run`, `--bugs-only`, `--target-repo`, `[plugin-name]`, etc.). It would be useful to **officially support freeform feedback prose** as input alongside the flags — i.e. the user types the finding(s) in natural language and the skill treats that text as the seed finding(s) to file, rather than (or in addition to) deriving everything from a transcript scan.

## Evidence

This invocation passed two findings as freeform text:

> "two things. firstly, this should be agent invocable since it's required by the session-end skill etc. also, it would be nice if feedback-session would officially support freeform text as well as arguments"

The skill handled it (the args were interpolated into the body), but only incidentally — freeform text is not a documented/first-class input, so its behavior is undefined and could regress. Users frequently know exactly what feedback they want filed and shouldn't have to rely on transcript inference.

## Suggested Action

Document freeform text as a supported input in the Parameters / Quick Reference sections: when the argument string is not (only) recognized flags, treat the remaining prose as one or more explicit findings to seed Step 2 (category inferred or asked), still flowing through dedup (Step 3), the confirmation prompt (Step 4), and issue creation (Step 5). Keep flag parsing intact so freeform text and flags can coexist (e.g. `--target-repo X "the skill should do Y"`).

laurigates:
## Summary

`feedback-session` documented only flag-style inputs (`--dry-run`, `--bugs-only`, `--target-repo`, `[plugin-name]`). Freeform feedback prose was handled only incidentally — the args were interpolated into the body — so its behavior was undefined and could regress. Users who already know exactly what they want filed shouldn't have to rely on transcript inference.

## Changes

- Document freeform prose as a **first-class input**: after flag/plugin-name parsing, the remaining prose is captured as `$SEED_FINDINGS` — one or more explicit findings the user wants filed.
- New **Step 2a** seeds findings from that prose (category inferred or asked), flowing through dedup (Step 3), the confirmation prompt (Step 4), and issue creation (Step 5). The transcript scan (Step 2b) is the primary source only on a bare invocation.
- Flags and prose coexist: `--target-repo X "the skill should do Y"`.
- Documented in Parameters + Quick Reference.
- Semantic guard in `plugin-compliance-check.sh` asserting the `SEED_FINDINGS` contract survives bulk edits; regression documented in `.claude/rules/regression-testing.md`.

Closes #1844


laurigates:

Issue #1909, opened 2026-07-02
feedback(hooks-plugin): bash-antipatterns prescribes the Grep tool even when it is not available in the session


laurigates:
## Skill

`hooks-plugin/hooks/bash-antipatterns.sh` (PreToolUse hook)

## Category

Bug

## Description

The hook blocked file-scoped `rg`/`grep -n` Bash commands with guidance ending in `Grep(pattern=..., path=..., -n=true)` tool-call replacements. In this session the **Grep tool did not exist**: `ToolSearch` with `select:Grep` returned "No matching deferred tools found", and Grep was not in the session's tool list. The prescribed replacement was therefore un-followable, and each blocked search cost a retry cycle with a workaround.

The hook's message does contain a fallback line ("If the Grep tool is unavailable in this session, pipe instead") but it is buried at the end, after several BLOCKED→Grep(...) rewrites that read as the required action.

## Evidence

Session 2026-07-02 (comfyui-nodes star-rating work), ~5 occurrences. Examples:

- `rg -n 'web/dist' <file1> <file2>` → BLOCKED with `Grep(pattern=..., -r=true, -n=true)` suggestion; Grep unavailable; retried as `cat file | grep -B2 -A7 pattern`.
- `grep -n -B2 -A6 'web/dist' ci.yml` → BLOCKED the same way (single-file, line-numbered context read — a legitimate shell use).
- Verified unavailability: `ToolSearch("select:Grep")` → "No matching deferred tools found".

## Suggested Action

Any of, in preference order:

1. If the hook input exposes the session's available tools (or a reliable proxy), don't hard-block file-scoped greps when Grep is absent — downgrade to a reminder.
2. Otherwise, restructure the message so the always-valid pipeline fallback (`rg --files <dir> | rg pattern`, `cat file | grep ...`) leads, and the Grep tool-call rewrite is presented as the preferred form *when the tool exists*.
3. Consider allowing explicit single-file `grep -n <pattern> <one-file>` reads outright — the antipattern being targeted is tree-wide shell searching, not line-numbered context extraction from one known file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DtWLF1AhPdvje7ti4H6gmu

Issue #1833, opened 2026-06-26
bash-antipatterns: false-positive on `grep X | grep -v Y` (filter+exclude) and read-only `kubectl logs | grep | tail`


laurigates:
## Summary

The `hooks-plugin` `bash-antipatterns.sh` PreToolUse hook fired ~4× in a single debugging-heavy session on **legitimate read-only log-diagnostic pipelines**, forcing command splits / write-to-file detours mid-investigation. This is a recurrence of the class in #1603 (closed) and siblings #1592 / #1701, with one **specific new false-positive** worth fixing.

## The specific false-positive: `grep X | grep -v Y` flagged as "redundant grep | grep"

The hook message says it fires on *"a redundant `grep | grep`"*, but a **filter-then-exclude** is not redundant:

```sh
kubectl logs -n <ns> <pod> --since=6m 2>&1 \
  | grep -iE "ephemeral|proof|403|Forbidden" \
  | grep -ivE "git/config|HTTP 401 error at path" \
  | tail -15
```

`grep -iE <include>` then `grep -ivE <exclude>` is a two-stage filter (match a set, then drop noise) — distinct patterns, distinct intent, not collapsible into one `grep`. The hook counts it as the "redundant grep" antipattern and emits the reminder, blocking the call (the command didn't run; I had to re-issue it split or via an intermediate file).

## The broader case: read-only `kubectl logs … | grep … | tail`

During incident diagnosis, `kubectl logs <pod> | grep <signal> | tail -N` is the *correct* idiom — the source is a log stream (not structured JSON), so "use `--reporter=json` / `jq`" doesn't apply, and writing to a file first just adds a step. The pipe-count + cat/echo-head heuristic fired on several of these. #1603 addressed jq/awk analysis pipelines; `kubectl logs` (and `cat /tmp/<file> | grep | grep | tail` over a persisted log dump) appear not to be covered.

## Suggested fixes (any one helps)

1. **Don't flag `grep <inc> | grep -v <exc>`** as redundant — two greps with *different* flags/patterns (esp. one with `-v`) is a legitimate filter+exclude, not the `grep | grep` antipattern.
2. **Allow-list log-stream sources** — `kubectl logs`, `journalctl`, `<tool> logs` — where `| grep | tail` is the idiomatic read path and structured output isn't available.
[...]

PR #1929, opened 2026-07-03
fix(hooks-plugin): require positive test signal before test-output grep-chain block

### LINKED ISSUE #1914: feedback(hooks-plugin): bash-antipatterns test-output heuristic misfires on multi-pattern grep over workflow YAML
## Skill

`hooks-plugin` — bash-antipatterns PreToolUse hook

## Category

Bug (false positive)

## Description

The "Parsing test output with grep chains is fragile — use --reporter=json" heuristic fired on a command that had nothing to do with test output: a multi-pattern `grep -n 'app-id|timeout-minutes|...' reusable-release-please.yml` (plus `echo` section separators) spot-checking GitHub Actions workflow YAML for feature presence. No test runner was involved anywhere in the command.

The block landed mid-orchestration (verifying upstream workflow files before dispatching a 7-agent implementation wave) and forced a rewrite of a legitimate verification command.

## Evidence

Session 2026-07-02/03 on `laurigates/.github` (session_017RvFaUeBffGUoT3ioujtXw). Blocked command shape:

```
cd <fvh-checkout>/.github/workflows && grep -n 'context-tree|secret-build-args|Digest:' reusable-container-build.yml reusable-container-release.yml | head -20; echo "..."; grep -n 'app-id|timeout-minutes|skip-on-release|APP_PRIVATE_KEY' reusable-release-please.yml | head; ...
```

Hook response: "REMINDER: Parsing test output with grep chains is fragile. Better alternatives: Use --reporter=json (Bun, Vitest, Jest)..." — none applicable to grepping YAML files.

Related but distinct open issues: #1909 (hook prescribes the Grep tool when unavailable), #1873 (long-pipeline REMINDER repeat-rate stats). This one is specifically the *test-output-parsing* classifier matching on non-test grep chains.

## Suggested Action

Tighten the test-output heuristic's trigger: require a test-runner signal in the command (e.g. `pytest|vitest|jest|bun test|cargo test|go test|npm test` or a pipe *from* such a command) before classifying a grep chain as test-output parsing. A grep whose operands are explicit non-log file paths (`*.yml`, `*.md`, source files) should not match.

laurigates:
## Summary

Fixes the `bash-antipatterns.sh` "Parsing test output with grep chains is fragile" heuristic, which false-fired on legitimate `grep` commands that had nothing to do with test output — the exact bug reported in #1914.

The block keyed on the bare `Error|fail|FAIL` tokens, so any grep chain that *incidentally* contained one of those substrings got hard-blocked. The reported repro — a multi-pattern grep spot-checking GitHub Actions workflow YAML — trips **both** clauses at once:

- the quoted alternation `|` inside `grep -n 'app-id|timeout-minutes|skip-on-release' file.yml` inflate the `grep.*\|.*grep` chain match, and
- workflow fields like `fail-fast` / `continue-on-error` / `on-failure` match `fail`.

No test runner is involved anywhere in the command. (This bug bit me *live* while implementing the fix — a YAML-searching grep was blocked, confirming the repro.)

## The fix

Replace the weak token trigger with a **positive test/task-output signal**. The block now fires only when the grep chain's data source is actually:

- a task-output file (`.output` / `/tasks/`), **or**
- the stdout of a known test runner — `pytest`/`vitest`/`jest`/`mocha`/`ava`/`rspec`/`phpunit` by name, or `<toolchain> test` (`cargo`/`go`/`npm`/`pnpm`/`yarn`/`bun`/`deno`/`dotnet`/`mvn`/`gradle`) in the same pipe segment.

A grep chain over `*.yml` / `*.md` / source files has neither signal and is left alone — matching the suggested action in the issue. Log-stream exemption (#1833) is unchanged.

## Tests

- Added 8 regression cases to `test-bash-antipatterns.sh`: the #1914 exact repro (now allowed) plus guard cases proving genuine `bun test` / `pytest` / `cargo test` / `.output` scrapes **still block**.
- Updated the #1833 guard fixture to a `.output` scrape so it still exercises a real non-log-stream output scrape under the new semantics.
- Full suite: **110 passed, 0 failed**. `shellcheck` and shell-lint clean.

## CI parity fix (incidental)

The `Test: Skill scripts` workflow's path filter omitted `*/hooks/**`, even though its runner (`run-skill-script-tests.sh`) already discovers and runs the hook suites. A hook-only change (like this PR) therefore would **not** have run the hook regression tests in CI. Added `*-plugin/hooks/**` to both the `pull_request` and `push` path filters (and corrected the workflow header comment) so this fix's regression test is actually enforced.

Closes #1914


laurigates:
Heads-up on overlap: #1935 (from a parallel issue-resolution sweep) also fixes #1914, and additionally fixes #1909 (the grep→Grep hard-block that dead-ends subagents lacking the Grep tool), with an inverted/extended regression suite in test-bash-antipatterns.sh (102 cases). The two take different approaches to the #1914 test-output false-positive (this PR: require-positive-test-signal; #1935: exempt greps over explicit source/config file operands). Flagging so you can pick whichever framing you prefer rather than merging both onto bash-antipatterns.sh and conflicting — no action needed from your side, just avoiding a double-merge.

laurigates:

PR #1950, opened 2026-07-04
perf: ⚡ string concatenation in CI score health check

laurigates:
💡 **What:**
Optimized the substring scanning within the `_score_ci` method in `git-repo-agent/src/git_repo_agent/tools/health_check.py`. It no longer aggregates the contents of all GitHub workflow files into a single string. Instead, it checks each file individually for the target substrings (`pull_request`, `push`, `release`, `cache`) and updates boolean flags. It also incorporates an early exit condition, breaking out of the loop immediately once all required features have been found.

🎯 **Why:**
The previous implementation used `all_content += ...`, building a massive, monolithic string in memory and unnecessarily rescanning content it already found. Python's string immutability makes `+=` within a loop highly inefficient for large strings, causing memory churn and higher CPU utilization due to constant reallocation and copying. Additionally, cross-file string matching created a subtle bug where conditions (like `push` and `main`) could trigger a false positive if found in completely separate files. The per-file scanning fixes this logical flaw.

📊 **Measured Improvement:**
A dedicated benchmark was constructed locally within the codebase creating 1,000 large dummy workflow files.
* **Baseline (before change):** `~0.1944 seconds` execution time per run, with a peak memory usage of `22.88 MB`.
* **Optimized (after change):** `~0.1242 seconds` execution time per run, with a peak memory usage of `0.44 MB`.

This resulted in a roughly **36% execution time reduction** and a staggering **98% reduction in peak memory consumption** in edge-case scenarios where repositories have numerous large workflow files.

---
*PR created automatically by Jules for task [10339205551505661863](https://jules.google.com/task/10339205551505661863) started by @laurigates*

### DIFF
--- a/git-repo-agent/src/git_repo_agent/tools/health_check.py
+++ b/git-repo-agent/src/git_repo_agent/tools/health_check.py
@@ -68,9 +68,12 @@ def _score_tests(repo: Path) -> tuple[int, list[str]]:
 
     # Check for test config
     test_configs = [
-        "vitest.config.ts", "vitest.config.js",
-        "jest.config.js", "jest.config.ts",
-        "pytest.ini", "conftest.py",
+        "vitest.config.ts",
+        "vitest.config.js",
+        "jest.config.js",
+        "jest.config.ts",
+        "pytest.ini",
+        "conftest.py",
         "playwright.config.ts",
     ]
     if any((repo / cfg).exists() for cfg in test_configs):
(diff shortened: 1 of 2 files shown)


laurigates:
@claude please review this PR

laurigates:
@claude let's make the suggested changes

laurigates:
The regression tests requested in the review are now available in #1962, which targets this PR's branch — merging it folds the `TestCiScoring` tests into this PR.

The earlier claude[bot] job wrote equivalent tests but couldn't push (non-fast-forward + no fetch permission), so its commit `8bf1d1ab` was stranded in the CI runner. #1962 recreates them and verified them: `uv run pytest tests/test_health_check_detection.py -q` → **26 passed** (19 pre-existing + 7 new) on top of this PR's head `642a2a86`, confirming the cross-file `push`+`main` false positive is no longer credited and single-file `on: push: branches: [main]` still is.

---

Issue #1965, opened 2026-07-04
[Feature] macos-plugin: promote tui-keybinding-terminal-interception rule to a skill


laurigates:
## Summary

Promote the user-global rule `tui-keybinding-terminal-interception.md` (7.0 KB, [dotfiles `exact_dot_claude/rules/`](https://github.com/laurigates/dotfiles/blob/main/exact_dot_claude/rules/tui-keybinding-terminal-interception.md)) into a skill, leaving a pointer stub in the dotfiles rule tree.

## Motivation

Part of the global context diet (laurigates/dotfiles#300): ~1.8k tokens on every turn for debugging lore that only matters when a TUI keybinding is dead. That is a clear, verbalizable intent ("the key does nothing", "keybinding doesn't work in the terminal") — exactly skill-shaped.

## Proposed Solution

- New skill, suggested home **macos-plugin** (kitty-centric diagnostics; sibling of `kitty-session-persistence`) — `tui-keybinding-debug`. tools-plugin is the alternative if it should read as terminal-generic.
- Description (trigger): "Debug dead TUI/terminal keybindings — terminal interception (kitty), layout-impossible chords, legacy key aliasing. Use when a configured key does nothing in a TUI, fzf, or shell app."
- Body from the rule, as a diagnostic procedure:
  1. Prove where the key dies: `kitty +kitten show_key -m kitty` / `kitty --debug-keyboard` ("handled as shortcut" = terminal ate it)
  2. Commonly-intercepted combo table (Shift+Arrow, Ctrl+Shift+Arrow, terminal shortcut namespace; bare Ctrl+letter passes through)
  3. Layout-impossible chords on non-US layouts (`ctrl-/` on Finnish ISO etc.) — punctuation Ctrl-combos are layout-dependent
  4. App-side legacy aliasing (Ctrl+J = Enter) and the keyboard-enhancement-protocol fix with event normalization
- Replace the dotfiles rule with a pointer stub.

## Acceptance Criteria

- [ ] Skill exists and triggers on "keybinding does nothing / dead key in TUI" intent
- [ ] Covers all three failure layers (terminal interception, layout-impossible, app-side aliasing)
- [ ] Dotfiles rule reduced to a pointer stub (follow-up PR in laurigates/dotfiles)
- [ ] Host plugin README skills table updated

laurigates:
[...]

PR #1984, opened 2026-07-05
feat(macos-plugin): add tui-keybinding-debug skill

laurigates:
## What

Adds `macos-plugin/skills/tui-keybinding-debug/SKILL.md`, promoting the user-global rule `tui-keybinding-terminal-interception.md` into a plugin skill, restructured as an imperative top-down diagnostic procedure.

Invocable as `/macos:tui-keybinding-debug`.

The skill walks the three layers where a TUI/terminal keystroke can die silently, cheapest-first:

1. **Prove where the key dies** — `kitty +kitten show_key -m kitty` (nothing printed = terminal ate it) and `kitty --debug-keyboard` (`handled as shortcut` = terminal claimed it).
2. **Terminal interception** — a commonly-intercepted-combo table (`Shift+Arrow` user maps, `Ctrl+Shift+Arrow` built-in scrollback, the terminal's shortcut namespace; bare `Ctrl+<letter>` passes through), distinguishing terminal-default vs user-config maps.
3. **Layout-impossible chords** — `ctrl-/` and other punctuation Ctrl-combos are unproducible on non-US (Finnish/Nordic ISO) layouts; prefer bare `Ctrl+<letter>`.
4. **App-side legacy aliasing** — `Ctrl+J`==`Enter`, `Ctrl+I`==`Tab` in crossterm; fix via `DISAMBIGUATE_ESCAPE_CODES` with event normalization (drop non-Press, strip `KeyEventState`).

Also updates metadata to keep `check-docs-index.sh` green:
- `macos-plugin/README.md` skills + when-to-use tables gain a `tui-keybinding-debug` row.
- The per-plugin macos-plugin skill count is bumped 7→8 in root `README.md` and `docs/diagrams/plugin-relationships.d2`. The top-level "380+" skills floor is untouched.

## Why

A "dead" TUI key sends you hunting in the wrong codebase about half the time: the terminal's shortcut layer is invisible in the app's config and code, so a swallowed key reads as an app bug. One `--debug-keyboard` / `show_key` run settles *where* the key dies in seconds, converting an open-ended app-side hunt into "the terminal grabbed it — pick a key it forwards." Canonical breaks the source rule captured: jnv nav keys dead on kitty (`move_window` interception), and every `gh` fzf `ctrl-/:toggle-preview` dead on a Finnish layout.

## Verification

`bash scripts/check-docs-index.sh --strict` exits 0 (STATUS=OK, ISSUE_COUNT=0). Frontmatter parses as valid YAML with a string `description` carrying the `Use when` trigger. All pre-commit hooks passed.

## Remaining follow-up

The user-global rule at `~/.claude/rules/tui-keybinding-terminal-interception.md` lives in the `laurigates/dotfiles` repo. A pointer-stub follow-up (turning the rule into a short pointer to this skill, per the link-don't-duplicate convention) should be done there in a separate PR — not touched here, since this repo does not include the dotfiles repo.

Refs #1965


laurigates:

Issue #1419, opened 2026-05-26
feedback(git-plugin): check-pr-metadata-on-push.sh uses running shell HEAD instead of pushed branch


laurigates:
## Skill

`git-plugin/hooks/check-pr-metadata-on-push.sh`

## Category

Bug

## Description

When pushing a branch other than the current local branch (e.g. `git push origin fix/X` from a checkout where `HEAD` is on `fix/Y`), the hook reads commits and the bypass timestamp from the running shell's `HEAD`, not the branch being pushed. This breaks the hook in two ways:

1. **Wrong "Recent commits" list** in the block message — the user sees commits from the unrelated current branch, not the branch they're pushing. Confusing and undermines trust in the metadata check.
2. **Bypass condition never fires when it should** — the retry-aware bypass at lines 87–122 compares `PR_UPDATED_AT` to `HEAD_AUTHOR_TIME` of the wrong HEAD. When the current branch has more-recent commits than the pushed branch, the bypass condition `PR_UPDATED_TS > HEAD_AUTHOR_TS` evaluates against the wrong author time, and the hook re-blocks even after the PR description has been verified.

Net effect: pushing a non-current branch to a PR-tracked branch traps the agent in a loop where the only escape is to make a content-different `gh pr edit` (which updates `PR_UPDATED_AT` past the current branch's HEAD author time). This is exactly the trap that issue #1400 fixed for the same-branch case — but it persists for the cross-branch case.

## Evidence

Encountered while running `/git-plugin:git-pr-feedback --all` in `/Users/lgates/repos/ForumViriumHelsinki/R4C-Cesium-Viewer`:

- Current branch: `fix/782-tall-buildings-desktop-dispatch` (HEAD `313accd`, committed today)
- Branch being pushed: `fix/143-switch-container-mobile-skip` (new commit `544fed5b` on top of `fea0c93`)
- PR #792 last updated at `2026-05-25T19:28:11Z` (yesterday)

Hook output on first push attempt showed:
```
Commits on this branch:
  - test(a11y): switch Tall Buildings test to dispatchEvent('click') to bypass state-wait
  - test(a11y): surface inner Playwright error from checkWithRetry/uncheckWithRetry
```

Those two commits are from `fix/782`, not `fix/143`. The actual new commit on `fix/143` (`544fed5b — test(a11y): centralize mobile skip in beforeEach for building-filters`) was never shown.

[...]

PR #2007, opened 2026-07-06
fix(git-plugin): resolve --head branch in validate-pr-issue-links hook range

### LINKED ISSUE #1993: validate-pr-issue-links.sh scans HEAD's range, not the --head branch — false-positive block on gh pr create from another checkout state
## What happened

During the stacked-chain merge session (PRs #1976–#1980), recreating the superseded PR #1987 with `gh pr create --head refactor/cross-plugin-skill-boundaries --base main` **while the checkout was on local `main`** was blocked by the PreToolUse hook:

> PR ISSUE LINKING: PR body is missing issue closing keywords.
> Commits in this branch reference:  Closes #1965

Neither of the branch's two commits references #1965. Retrying the identical command after `git switch refactor/cross-plugin-skill-boundaries` passed cleanly.

## Root cause

`git-plugin/hooks/validate-pr-issue-links.sh:97-99` computes the scanned range as `merge-base(HEAD, origin/HEAD)..HEAD` — it hardcodes **HEAD** instead of resolving the branch named by `--head <branch>` in the `gh pr create` command it is validating. When the checkout is on a different branch (or local main is ahead of origin/HEAD, e.g. a coworker session's unpushed commit), the hook scans an unrelated commit range and attributes those commits' closing keywords (and possibly digit-runs inside release-please changelog bodies — `bba1965…` contains "1965") to the PR being created.

This is the same HEAD-vs-named-ref class as #1419 (`check-pr-metadata-on-push.sh` read commits from the running shell's HEAD instead of the pushed branch); the fix shape and regression-test pattern from that issue apply directly.

## Fix shape

1. Parse `--head <branch>` / `-H <branch>` from the `gh pr create` command; fall back to HEAD only when absent (creating from the current branch).
2. Scan `merge-base(<head-ref>, origin/<base|HEAD>)..<head-ref>`.
3. Consider anchoring the closing-keyword grep so digit-runs inside commit hashes/URLs in release-please bodies can't match (`#[0-9]+` already requires the `#`, so verify what actually matched — the wrong *range* is the primary bug).
4. Regression test per `.claude/rules/regression-testing.md`: from a checkout on `main` with an unrelated closing-keyword commit in HEAD's range, a `gh pr create --head <clean-branch>` must NOT be blocked; counter-test: a branch whose own commits reference an issue still blocks when the body lacks keywords. Add the Known Regressions row.

## Impact

False-positive hard block on `gh pr create` — cost a retry + branch switch mid merge-recovery. Any workflow that creates PRs for a branch other than the checked-out one (stacked-chain recovery, API-driven multi-repo rollouts per `ci-cd-multirepo.md`) hits it.

laurigates:
## What

`git-plugin/hooks/validate-pr-issue-links.sh` (PreToolUse hook that validates `gh pr create` includes issue-closing keywords) now resolves the branch named by `--head <branch>` / `-H <branch>` when scanning commits for closing keywords, instead of hardcoding `HEAD`.

Fixes #1993

## Why

The hook computed its commit-scan range as `merge-base HEAD origin/HEAD .. HEAD`, hardcoding `HEAD`. When the checkout is on a **different** branch than the PR's `--head` branch — e.g. creating a PR for another branch while on local `main`, or local `main` ahead of `origin/HEAD` with a coworker's unpushed commit — the hook scanned an **unrelated** commit range and attributed those commits' closing keywords to the PR being created. The result was a **false-positive hard block** on `gh pr create`.

Same HEAD-vs-named-ref bug class as #1419 (`check-pr-metadata-on-push.sh`).

## How

- Parse `--head <b>` / `--head=<b>` / `-H <b>` / `-H=<b>` from the command string with `perl` (BSD-grep-safe on macOS), mirroring the existing `--body-file`/`-F` parsing style.
- Resolve the merge-base range against the parsed branch (`HEAD_REF`), falling back to `HEAD` only when no `--head`/`-H` is present (the legitimate "PR from the current branch" case).
- If the named branch does not resolve locally (`git rev-parse --verify` fails), **defer** (exit 0) rather than block — never hard-block on unresolvable input, per `.claude/rules/hook-block-vs-nudge.md`.
- All existing early-exit/allow paths (body-keyword check, command-substitution safety net, body-file-unresolvable defer) are untouched; only the git-log range logic changed.

## Testing

`bash git-plugin/hooks/test-validate-pr-issue-links.sh` → **22 passed, 0 failed**.

New `#1993` regression cases (temp git repo: checkout left on `feature` carrying `Closes #42`, a separate `clean-branch` with no issue refs):

- `--head clean-branch` (and `-H` short alias) with no own issue refs → **allowed** (exit 0). Verified this exact case **blocks (exit 2)** against the pre-fix hook.
- Counter-test: `--head feature` whose own commits reference an issue, body lacking keywords → **still blocks** (exit 2). Proves the guard still fires; the fix didn't disable it.
- `--head no-such-branch` (unresolvable locally) → **defers** (exit 0).

Also green: `shellcheck` (both files clean), `scripts/check-git-sandbox-guards.sh` (`STATUS=OK`), and pre-commit hooks on commit.


laurigates:

PR #2049, opened 2026-07-12
feat(pi): tier-driven installer + local-model recipes

laurigates:
> **Stacked on #2048** (base = `feat/pi-tier-manifest`). Merge #2048 first. Per `git-hazards.md` #3: **retarget this PR to `main` _before_ merging #2048**, or GitHub auto-closes it when the base branch is deleted — and a closed PR whose base is gone cannot be reopened.

## What

Slices 2 + 3 of the pi integration. Consumes the `pi/tiers.yaml` manifest locked by #2048 — the piece that actually gets skills into [pi](https://pi.dev) and points a local model at them.

## The installer — `scripts/install-pi.sh`

Reads the manifest, populates pi's two native scopes **additively** (existing skills under the target are preserved):

| Invocation | Tier | Target |
|---|---|---|
| *(no flags)* | `general` | `~/.pi/agent/skills/` (94 skills) |
| `--category <cat>` | `domain` | `<project>/.pi/skills/` |
| — | `exclude` | never copied |

Cherry-picked plugins install **only** their listed skills — git-plugin lands its core **10**, not all 38. Flags: `--scope global|project`, `--category <cat>`, `--dry-run`, `--list`. `PI_HOME` / `PI_PROJECT_DIR` override targets for hermetic tests. Drops a `.claude-plugins-pi-receipt`; emits the structured `KEY=VALUE` / `STATUS=` block.

## Recipes — `just` group `pi`

| Recipe | Does |
|---|---|
| `pi-tiers` | Print the install plan (what lands where), no writes |
| `install-pi` | general tier → global scope |
| `install-pi-domain <cat>` | that domain → project `.pi/skills/` |
| `serve-pi-model` | `mlx_lm.server --model <pi_model> --port <pi_port>` |
| `setup-pi` | `install-pi`, then print the `models.json` provider block + run command |

`pi_model` / `pi_port` are overridable (`PI_MODEL` / `PI_PORT`).

## Docs — `docs/pi-export.md`

[...]

PR #2056, opened 2026-07-12
feat(scripts): add stranded-work audit for branches that never reached main

laurigates:
## Why

PR #2049 was **auto-closed by GitHub and never merged** — its work sat stranded on a branch until a manual check found it. It was stacked on `feat/pi-tier-manifest`; when that base PR (#2048) was squash-merged and its branch deleted, GitHub closed the child instead of retargeting it. A closed PR whose base ref is gone **cannot be reopened** (`gh pr reopen` → *"Could not open the pull request"*), so the only recovery was a rebase-and-refile (#2054).

A sweep afterwards found **26 dead branches**, two of which carried work that **never had a PR opened at all** — so no `pull_request` event ever fired for them.

## Why a scheduled sweep, not a `pull_request: closed` workflow

The obvious guard is an event handler on auto-close. It can't cover this:

1. **The race is already lost.** By the time the event fires, GitHub has deleted the base ref — the reopen window is gone. An event handler could only file a replacement, exactly like this sweep does.
2. **It needs a caller file in every repo.** `pull_request` events only fire in the repo they happen in. A scheduled sweep queries other repos via the API, so one workflow covers the portfolio (`--repo`).
3. **It is structurally blind to never-PR'd branches** — the shape that produced 2 of the 4 real strands.

The tradeoff is latency (up to a week instead of seconds), which is acceptable for something that otherwise goes unnoticed indefinitely.

## Classification

The **negative** cases are the load-bearing part — a noisy audit trains the reader to ignore it.

| Verdict | Condition | Reported? |
|---|---|---|
| `stranded_autoclose` | closed-unmerged **and** base ref 404s | ✅ |
| `stranded_no_pr` | commits, no PR, older than `--min-age-days` (default 7) | ✅ |
| `landed` | merged PR | silent |
| `closed_deliberate` | closed-unmerged but base ref **alive** | silent |
| `in_flight` | pushed within the grace period | silent |

## Two traps encoded

- **`git cherry`, not `rev-list`/ancestry/tree-containment.** A squash-merge rewrites the SHA, so ancestry reports a fully-landed branch as unmerged. Tree-containment *also* false-negatives once `main` drifts over the same files — it misreported merged branches during the investigation.
- **The base-ref-404 check is the sole discriminator** between an accident and a deliberate close. 11 of the 26 dead branches were deliberate closes (superseded/duplicate). Without this check the audit reports all of them and becomes noise.

The grace period came directly from the live run, which flagged a branch pushed the *same day* as stranded — that was in-flight work, not a strand.

## Verification

- `scripts/tests/test-check-stranded-work.sh` — 19 assertions, hermetic (drives the pure `classify` stage via `--fixture`; no network, no `gh`). Covers all six branch shapes from the real sweep, pinning the three negatives hardest.
- Live run against this repo: `STATUS=PASS`, ~5s.
- `shellcheck`, `actionlint`, `lint-shell-scripts.sh` (0 errors), full pre-commit suite: clean.

Wired into `.pre-commit-config.yaml`; runs weekly via `.github/workflows/stranded-work-audit.yml`, opening one `stranded-work` issue and emitting nothing when clean.

Refs #2049

https://claude.ai/code/session_01Ap1qbGLEkvv1gaG5hSasps


laurigates:

PR #39, opened 2025-12-28
chore(main): release git-plugin 1.1.1

laurigates:
:robot: I have created a release *beep* *boop*
---

### Bug Fixes

* use JSON updater format for extra-files in release-please config ([ec1fac9](https://github.com/laurigates/claude-plugins/commit/ec1fac989e0028874e409dd808c95adff19c6edc))
* use package-relative paths for extra-files in release-please ([e1c9dcf](https://github.com/laurigates/claude-plugins/commit/e1c9dcf9828178b8b74cf3bc8e92778f144c5a5c))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #2074, opened 2026-07-17
fix(configure-plugin): warn that colon-form deny rules widen to longer flags

### LINKED ISSUE #2038: feedback(configure-plugin): claude-security-settings should warn that colon-form deny rules prefix-match longer flags
## Skill

`configure-plugin/skills/claude-security-settings/SKILL.md`

## Category

Enhancement

## Description

During a permission-narrowing pass (loractl PR #39), an implementer agent followed the skill's guidance to build per-subcommand allow rules plus a force-push deny backstop. It wrote the deny rules in colon form:

```json
"deny": ["Bash(git push --force:*)", "Bash(git push -f:*)"]
```

Colon-form patterns prefix-match the raw command string, so `git push --force:*` also matches `git push --force-with-lease …` — silently hard-blocking the safe recovery form that stacked-PR workflows depend on (deny rules cannot be overridden except via bypassPermissions). The skill validated the result without flagging this widening.

## Evidence

- loractl PR https://github.com/laurigates/loractl/pull/39 — the lead reviewer caught it pre-push and corrected to space form (`Bash(git push --force *)`), whose trailing space stops the match at the exact flag.
- The user-global auto-mode rule's canonical deny example already uses space form; the skill did not surface the distinction.

## Suggested Action

Add an explicit caveat (and ideally a validation check) to the skill: flag-scoped deny patterns must use the space-delimited form, not `:*` — colon-form prefix-matching widens `--force` to `--force-with-lease` and any other longer flag sharing the prefix. A one-line lint over proposed deny entries (`ends with a flag + ':*'` → warn) would catch it mechanically.

laurigates:
## What

Adds an explicit caveat to `configure-plugin/skills/claude-security-settings/SKILL.md`: flag-scoped deny rules must use the **space form** (`"Bash(git push --force *)"`), never the colon form (`"Bash(git push --force:*)"`), plus a pointer from the Deny Patterns quick-reference section and audit guidance to flag deny entries ending in `<flag>:*`.

## Why

During a permission-narrowing pass (laurigates/loractl#39), an implementer agent following this skill wrote a force-push deny backstop in colon form. The `:*` suffix was observed prefix-matching the raw command string, so `git push --force:*` also matched `git push --force-with-lease …` — silently hard-blocking the safe recovery form stacked-PR workflows depend on (deny rules cannot be overridden except via `bypassPermissions`). The skill validated the result without flagging the widening.

**Verification against official docs** (per the repo's verify-feedback-against-docs rule): the current permissions doc at code.claude.com/docs/en/permissions states *"The `:*` suffix is an equivalent way to write a trailing wildcard, so `Bash(ls:*)` matches the same commands as `Bash(ls *)`"* — i.e. per current docs both forms enforce the word boundary. However, the Claude Code changelog contains no entry pinning when `:*` gained word-boundary semantics, and the widening was observed live in loractl PR #39. The warning is therefore phrased as **observed behavior** with the issue as evidence, and recommends the space form as the explicit, version-stable choice — it is also the form the permission dialog itself writes when approving a command prefix.

## How

- **SKILL.md**: new `### Flag-Scoped Deny Rules: Use the Space Form` subsection under Wildcard Permission Patterns (correct space-form example first, then the colon-form gotcha as a blockquote callout, then the one-line audit heuristic from the issue); `modified`/`reviewed` dates bumped.
- **Regression check** (`.claude/rules/regression-testing.md` requirement): `scripts/plugin-compliance-check.sh` `check_skill_body()` gains a guard asserting the skill body retains both `--force-with-lease` and `Bash(git push --force *)`. Verified **RED** (guard added first, compliance check failed with 2 errors against the unfixed skill) then **GREEN** (Body column ✅ after the skill edit).
- **Known Regressions table** row added in `.claude/rules/regression-testing.md`.

## Test results

- `bash scripts/plugin-compliance-check.sh configure-plugin` — Body check ✅ (was ❌ pre-fix with the new guard); remaining ⚠️ are pre-existing size/bash-pattern recommendations on unrelated skills.
- `shellcheck scripts/plugin-compliance-check.sh` — exit 0.
- `python3 scripts/audit-skill-descriptions.py --strict-length` — exit 0.
- Full pre-commit suite ran on commit — all hooks passed (README-currency nudge is advisory; the skill's documented purpose/description is unchanged).

Closes #2038


laurigates:

PR #2075, opened 2026-07-17
fix(taskwarrior-plugin): reject unknown reconcile.sh arguments instead of silently ignoring them

### LINKED ISSUE #2057: fix(taskwarrior-plugin): reconcile.sh silently ignores unknown flags, turning a bounded --only-verdicts apply into an unbounded one
## What

`reconcile.sh`'s argument parser ends in a silent catch-all:

```sh
    *) ;;
```

An unrecognised flag is **swallowed, not rejected**. Because `--only-verdicts` is the *only* thing bounding a destructive `--apply`, any version skew between a caller and the script silently converts a bounded apply into an **unbounded** one — it closes every stale verdict, including `pr-closed`, which the flag exists specifically to protect.

## Observed (reproduced this session)

My installed plugin cache was `taskwarrior-plugin` **v1.10.0**, which predates `--only-verdicts` (added in the #1793 work; repo is now v1.13.0). Running the documented bounded form:

```sh
reconcile.sh --all --apply --only-verdicts=pr-merged,issue-closed
```

→ `CLOSED_COUNT=5`, not the expected 4. It closed a `pr-closed` task (`uuid 4bfdb2c9…`, *"fix YAML Syntax Validation failing on PR #738"*) that the flag explicitly promises to keep (`method=keep`). No warning, no error, exit 0. I only caught it because the count was one higher than the dry-run's stale set predicted.

A `pr-closed` task is exactly the case that must not be auto-closed: a PR closed **without** merging may be *stranded work*, not finished work (cf. #2049, closed-unmerged and never landed).

## Why this is more than a stale-cache problem

The fail-open is still present in current `main` (v1.13.0), and `scripts/scheduled-reconcile.sh` — a **LaunchAgent, running unattended** — passes the flag through:

```sh
rc_args+=(--apply --only-verdicts="$APPLY_VERDICTS")
```

So if the scheduled wrapper and `reconcile.sh` ever version-skew (different plugin cache, a partially-updated install, a vendored copy), the unattended sweep performs an unbounded apply with nobody watching. The bound is only as strong as the parser's willingness to reject what it doesn't understand.

This is the `fail fast / let it crash` principle from `code-quality.md`: a destructive operation must not silently ignore the argument that constrains it.

## Suggested fix

Reject unknown arguments instead of ignoring them:

```sh
    *) echo "reconcile.sh: unknown argument: $1" >&2; exit 2 ;;
```

A caller that passes `--only-verdicts` to a version that cannot honour it then **fails loudly** rather than quietly over-closing. Worth auditing the sibling scripts' parsers for the same `*) ;;` shape.

## Verification note

The existing regression test asserts `--only-verdicts` keeps `pr-closed` out of the apply set — and it passes, because it runs against a version that *has* the flag. It cannot catch this, since the failure only appears when the flag is **absent**. A test that passes an unknown flag and asserts a non-zero exit would.

---
Filed from a session where the wrongly-closed task was restored by hand (`task <uuid> modify status:pending`).

laurigates:
## What

`reconcile.sh`'s argument parser ended in a silent `*) ;;` catch-all. Any unrecognised flag was swallowed rather than rejected. This PR replaces the catch-all with a fail-fast reject: the offending argument plus a usage line on stderr, exit 2, before any classification or mutation.

## Why

`--only-verdicts` is the *only* thing bounding a destructive `--apply`. Under caller/script version skew (a stale plugin cache predating the flag, or a typo like `--only-verdictz`), the flag was silently dropped and the bounded apply became an **unbounded** one — observed live closing a `pr-closed` task the flag explicitly promises to keep (`method=keep`), with no warning, no error, exit 0. The unattended `scripts/scheduled-reconcile.sh` LaunchAgent passes the flag through, so a skewed install would over-close with nobody watching. This is the fail-fast / let-it-crash principle: a destructive operation must not silently ignore the argument that constrains it.

Reproduced at HEAD before the fix: `reconcile.sh --all --apply --only-verdictz=...` ran with `MODE=apply` and `ONLY_VERDICTS=` (empty), exit 0.

## How

- `reconcile.sh`: `*) ;;` → error + usage on stderr, `exit 2`; exit-code contract added to the header comment (0 clean, 1 task missing, 2 unknown argument).
- Regression test G in `taskwarrior-plugin/scripts/tests/test-scheduled-reconcile.sh`: passes a misspelled `--only-verdictz` alongside `--apply` against the stubbed store and asserts exit 2, the flag named on stderr, usage printed, and **nothing imported**. Verified RED before the fix (4 failures — the stub captured an unbounded close of all three stale tasks) and GREEN after (32/32; `test-reconcile-refs.sh` 17/17 still green).
- Docs in the same commit: `task-reconcile/REFERENCE.md` exit-code/parallel-safety section, plugin `README.md` scheduled-reconcile section, `SKILL.md` `modified:` bump, and a Known Regressions row in `.claude/rules/regression-testing.md`.

Sibling-parser note (from the issue's "worth auditing" aside): the other `*) ;;` parsers in this plugin (`scheduled-reconcile.sh`, `drain-ghsync-queue.sh`, `release-stale-claims.sh`, `ensure-udas.sh`, `resolve-project.sh`, `detect-gh-mode.sh`, `install-hooks.sh`) fail in the *safe* direction — a swallowed flag there degrades to dry-run/notify-only or a no-op, never a widened destructive apply — so they are deliberately left out of this minimal fix.

Closes #2057


laurigates:

PR #2078, opened 2026-07-17
feat(hooks-plugin): add # allow-timeout escape hatch to the timeout block

### LINKED ISSUE #2041: hooks-plugin: bash-antipatterns hard-blocks all `timeout`, no escape hatch for genuinely-unbounded processes
## What
`hooks-plugin` `bash-antipatterns.sh` (v2.5.13, lines 181-183) **hard-blocks** any command starting with `timeout `:

```sh
if echo "$COMMAND" | grep -Eq '^\s*timeout\s+'; then
    block "REMINDER: The 'timeout' command is usually unnecessary - the Bash tool has its own timeout parameter..."
```

## Why the blanket block is too absolute
The guidance is *usually* right — the Bash tool has its own timeout. But there's a real class the hook doesn't account for: **bounding a process that genuinely never exits on its own** — an interactive REPL, or an MCP **stdio server** launched to warm a build cache (e.g. `uvx --refresh --from git+… <server>`). For those:

- The **Bash-tool timeout** kills the whole tool call and returns an **error/timeout state**, not a clean "the command's own timeout fired, here's the captured output" — so it's a worse signal than `timeout N cmd` producing a clean exit 124.
- `timeout` is the *correct* POSIX tool for "run this, but don't let it run forever."

Encountered this session bounding a stdio MCP server during a `uvx --refresh` cache warm. (In fairness, the hook nudged me to a cleaner `</dev/null` EOF solution that time — but that only works for servers that exit on stdin EOF; `--help`-less, non-EOF-respecting processes have no such out.)

## Suggested change
Make the timeout rule a **soft reminder (non-blocking)** rather than a hard `block`, OR add a narrow allowance — e.g. permit `timeout` when the wrapped command is a known long-runner / server invocation, or honor an inline escape comment (`# allow-timeout`). A hard block with zero override forces awkward rewrites for the legitimate minority case.

## Severity
Low — a nudge that's right most of the time; this is about the missing escape hatch for the correct-use minority, not a correctness bug.

laurigates:
**Stacked on PR1 (#2077)** — base is `fix/bash-antipatterns-false-positives`; retarget to `main` before merging once #2077 lands.

## What

Adds the `# allow-timeout` escape hatch to the bash-antipatterns `timeout` hard block.

## Why

The block is right for the majority case (the Bash tool has its own timeout parameter), but it had zero override for the legitimate minority: bounding a process that genuinely never exits on its own — an interactive REPL, or a stdio MCP server launched to warm a build cache (`uvx --refresh --from git+… <srv>`). For those, the Bash-tool timeout kills the whole tool call and returns an error state, while `timeout N cmd` yields a clean exit 124 with the captured output — a strictly better signal (#2041).

## How

- A trailing `# allow-timeout` comment (with or without the space after `#`) passes the block; everything else still blocks. The comment is a deliberate, visible, per-command opt-in, so the default steer is unchanged.
- The block message now documents the escape hatch.
- Regression tests: bare `timeout` still blocks; both comment forms pass; the message names the hatch (`test-bash-antipatterns.sh` 131 passed / 0 failed on this branch).
- Both hooks-plugin READMEs updated in the same commit (docs-currency).

Closes #2041

### DIFF
--- a/hooks-plugin/README.md
+++ b/hooks-plugin/README.md
@@ -26,7 +26,7 @@ A PreToolUse hook that intercepts Bash commands and blocks those that should use
 | `sed -i` (repo files; `/tmp`/scratch targets exempt, #2052) | Use **Edit** tool instead |
 | `echo > file` | Use **Write** tool instead |
 | `cat > file` | Use **Write** tool instead |
-| `timeout cmd` | Remove timeout (human approval time exceeds it) |
+| `timeout cmd` | Remove timeout (human approval time exceeds it); append `# allow-timeout` for genuinely-unbounded processes (#2041) |
 | `git add -A` / `git add .` | Stage specific files by name instead |
 | Multi-grep test parsing | Use `--reporter=json` instead |
 | `curl \| sh` / `wget \| bash` | Download first, review, then execute |
--- a/hooks-plugin/hooks/README.md
+++ b/hooks-plugin/hooks/README.md
@@ -15,7 +15,7 @@ A PreToolUse hook that intercepts Bash commands and blocks those that should use
 | `sed -i` on repo files (targets under `/tmp`, `/private/tmp`, `/var/folders`, `$TMPDIR` are exempt — #2052) | Use **Edit** tool instead |
 | `echo > file` | Use **Write** tool instead |
 | `cat > file` | Use **Write** tool instead |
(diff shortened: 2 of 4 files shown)


laurigates:

Issue #2058, opened 2026-07-12
fix(hooks-plugin): bash-antipatterns false-positives on heredoc inside command substitution (gh pr create --body)


laurigates:
## What

`hooks-plugin/hooks/bash-antipatterns.sh` blocks a heredoc used inside a **command substitution**, treating it as a file write:

> REMINDER: Use the Write tool instead of 'cat > file' to create files. The Write tool is the proper way to write file contents.

## Repro

The canonical `gh pr create` idiom (straight from this repo's own `CLAUDE.md` git-workflow examples) is blocked:

```sh
gh pr create --title "feat(x): thing" --body "$(cat <<'EOF'
## What
...
EOF
)"
```

There is no `>` and no file target anywhere in the command. `cat` reads a heredoc and writes to **stdout**, which is captured by `$( )` and passed as an argument. Nothing is written to disk.

## Why it's a false positive

`.claude/rules/bash-tool-replacements.md` already documents heredocs and pipelines as exempt:

> **`cat` / `head` / `tail`** — blocked for a plain file read; passes through when the file path is `/dev/stdin`, `/dev/null`, a **here-doc target**, or a pipeline. The hook is checking for *file reads*, not stream handling.

The command-substitution form appears to fall outside whatever exemption covers the bare `cat <<EOF` case, so the documented behaviour and the implemented behaviour disagree.

## Impact

Low severity, but it fires on a **very common** action (opening a PR with a multi-line body) and the suggested remedy is wrong — the Write tool does not help, because the goal isn't to create a file. The workaround is to write the body to a temp file and use `--body-file`, which is strictly more work and leaves a stray file behind. It cost a round-trip this session.

## Suggested fix

Treat `cat` as exempt when its output is consumed by a command substitution or pipe with no redirection to a real (non-`/dev/`) file — i.e. the existing "stream handling, not file writing" test, extended to the `$(cat <<EOF … EOF)` shape.

A regression case for `hooks-plugin/hooks/test-bash-antipatterns.sh`:

```sh
# must be ALLOWED (exit 0) — heredoc into a command substitution, no file written
gh pr create --body "$(cat <<'EOF'
body
EOF
)"

# must still BLOCK (exit 2) — genuine file write
cat > out.md <<'EOF'
body
EOF
```

Issue #2052, opened 2026-07-12
bash-antipatterns: false positives on stdin secret writes, multi-command pipe counts, scratch-file edits, and quoted examples


laurigates:
## What happened

During a `gitops` session (provisioning GSM secrets + OpenTofu work), the `bash-antipatterns` hook fired on **four legitimate commands**, each requiring a rewrite that made the command *worse* or forced a detour. Two of them pushed toward alternatives that were **less safe for secret material**.

### 1. `printf %s '<value>' | gcloud … --data-file=-` → blocked as "use the Write tool"

```
printf %s "$pass" | gcloud secrets versions add <id> --project <p> --data-file=-
```

Hook: *"Use the Write tool instead of 'echo/printf > file'"*.

This is **the recommended, safest way to write a secret** — it pipes the value via **stdin** so it never touches disk or the process table. No file is being created. Complying would be strictly *worse*: it would persist a plaintext credential to the filesystem.

The matcher seems to key on `printf` in a redirect-ish context and misses that the target is a **pipe into a command**, not `> file`. Piping into a command that reads `--data-file=-` / `-` shouldn't trip the file-creation rule.

### 2. Pipe-count heuristic counts across independent commands

```
gcloud secrets versions disable 1 --secret <id> 2>&1 | cat
newlen=$(gcloud secrets versions access latest --secret <id> | wc -c | tr -d ' ')
echo "=== section ==="
grep -E '^Plan:' /tmp/x.log | cat
```

Hook: *"This command has 8 pipes fed from a discouraged head stage (cat/echo/printf…)"*.

The rule's own text says *"A long pipeline of legitimate transforms … is fine — this block only fires on a cat/echo text-scrape head."* But it counted pipes **across several independent commands in one Bash call**, each with its own short, legitimate pipeline. Adding `echo "=== label ==="` progress headers was enough to trip it.

**Suggested fix:** count pipes **per pipeline**, not per Bash invocation; and don't treat a leading `echo "=== label ==="` (a progress/label line, not a data source) as a "discouraged head stage."

### 3. In-place stream edit on a scratch (non-repo) file

Editing a generated throwaway file under the session scratchpad with an in-place stream editor was blocked in favour of the Edit tool. Reasonable **inside a repo** — but here it meant a Read + Edit round-trip on a throwaway file, and the Read then surfaced a **freshly generated WireGuard private key into the transcript**, which the stream edit would not have.

The "safer" tool had the worse security outcome.

[...]

Issue #1900, opened 2026-07-01
bash-antipatterns hook false-positives on `ssh host 'ls ...'` (remote listing, Glob inapplicable)


laurigates:
## What

The `hooks-plugin` `bash-antipatterns.sh` PreToolUse hook fires its `ls → Glob` reminder (and blocks) when `ls` appears **inside a remote SSH command**, e.g.:

```sh
ssh host 'ls -1 /remote/path | grep foo'
ssh host 'ls -lt --time-style=... "$dir"/*.json | head'
```

## Why it's a false positive

The Glob tool operates on the **local** filesystem via the harness. It cannot list files on a remote host reached over SSH. When the `ls` is the payload of an `ssh host '...'` invocation (or otherwise targets a path the local Glob can't see), the suggested replacement is inapplicable — the command has to run remotely. The reminder/block just forces a reword (pipe through `python3`/`printf`, or restructure) with no safety benefit.

Observed repeatedly this session while inspecting a remote ComfyUI install over SSH.

## Suggested fix

In the hook's `ls` detection, skip (or downgrade to non-blocking) when the command is an SSH payload — e.g. the `ls` is an argument to `ssh`/`scp`/`rsync`, or appears inside the quoted remote-command portion of an `ssh <host> '<cmd>'`. A cheap heuristic: if the line matches `^\s*(ssh|rsh|dokku|kubectl exec|docker exec)\b` (a remote-exec prefix), don't apply local-filesystem-tool reminders (`ls`→Glob, `cat`/`head`/`tail`→Read, `find`→Glob) to the remainder.

The same reasoning applies to the `cat`/`head`/`tail`→Read and `find`→Glob reminders inside remote-exec commands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1848, opened 2026-06-28
bash-antipatterns hook false-positives on 'head'/'tail' as a Python variable inside a heredoc


laurigates:
## What

The `bash-antipatterns` PreToolUse hook blocked a `Bash` call whose body was a Python script (passed via `python3 - <<'PY' … PY`) because the Python code contained a variable named `head`:

```python
head = txt[:idx]
```

The hook reported the `head -50 file.md → Read(...)` antipattern and blocked the entire tool call. Renaming the variable (`head` → `out`) made the identical command pass.

## Why it's wrong

The hook scans the raw command string for `head`/`tail` tokens, but here `head` is an *identifier inside a quoted heredoc payload handed to another interpreter* — not a `head(1)` invocation. A bare `head`/`tail` that is a variable assignment (`head=`/`head =`), or inside a single-quoted heredoc (`<<'PY' … PY`) destined for `python3`/`jq`/`awk`, is not the `head <file>` antipattern the rule targets.

## Minimal repro

```sh
python3 - <<'PY'
head = "x"      # blocked: matched as the head-file antipattern
print(head)
PY
```

## Suggested fix

Only match `head`/`tail` in **command position** (line start or after `|`, `&&`, `;`, `$(`, backtick) with a following file/flag argument — not when preceded by `=`/`.`/identifier chars or inside a quoted heredoc body.

## Context

Hit repeatedly this session running multi-repo `python3` heredoc transforms across the comfyui-* packs. Low severity — the workaround (rename) is trivial — but it recurs for any heredoc whose payload legitimately uses `head`/`tail` as identifiers.

PR #2114, opened 2026-07-18
feat(hooks-plugin): structural ast-grep classifier for bash-antipatterns read/write detectors

### LINKED ISSUE #2008: feat(hooks-plugin): ast-grep bash-parse fast path for bash-antipatterns.sh (gated on next false-positive)
## Problem

`hooks-plugin/hooks/bash-antipatterns.sh` (~522 lines, ~44 regex match sites) parses shell commands with regex, and its regression history reads like a changelog of parser bugs:

- fd-redirect vs file-redirect confusion (#1701, #1721, #1722)
- pipeline and heredoc exemption edge cases
- command-substitution scoping (`echo label $(gh … 2>/dev/null)`)
- remote-exec first-token anchoring (#1900)

Every one of these was a case where a real bash parse would have been trivially correct: with tree-sitter-bash (ast-grep supports `--lang bash`), "`cat` reading a plain file" vs "`cat` in a pipeline" vs "`cat <<EOF`" vs "`2>/dev/null` on an echo" are **distinct AST shapes**, not regex edge cases.

## Proposal

Add an ast-grep-based fast path to the hook's command classification:

- When `ast-grep` (or `sg`) is on PATH, parse the command with `--lang bash` and classify structurally (redirect nodes, pipeline nodes, heredoc nodes, command position).
- **Keep the current regex path as the fallback** — hooks fire in every environment (sandboxes, subagents, remote sessions) and must fail open; ast-grep cannot be assumed installed.

## Trade-off / gating condition

The dual-path cost is real: two classifiers to keep in agreement, and the existing regex + regression-test treadmill *is* working (expensively). Per the tier-1 assessment that produced this issue:

> deferred until the next bash-antipatterns false-positive shows up to justify the dual-path cost.

**Do not implement preemptively.** This issue is the parking spot: when the next false-positive regression lands in `bash-antipatterns.sh`, evaluate implementing the structural fast path *instead of* another regex patch.

## Acceptance criteria (when triggered)

- [ ] Structural classification via `ast-grep --lang bash` when available; byte-identical verdicts to the regex path on the full `test-bash-antipatterns.sh` suite
- [ ] Regex fallback preserved and exercised in tests (simulate ast-grep absent)
- [ ] All documented regression cases (#1701, #1721, #1722, #1900) covered by both paths
- [ ] Hook latency stays acceptable (~tens of ms; ast-grep is a single fast binary)

Part of the tree-sitter/ast-grep offload sweep (see tracking issue).

laurigates:
## What

Implements #2008: classify the scoping-sensitive read/write anti-patterns in `bash-antipatterns.sh` **structurally** with `ast-grep --lang bash` (tree-sitter-bash) instead of regex + three pre-stripping passes + the remote-exec guard.

Ported to the AST path (their regex, the `COMMAND_NO_DEVNULL` strip, and the `IS_REMOTE_EXEC` guard **deleted**): `cat`/`head`/`tail` reads, `echo`/`printf`/`cat` writes, `sed -i`, task-output reads. A real parse gives distinct AST node shapes for "cat reading a file" vs "cat in a pipeline" vs "cat <<EOF" vs "cat > file" vs a `cat` *mentioned* inside a string/heredoc body — retiring the false-positive treadmill these detectors carried (#1701, #1721, #1722, #1848, #1900, #2052, #2058).

## No-twin design (per maintainer decision)

The issue's original framing kept "the current regex path as the fallback" — two classifiers in lockstep. That was rejected as unmaintainable. Instead:

- **AST detectors are STYLE nudges that fail open.** When `ast-grep` is absent (sandboxes, subagents) they simply do not fire. Losing a "use Read instead of `cat`" steer where the parser is unavailable costs nothing irreversible. **There is deliberately no regex twin.**
- **SAFETY / correctness blocks stay pure-regex and fire in every context:** `curl|bash`, over-permissive `chmod`, broad `git add`, `git reset --hard`, `git push -u` footgun, block-device writes, fork bombs, the index-lock chain, the grep-chain test-output scrape, `awk`/`cat`-to-commit-file.

Net effect: **less** regex than before — the remote-exec triple-grep and the `/dev/null` sed-strip and five detectors' `grep` invocations are gone, replaced by one `ast-grep` call and a 6-rule inline config.

## The remote-exec guard is gone for free

A read inside `ssh host <<EOF … EOF`, `ssh host 'ls|grep'`, or `kubectl exec … -- cat` is a heredoc-body / string / **argument** node — never a `command_name` — so it is never mis-detected. `cat local.txt && ssh …` still blocks the local `cat` (a real command).

## Verification

- **Byte-identical to the regex path** across the full suite (**141 passed / 0 failed**), and strictly better on the documented false positives.
- Verified **identical output on the CI-pinned `ast-grep 0.43.0`** and local 0.44.1.
- New **fail-open regression section** exercises the no-op path via `CLAUDE_HOOKS_BASH_ANTIPATTERNS_NO_ASTGREP=1` (read/write nudges no-op; safety + grep-chain still block). Suite SKIPs where ast-grep is genuinely absent, mirroring `code-quality`'s rule-project suites (CI installs `@ast-grep/cli`).
- Sibling suites intact: teach 29/0, git-env-isolation 3/0. `shellcheck` + `lint-shell-scripts` clean.
- **Latency ~129ms vs ~155ms** before — the deleted greps/strips more than offset the single ast-grep call.

## Docs (same commit)

hooks-plugin READMEs, `.claude/rules/bash-tool-replacements.md`, and the feature-flags catalog (new `CLAUDE_HOOKS_BASH_ANTIPATTERNS_NO_ASTGREP` opt-in flag).

Closes #2008

https://claude.ai/code/session_01D4648Kh3yEn4GsHg98qmnt

### DIFF
--- a/.claude/rules/bash-tool-replacements.md
+++ b/.claude/rules/bash-tool-replacements.md
@@ -60,9 +60,16 @@ The hook's logic for each:
 
 - **`grep` / `rg`** — never blocked (demoted to the opt-in teach nudge, #1909/#1871). `Grep` is the context-efficient choice when present, but the Bash form always runs.
 - **`ls`** — never blocked (demoted to the opt-in teach nudge, #2036). `Glob` is the context-efficient choice for pattern listings when present, but the Bash form always runs.
-- **`cat` / `head` / `tail`** — blocked for a plain file read; passes through when the file path is `/dev/stdin`, `/dev/null`, a here-doc target, or a pipeline. The hook is checking for *file reads*, not stream handling.
+- **`cat` / `head` / `tail`** — blocked for a plain file read; passes through when the file path is `/dev/stdin`, `/dev/null`, a here-doc target, or a pipeline. The hook is checking for *file reads*, not stream handling. As of #2008 this (and the `echo`/`printf`/`cat`-write and `sed -i` blocks) is decided **structurally** by `ast-grep --lang bash` rather than regex — a real parse tells a command apart from a string/heredoc-body/pipeline/argument, so the block **fails open** (no-op) where `ast-grep` is unavailable. It is a style nudge with no regex twin; the safety blocks (`chmod 777`, `curl|bash`, `git add -A`, …) stay pure-regex and fire everywhere.
 
 ### Remote-exec commands are exempt (issue #1900)
+
+> **As of #2008 the remote-exec guard is only needed by the safety blocks.** The
+> read/write nudges moved to the structural `ast-grep` path, which never treats a
+> read inside `ssh host <<EOF…EOF`, `ssh host 'ls|grep'`, or `kubectl exec … --
+> cat` as a local command in the first place (it is a heredoc-body / string /
+> argument node). The first-token suppression below still governs the pure-regex
+> safety blocks.
 
 When the command's **first token** is a remote-exec launcher — `ssh`, `rsh`,
 `slogin`, `dokku`, `kubectl exec`, `docker exec` (and `podman`/`nerdctl`/`oc`
--- a/hooks-plugin/README.md
+++ b/hooks-plugin/README.md
@@ -34,6 +34,15 @@ A PreToolUse hook that intercepts Bash commands and blocks those that should use
 | `chmod 777` | Use restrictive permissions (755, 644, 600) |
 | Write to block device | Blocked unconditionally |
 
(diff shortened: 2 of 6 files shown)


laurigates:

Issue #2089, opened 2026-07-17
feat(adapters): harness-agnostic skill-discovery core (embedding index + BM25 fallback) with eval gate


laurigates:
## What

Build the harness-agnostic skill-discovery core that pi/OpenCode adapters will bind to, plus the evaluation harness that gates each cutover. Decided in ADR-0022 (adapter over export).

**Scope (plain TS module, no harness imports):**

- **Indexer**: scan the claude-plugins checkout (`.claude-plugin/marketplace.json` + `*/skills/*/SKILL.md`), extract `name`, `description`, plugin, path, and `compatibility` frontmatter. Skip skills marked Claude-Code-only (see the compatibility-sweep issue).
- **Embedding index** over name+description (optionally chunked bodies), built once, cached by content hash. Embedding endpoint configurable (default: local ollama `/api/embed`, e.g. `nomic-embed-text`).
- **BM25/keyword fallback + hybrid rank fusion** so search never hard-fails without an embedding endpoint and vague-intent queries still rank sensibly.
- **Query API**: `search(query, k, filters) -> [{name, description, path, score}]` — returns paths; delivery is the model `read`ing the SKILL.md (both harnesses support this natively).
- **Eval harness**: same task set on the same local quant, baseline (current export: tier-installed for pi / rulesync for OpenCode) vs adapter. Metrics: correct-skill-invocation rate and standing tokens/turn. This is the operational gate for the cutover issues — the export pipelines are not deleted until the adapter meets or beats baseline invocation rate at lower standing cost (`tool-migration-cutover` rule).

## Why

Every installed skill costs ~111 tokens/turn of standing context in pi (uncapped; 401 skills hangs the turn), and OpenCode's `<available_skills>` listing has the same linear shape. The current answer — curation manifests (`pi/tiers.yaml`) and export pipelines (rulesync) — creates per-ecosystem projections that drift from the canonical source. Retrieval replaces curation: all ~400 skills stay at arm's reach while the per-turn cost is bounded by top-k, and the marketplace remains the single source of truth.

## How

Suggested layout: `adapters/core/` in this repo (the marketplace is its data source; a separate repo would re-create the drift problem). Per-harness bindings live in sibling dirs and are tracked in their own issues.

## Related

- ADR-0022 (adapter over export)
- pi binding, OpenCode binding, compatibility sweep, and the two gated cutover issues (linked after creation)

PR #2123, opened 2026-07-20
feat(adapters): OpenCode plugin binding — pull-first skill discovery

### LINKED ISSUE #2091: feat(adapters): OpenCode plugin binding — pull-first skill discovery
## What

OpenCode plugin binding for the skill-discovery core (#2089): the OpenCode adapter that replaces the rulesync export pipeline. Decided in ADR-0022.

**Scope (thin binding, pull-first):**

- **Pull** (stable API): register `search_skills` via the plugin `tool: {}` surface; model `read`s returned SKILL.md paths. OpenCode reads Claude Code SKILL.md format natively (Agent Skills standard), so no content conversion is needed.
- **Suppress the native listing**: `tools: { skill: false }` (or `permission.skill` deny patterns / the `tool.definition` hook, which rewrites the `skill` tool description where `<available_skills>` lives).
- **Push (behind a version check)**: `experimental.chat.system.transform` / `experimental.chat.messages.transform` for per-turn top-k injection. These hooks are `experimental.*`-prefixed and may break across OpenCode releases — ship pull-first, add push guarded, degrade gracefully.
- **Config**: per-project scoping via `opencode.json`.

## Why

OpenCode's native `<available_skills>` listing has the same uncapped linear cost as pi's. More importantly, OpenCode now supports the Agent Skills standard natively — including discovery from `.claude/skills/` paths — which makes the rulesync conversion pipeline (`docs/opencode-export.md`) largely obsolete. The adapter replaces it with the same core pi uses, validating the harness-agnostic seam.

## Notes

- OpenCode enforces `name` == parent directory (pi doesn't). Bypassing the native `skill` tool avoids the old `rewrite-skill-name-to-dir` normalization entirely; only a pinned native-discovery set would resurface it.
- `compatibility` frontmatter is recognized by OpenCode per the same spec — the sweep issue serves both harnesses.

## Acceptance

Eval gate from #2089 on OpenCode: adapter meets or beats the rulesync-exported baseline's correct-skill-invocation rate at lower standing tokens/turn. Passing the gate unblocks the rulesync-retirement issue.

## Related

- ADR-0022, #2089 (core), #2090 (pi binding), and the gated rulesync cutover issue

laurigates:
## What

The OpenCode plugin binding (`adapters/opencode/`) over the discovery core landed in #2089. Implements ADR-0022 §2's OpenCode half.

- **Pull** (stable API): `search_skills` via the plugin `tool: {}` surface (zod args), returning top-k `{name, description, path}`; OpenCode reads Claude Code SKILL.md natively, so no content conversion.
- **Suppression**: primary is consumer-side `"permission": {"skill": "deny"}` (stable config; removes both the `<available_skills>` system-prompt block and the native tool), with a defensive in-plugin strip as backup.
- **Push** (guarded): `experimental.chat.system.transform` injects pins + ranked top-k; declared unconditionally since unknown hooks are silently never called on older versions, plus a non-gating `/global/health` version log. Ranking input is captured from `experimental.chat.messages.transform` (one-turn-stale caveat documented in-code).

### Two blockers caught in review

Both were found by adversarial review against upstream source, not by tests:

1. **Reassigning `output.system` silently no-ops.** Upstream reads its local `system` array after `plugin.trigger` with no copy-back, so both the strip and the push were being discarded. Now mutates in place (`length = 0; push(...)`), and the test harness returns the *original* array reference so the suite pins the in-place contract.
2. **The whole-element strip would have deleted the entire system prompt.** Upstream pre-joins the whole prompt into a single array element, so dropping any element containing `<available_skills>` removes everything — in exactly the scenario the strip exists for. Replaced with substring removal (`stripAvailableSkillsBlocks` in `core/render.ts`, shared with the pi binding's strip semantics); elements are only removed if they become empty.

Plus 5 lower-severity fixes (phantom `@opencode-ai/sdk` test import, empty-block push on a first turn with no pins, unfiltered `ignored` text parts, missing import-restriction coverage for `opencode/`, fixture tmpdir leak) in `e15d3a69`.

## Verification

- `bun test`: 121 pass / 0 fail; `tsc --noEmit`, `biome ci`: clean
- Smoke eval still `=== GATE === STATUS=PASS` (in-tier hit@5 0.370, excluded 0.694, ADAPTER_TOKENS=608)
- Import-restriction test now covers `opencode/` (`node:*` / relative / `@opencode-ai/plugin` only)

Closes #2091

https://claude.ai/code/session_01F5iwac3d7LWd8Jr9iBArXv

### DIFF
--- a/adapters/README.md
+++ b/adapters/README.md
@@ -9,11 +9,10 @@ always the model **reading the SKILL.md path** — skills are never copied.
 Architecture: [`docs/adrs/0022-adapter-over-export-for-foreign-harnesses.md`](../docs/adrs/0022-adapter-over-export-for-foreign-harnesses.md).
 `core/` + `eval/` + tests landed in
 [#2089](https://github.com/laurigates/claude-plugins/issues/2089); the pi
-binding (`pi/`) landed in
-[#2090](https://github.com/laurigates/claude-plugins/issues/2090). The
-OpenCode binding
-([#2091](https://github.com/laurigates/claude-plugins/issues/2091)) is
-documented ahead so the shape is stable, but its files do not exist yet.
+binding (`pi/`) in
+[#2090](https://github.com/laurigates/claude-plugins/issues/2090); the
+OpenCode binding (`opencode/`) in
+[#2091](https://github.com/laurigates/claude-plugins/issues/2091).
 
 ## Layout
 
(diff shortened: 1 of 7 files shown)


laurigates:

Issue #2093, opened 2026-07-17
chore(pi): remove tier system after adapter eval gate passes


laurigates:
## What

Remove the pi tier system once the pi adapter (#2090) passes its eval gate. Gated cutover per ADR-0022 and the `tool-migration-cutover` rule: **do not merge before the gate** — the adapter must be observed meeting/beating the tier-installed baseline (correct-skill-invocation rate, lower standing tokens/turn) on a local quant.

**Complete-sweep checklist (per `incomplete-migration-leftover-tooling`, find every wiring):**

- [ ] `pi/tiers.yaml` (migrate any still-relevant `exclude` reasons into `compatibility` frontmatter first — #2092)
- [ ] `scripts/install-pi.sh`
- [ ] `scripts/check-pi-tiers.sh` + its `.pre-commit-config.yaml` entry + the `Plugin: PR checks` workflow step
- [ ] justfile `pi` group recipes: `pi-tiers`, `install-pi`, `install-pi-domain` (keep `serve-pi-model`; rewrite `setup-pi` for the adapter flow)
- [ ] `docs/pi-export.md` — rewrite around the adapter; delete the tier/installer sections and the "Out of scope (deferred)" entry this work fulfilled
- [ ] `CLAUDE.md` "Local-model export" table row for pi
- [ ] Grep the repo for remaining `tiers.yaml` / `install-pi` references

## Why

Overlapping solutions (tier manifest + adapter) are worse than either alone: two sources of curation truth that drift. Once the adapter is operational, the manifest is dead weight — but removing it before the gate would leave pi with no working discovery (`tool-migration-cutover`: config presence is not operation).

## Related

ADR-0022, #2089 (eval harness), #2090 (the gate), #2092 (must land first)

laurigates:
&lt;!-- routine:needs-decision v1 --&gt;
@laurigates

Same shape as #2094's sibling decision: "do not merge before the gate" — the pi adapter (#2090) must be observed meeting/beating the tier-installed baseline's correct-skill-invocation rate at lower standing tokens/turn, on a local quant. #2090 is closed/merged (PR #2122), but neither its thread nor #2089's records an explicit gate-pass measurement — only that the adapter shipped. Note #2092 ("must land first") is also already merged, so that prerequisite is clear either way.

1. Has the eval gate actually been measured and passed?
   - **A. Yes, informally verified** — go ahead and remove the tier system now per the issue's complete-sweep checklist.
[...]

PR #2132, opened 2026-07-23
docs(adapters): pi-adapter trial recipes + zero-config trial path

laurigates:
Makes the ADR-0022 pi adapter trivial to try, and documents it where the reader lands.

## What's new

Two `just` recipes in the `adapters` group:

| Recipe | Does | Cost |
|---|---|---|
| `just pi-adapter-check` | Verify prereqs: extension present, `node_modules` populated, ollama reachable with `nomic-embed-text` | none — deterministic, no model call |
| `just pi-adapter [args]` | Launch pi with the extension via `-e`, **zero config mutated** | one pi session |

## Verified end-to-end (pi 0.80.7)

`just pi-adapter -p "how many skills are in your system prompt?"` → **5** (the ranked top-k), versus **92** for plain `pi`. The uncapped `<available_skills>` listing is replaced in place and `search_skills` is registered for pull.

## Docs (link, don't duplicate)

- `adapters/README.md` § pi — the zero-config trial up front, plus a "you do **not** need to uninstall the native tier skills first" note (the binding strips `<available_skills>` regardless).
- `docs/pi-export.md` — the incumbent tier installer gains a successor pointer.
- `CLAUDE.md` — the local-model export table gains an adapter row.

No source change; `adapters` scope bumps no release-please package. The tier system stays fully operational until #2093 retires it.

Refs #2093

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -261,10 +261,11 @@ Run this marketplace's skills inside local-model coding agents. Two targets, two
 
 | Target | Source of truth | Doc | Recipes |
 |--------|-----------------|-----|---------|
-| **pi** (pi.dev) | `pi/tiers.yaml` (tier classification, enforced by `scripts/check-pi-tiers.sh`) | [`docs/pi-export.md`](docs/pi-export.md) | `just check-pi-tiers`, `pi-tiers`, `install-pi`, `install-pi-domain <cat>`, `serve-pi-model`, `setup-pi` (group `pi`) |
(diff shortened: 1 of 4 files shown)


laurigates:
Added `pi-adapter-register` / `pi-adapter-unregister` (commit 3288ed0e) — closes the last manual step of hand-editing `~/.pi/agent/settings.json`.

Key finding while building it: pi loads a local extension from the **`extensions`** array, but `pi install`/`pi remove`/`pi list` manage a **separate `packages`** array. Verified on pi 0.80.7 that an `extensions` entry loads persistently (system prompt → 5 skills, no `-e`) while `pi list` shows only `packages` — so the native `pi install` targets the wrong key here, and the recipe edits `extensions` directly. That's why `pi list` won't display the adapter; it's expected. README and CLAUDE.md now record this.

Recipes are idempotent, non-clobbering, mode-600-preserving, and `PI_SETTINGS`-overridable for project scope; round-trip tested against a temp file plus an idempotency check against live global settings.

Issue #1326, opened 2026-05-14
refactor(agent-patterns-plugin): extract parallel-agent-dispatch reference material to REFERENCE.md


laurigates:
## Context

After PR #1325 (Wave 3b of the 2026-05-14 feedback-promotion sweep) added the `Refactor-brief template` and `Reviewer-agent verification` subsections, `agent-patterns-plugin/skills/parallel-agent-dispatch/SKILL.md` is at **exactly 500 lines** — the local hard ceiling enforced by `scripts/plugin-compliance-check.sh`'s `check_skill_size()` (per `.claude/rules/skill-quality.md`: WARN at 251–500, ERROR at >500).

The compliance bot already flags it on every PR touching the file. The bot's recommendation is right: extract reference material to a sibling REFERENCE.md.

## What needs to happen

Any future addition to this skill will trip the ERROR ceiling. The next feedback-promotion or new failure-mode entry cannot land additively without an extraction pass first.

## Candidate sections to extract

Read the post-#1325 SKILL.md end-to-end first, but likely candidates (from the existing structure):

- **`## Failure modes → schema field` table** — long reference table; pure lookup, doesn't drive the in-skill workflow.
- **`### When the agent stalled at commit / push` salvage routine** (added by #1317) — a worked recovery procedure that could live next to the in-skill cross-reference.
- **Specific worked examples** under "Handling a Missing Return" — keep one canonical example in SKILL.md; move the rest.
- **`## Concurrent rate-limit risk` recovery-dispatch routine** (added by #1317) — keep the cap-and-cite in SKILL.md, move the recovery procedure to REFERENCE.md.

Goal: SKILL.md back to ≤ ~350 lines with the operational workflow intact; REFERENCE.md carries the lookup tables and worked examples.

## Constraints

- Per `.claude/rules/docs-currency.md`, the extraction lands in the same commit as any cross-references that point to REFERENCE.md.
- Preserve all `Use when...` triggers, sibling cross-refs (`agent-patterns-plugin:parallel-safe-queries`, `.claude/rules/skill-fork-context.md`), and the bumped `modified:` / `reviewed:` dates.
- The Anthropic ideal is 200 lines; target a 200–300 line operational body in SKILL.md.

## Evidence

- Compliance bot warnings on PRs #1317 and #1325 (this PR's hit was the trigger).
- `.claude/rules/skill-quality.md` size band rationale.
- Bot output trail across the recent W20 PR series.

Issue #1156, opened 2026-04-25
docs: standardise When to Use This Skill tables across all plugins


laurigates:
## Context

The skill quality rule at `.claude/rules/skill-quality.md` requires every skill's `SKILL.md` (or `skill.md`) to include a `## When to Use This Skill` section immediately after the title, with a 2-column table contrasting the skill against its real siblings.

A repo-wide audit on 2026-04-25 found **162 of 298 skills (54%)** missing this required section, distributed across 28 plugins.

This is an umbrella tracking issue. Track B of Wave 3 dispatches one PR per plugin so each PR is mechanically scoped (table format only, no behaviour changes) and reviewable.

## Per-plugin violation counts

| Plugin | Missing | Total |
|--------|---------|-------|
| git-plugin | 22 | 35 |
| blueprint-plugin | 19 | 33 |
| obsidian-plugin | 13 | 13 |
| testing-plugin | 10 | 16 |
| python-plugin | 10 | 17 |
| tools-plugin | 9 | 14 |
| code-quality-plugin | 9 | 17 |
| kubernetes-plugin | 8 | 8 |
| configure-plugin | 8 | 46 |
| terraform-plugin | 6 | 6 |
| networking-plugin | 6 | 6 |
| container-plugin | 6 | 9 |
| typescript-plugin | 5 | 17 |
| project-plugin | 5 | 7 |
| github-actions-plugin | 5 | 9 |
| finops-plugin | 5 | 7 |
| agent-patterns-plugin | 5 | 11 |
| taskwarrior-plugin | 4 | 4 |
| langchain-plugin | 4 | 4 |
| rust-plugin | 3 | 5 |
| documentation-plugin | 3 | 5 |
| communication-plugin | 2 | 2 |
| home-assistant-plugin | 1 | 4 |
| component-patterns-plugin | 1 | 2 |
| blog-plugin | 1 | 2 |
| api-plugin | 1 | 2 |
| agents-plugin | 1 | 1 |
| accessibility-plugin | 1 | 2 |

## Required format (per `.claude/rules/skill-quality.md`)

\`\`\`markdown
## When to Use This Skill

| Use this skill when... | Use X instead when... |
|------------------------|----------------------|
| Specific scenario A    | Alternative scenario |
| Specific scenario B    | Alternative scenario |
\`\`\`

The right-hand contrast column must reference **real sibling skills** in the same plugin — not placeholders like \`X\`.

## Per-plugin PRs

[...]

PR #2139, opened 2026-07-26
feat(evaluate-plugin): measure the marketplace against the Claude 5 context-engineering rules

laurigates:
## What

Evaluates this marketplace against Anthropic's [**The new rules of context engineering for Claude 5 generation models**](https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models), and ships the tooling to re-run it on the next model release.

Three deliverables:

1. **Channel M scanner** — `check-context-engineering.py`, deterministic C1–C6 proxies over all 407 skills plus the always-loaded surface. Byte-identical across runs.
2. **Channel J benchmark** — `docs/benchmarks/2026-07-context-engineering/`: a rubric frozen *before* judging, 10 stratified units × 2 independent judges, full audit trail.
3. **Durable tooling** — `/evaluate:context-engineering`, the `.claude/rules/context-engineering.md` standing position, and a CI ratchet on the always-loaded budget.

## The finding

**The corpus is not over-constrained — it is over-duplicated and over-resident.** That inverts what the post's lead shift (rules → judgment) predicts for a marketplace written under the old rules, and it changes the remediation order.

| Dim | Judged mean | Reading |
|---|---|---|
| C1 rules → judgment | **3.9** | Healthy — only 9 of 407 skills are constraint-dense |
| C2 examples → interface | **3.86** | Healthy |
| C6 specs → rich references | **3.15** | Competent |
| C3 progressive disclosure | **2.9** | 242 skills are one flat file; **0** `references/` directories |
| C4 single source of truth | **2.65** | One cross-plugin pair shares **824 shingles** — and has drifted |
| C5 always-loaded budget | **1.67** | **~24,600 tokens injected into every session** |

Two numbers worth the click: the mandated `## When to Use This Skill` table costs **~70,000 tokens (9.2% of all skill body mass)** for guidance the post locates in the `description` field; and of 1,241 constraint markers, only **11.8%** sit next to a safety word — yet C1 still scores 3.9, because the ones that remain are the ones that matter.

So the order is: scope-or-promote the unscoped rules → dedupe → split. Cutting skill bodies is where the big raw numbers are and the smallest real saving, because a skill body is on-demand and an unscoped rule is every-turn.

## Evidence discipline

Carried over from the 2026-07 marketplace benchmark:

- **150/150 evidence quotes mechanically verified** against the files on disk, 0 fabricated, 0 extreme scores without evidence (`quote_check.py`).
- Judges never saw Channel M's numbers; the two channels are reported separately and never blended.
- The house conventions under test are quarantined in the rubric's exclusion list — this run tests them, so they cannot also be the yardstick.
- 1 contested pair out of 55.
- Unlike the prior benchmark, the reproduction scripts are **committed**.

`§6` publishes what this run does *not* establish: n=2 judges cannot separate agreement from correlation, 397 skills were never judged, the duplication finding is one-sided by construction, two rubric anchors rest on paraphrase, and **nothing is ablated**.

## Dogfooding notes

- `.claude/rules/context-engineering.md` is **path-scoped**, so it adds nothing to the always-loaded surface it criticises. Verified: the C5 total was unchanged by adding it.
- The scanner lives in the skill that owns it; `scripts/check-context-engineering.py` is a shim, so there is one implementation rather than two — the C4 property it measures.
- That move surfaced a real bug it caused: the default `--project-dir` was derived from `__file__`, so once the scanner lived inside a skill it scanned its own directory and reported `SKILL_COUNT=0` — a silent "nothing to fix". Fixed, with a regression assertion.

## Verification

- `check-context-engineering.py --json | sha256sum` stable across runs, and under `PYTHONHASHSEED=random` (self-test asserts both).
- Self-test: 27/27 pass. `check-docs-index.sh` OK. `lint-context-commands.sh` OK. `check-git-sandbox-guards.sh` OK. `lint-shell-scripts.sh` clean.
- Ratchet (`--strict`) passes at 98,508 / 100,000 chars.

**Two pre-existing failures found while verifying and deliberately not fixed here** — both reproduce on `main`, and each wants its own PR:

- `agent-patterns-plugin/parallel-agent-dispatch` is 26,024 chars, 24 over the house ceiling (`lint-compliance` ERROR). It is also this benchmark's top C3 finding, so the fix and the finding agree.
- `project-plugin/project-init` Context command exits 1 in a bare sandbox (`check-context-command-execution.sh` FAIL).

## Follow-up (not this PR)

Ablation. The post's claim is an ablation result; ours are judgments. Turning them into measurements means authoring `evals.json` for the top-ranked units and re-measuring after each cut — blocked on coverage, since **1 of 407** skills has one today. Every finding in the report is labelled **proven** or **proxy-only** so that work has a queue.

Applying the cuts is also out of scope: each lands as its own reviewable PR, with always-loaded migrations going through `meta-context-diet`'s per-candidate confirmation gate.

https://claude.ai/code/session_01YZscmc38rV24NBvBnPnCPE

---
_Generated by [Claude Code](https://claude.ai/code/session_01YZscmc38rV24NBvBnPnCPE)_


laurigates:
Follow-up issues opened from this benchmark's findings:

| Issue | Finding | Why it ranks there |
|---|---|---|
| #2140 | **F1** — trim the always-loaded surface | Highest value. ~24,600 tokens on *every* turn; C5 judged 1.67, the worst dimension |
| #2141 | **F2** — apply the demoted body-section mandates | ~103K tokens of required boilerplate (13.5% of all body mass). Partially reverses #1156 |
| #2142 | **F3** — dedupe `comfy-node` ↔ `foundryvtt-module` | 824 shared shingles, and the copies have already drifted — a correctness problem, not just cost |
| #2143 | **F4** — multi-file `references/` splitting | 242 flat files, 0 `references/` dirs. Starts with `parallel-agent-dispatch`, which also clears its pre-existing size-ceiling ERROR. Subsumes the regrown part of #1326 |
| #2144 | Ablation | Turns these judgments into measurements. Blocked on eval coverage (1 of 407 skills) |
| #2145 | `project-init` Context abort | Unrelated pre-existing bug found while verifying; filed separately |

Two sequencing notes for whoever picks these up:

- **#2143 must wait for this PR to merge.** `quote_check.py` verifies 150 evidence quotes against the files as they currently stand; editing `parallel-agent-dispatch/SKILL.md` first would invalidate the audit trail.
- **#2141 should loosen the lint before sweeping**, or every per-plugin PR will fight `check_skill_when_to_use()`.

#2144 is the one that would make the rest falsifiable — everything here is currently a judgment, not a measurement.

---

Issue #2143, opened 2026-07-27
refactor(skills): adopt multi-file references/ splitting — 242 skills are a single flat file, 0 use references/


laurigates:
Follow-up to #2139 (context-engineering benchmark), finding **F4**. Subsumes the still-unfixed part of #1326.

## The measurement

| Signal | Value |
|---|---|
| Skills that are a single flat file (no supporting files at all) | **242 of 407** |
| Skills with a supporting file (2-level disclosure) | 145 |
| Skills reaching body → references → scripts (3-level) | 20 |
| `REFERENCE.md` sidecars | 129 |
| **`references/` directories** | **0** |
| Bodies >10,000 chars with zero supporting files | 53 |

Judged C3 mean: **2.9**.

The post asks for something the corpus has never done: *"For long skills, try and use progressive disclosure as much as possible — **divide it into many files and split them out**."* Our house pattern is a *single* `REFERENCE.md`, which defers content but doesn't let a scoped run load only what it needs.

Three judges landed on the same fix independently, for three unrelated skills — split the sidecar **by the path that needs it**:

> Split the single ~13KB sidecar into per-layer references (dependency, SAST, secrets, policy, CI workflow) so a `--type secrets` run loads only secret-detection templates instead of every template in the skill.

## Start here: `parallel-agent-dispatch`

`agent-patterns-plugin/skills/parallel-agent-dispatch/SKILL.md` is **26,024 chars — 24 over the 26,000 house ceiling**, a live `plugin-compliance-check.sh` ERROR that predates #2139 and reproduces on `main`. It is also the benchmark's largest skill body and its sharpest C3 case, so the lint failure and the finding agree on the same fix.

#1326 already extracted reference material to `REFERENCE.md` and was closed — but the body has since grown back over the ceiling. That is the argument for the *multi-file* shape rather than another one-off extraction: a single sidecar has no natural place to put the next case study, so it lands in the body.

The judged target:

> Split the single `REFERENCE.md` into path-specific files and move the rare worktree-hazard case studies (#1838 nested-repo, #1692 `GIT_DIR` export, #1480 cwd-reset) out of the entry point into them, leaving one-line pointers behind.

Those hazards are individually valuable and collectively rare — exactly the shape that belongs behind a path-specific reference.

[...]

PR #2152, opened 2026-07-27
fix(plugin-infrastructure): gate skill size on characters, not bytes

### LINKED ISSUE #2135: fix: skill-size gate measures bytes but documents characters — a file can fail on multi-byte UTF-8 alone
## What

`plugin-compliance-check.sh` `check_skill_size()` gates on `wc -c`, which counts **bytes**. `.claude/rules/skill-quality.md` documents the gate as measuring **characters**:

> We therefore gate on **characters** (`wc -c`), the cheapest tight proxy for tokens (~4 chars/token for English prose)

For an all-ASCII body the two are identical, so the conflation is invisible. For a body containing em-dashes, arrows, or box-drawing characters — which is most skills in this repo — bytes exceed characters, and a file can cross the ERROR ceiling on encoding overhead alone.

## Evidence

`agent-patterns-plugin/skills/parallel-agent-dispatch/SKILL.md` sits exactly in the ambiguity window:

| Measure | Value | vs 26,000 ceiling |
|---|---|---|
| `wc -c` (bytes) — what the gate uses | 26,024 | **24 over → ERROR** |
| UTF-8 decoded characters — what the rule says | 25,850 | 150 under → passes |

The 174-byte gap is fully accounted for: 87 non-ASCII characters, all 3-byte UTF-8, contributing `87 × 2 = 174` extra bytes.

```
$ wc -c < .../parallel-agent-dispatch/SKILL.md      # 26024  (bytes)
$ python3 -c "print(len(open(F).read()))"           # 25850  (characters)
$ python3 -c "print(sum(1 for c in open(F).read() if ord(c)>127))"   # 87
```

Note `wc -m` also returns 26,024 in a C/POSIX locale — it only counts characters under a UTF-8 locale, so `wc -m` is not a drop-in fix unless `LC_ALL` is pinned.

## Why it matters

The file currently fails a whole-repo compliance run. It passes PR CI only because `plugin-pr-checks.yml` scopes the scan to changed plugins (`plugin-compliance-check.sh $PLUGINS`), so the ERROR surfaces only when someone touches `agent-patterns-plugin` or runs the check unscoped. That's a latent trip-wire on an unrelated PR.

More broadly: the threshold was calibrated as a token proxy at ~4 chars/token. Measuring bytes silently tightens the effective budget for any skill using non-ASCII punctuation, and this repo uses em-dashes heavily. The tightening is unintentional and unevenly distributed.

## Which unit is actually right

Worth deciding deliberately rather than by accident:

- **Characters** match the documented rationale and the `chars / 4` token estimate the same rule uses for the description budget.
- **Bytes** are arguably a *better* token proxy in some cases (a 3-byte em-dash is usually still 1 token, but non-Latin scripts tokenize far worse per character). If bytes are preferred, the rule text and the token-estimate arithmetic should say so.

Either resolution is fine; the current state — gate says bytes, docs say characters, threshold calibrated as characters — is the defect.

## Suggested fix

1. Pick a unit and make `check_skill_size()` and `skill-quality.md` agree. For characters, count with a locale-independent method (`python3 -c` or `LC_ALL=C.UTF-8 wc -m`) rather than bare `wc -m`.
2. Extend `scripts/tests/test-plugin-compliance-skill-size.sh` with a fixture whose byte and character counts straddle the ceiling — that is the semantic invariant, and no current fixture exercises it (all existing boundary fixtures are presumably ASCII).
3. Independently of the unit: `parallel-agent-dispatch/SKILL.md` is a 26 KB body next to an already-30 KB `REFERENCE.md`, and `## The Three Pillars` alone is 12,399 chars — nearly half the body. It's a reasonable extraction target on its own merits, not only to clear the ceiling.

## Provenance

Surfaced while verifying #2133 (blueprint-plugin #2128 + #2129). The file is untouched by that PR and byte-identical at its base commit; it was flagged there as a pre-existing item and is filed here so it isn't lost.

laurigates:
## What

`scripts/plugin-compliance-check.sh` `check_skill_size()` gated on `wc -c` (**bytes**) while `.claude/rules/skill-quality.md` documented the gate as measuring **characters** and calibrated the 26,000 ceiling as a `chars / 4` token proxy. This makes the unit and the implementation agree on **characters**, and adds the regression fixture that can actually detect the difference.

## Reproduction

Confirmed on `agent-patterns-plugin/skills/parallel-agent-dispatch/SKILL.md` before changing anything:

| Measure | Value | vs 26,000 ceiling |
|---|---|---|
| `wc -c` (bytes) — what the gate used | 26,024 | 24 over → **ERROR** |
| decoded UTF-8 characters — what the rule says | 25,850 | 150 under → passes |
| non-ASCII characters | 87 | all 3-byte → `87 × 2 = 174` extra bytes |

The 174-byte gap is fully accounted for. For an all-ASCII body bytes and characters are identical, which is why the conflation stayed invisible.

## Unit decision: characters

Characters match the documented rationale and the `chars / 4` token arithmetic the same rule already uses for the description budget. Gating on bytes silently tightens the effective budget by encoding overhead, unevenly distributed — a skill is penalised for using em-dashes.

## Locale handling

Counting uses `python3`, not `wc`:

```
char_count=$(python3 -c 'import sys; print(len(open(sys.argv[1], encoding="utf-8", errors="surrogateescape").read()))' "$skill_file")
```

Bare `wc -m` is **not** a drop-in fix — it counts characters only under a UTF-8 locale and silently returns bytes under `C`/`POSIX`, the same silent-wrong-answer failure mode being fixed. Verified on this machine: `LC_ALL=C wc -m` → 26,024 (bytes); `LC_ALL=C.UTF-8 wc -m` → 25,850. The default locale in the agent shell returned **26,024**, so a bare `wc -m` would not even have fixed it here.

`python3` was chosen over pinning `LC_ALL=C.UTF-8` because it is locale-independent by construction, is already a hard dependency of this script (`check_skill_frontmatter` parses YAML with it), and cannot degrade silently if a locale is missing. `errors="surrogateescape"` keeps a malformed byte as one character rather than raising. Cost: ~7s added to a ~100s whole-repo run (408 skill files); the PR CI path is plugin-scoped and pays far less.

## Before / after — whole-repo `bash scripts/plugin-compliance-check.sh` (unscoped)

| | ERROR | WARN | exit |
|---|---|---|---|
| before | 1 | 90 | 1 |
| after | 0 | 90 | 0 |

The single before-ERROR was the one under investigation:

```
- ❌ agent-patterns-plugin/parallel-agent-dispatch: SKILL.md is 26024 chars (~6506 tokens, >26000 ceiling)
```

After, the `### Issues Found` section is gone entirely and the skill reports as a WARN at its true size:

```
- ⚠️ agent-patterns-plugin/parallel-agent-dispatch: SKILL.md is 25850 chars (~6462 tokens, >10000)
```

Exactly two skills change severity band; every other count shifts down only by its own encoding overhead:

- `agent-patterns-plugin/parallel-agent-dispatch`: ERROR → WARN (the fix's purpose)
- `bevy-plugin/bevy-game-engine`: WARN → OK — 9,929 characters / 10,083 bytes, i.e. it was over the 10,000 WARN threshold on 154 bytes of encoding overhead alone

## Regression coverage

Every pre-existing fixture in `scripts/tests/test-plugin-compliance-skill-size.sh` is ASCII, where bytes == characters, so none could exercise this invariant. Added a `straddle` fixture — ASCII filler plus 200 × U+2014 EM DASH — whose **byte count sits above the ceiling while its character count sits below**:

```
PASS: straddle fixture is valid (26258 bytes > 26000 >= 25858 chars)
PASS: bytes-over/chars-under body does NOT trip ERROR (#2135)
PASS: size line reports the CHARACTER count (#2135)
PASS: size line does not report the BYTE count (#2135)
```

Three properties are asserted, not just "it passes": it must not trip ERROR, the reported number must equal the character count, and it must never equal the byte count. A fixture-validity guard asserts the straddle property itself, so the test cannot silently degrade into a no-op if the header length or filler changes.

**Verified the gate is real** by temporarily reverting `check_skill_size()` to `wc -c` and re-running — all three new assertions fail:

```
FAIL: bytes-over/chars-under body does NOT trip ERROR (#2135)
FAIL: size line reports the CHARACTER count (#2135)
FAIL: size line does not report the BYTE count (#2135)
---
passed: 5, failed: 3
```

The pre-existing `toobig` fixture (30,000 chars) still covers a body genuinely over the ceiling in characters, and still FAILs. Full suite: **8 passed, 0 failed**.

## Verification run

| Check | Result |
|---|---|
| `bash scripts/tests/test-plugin-compliance-skill-size.sh` | 8 passed, 0 failed |
| same test vs. reverted `wc -c` implementation | 5 passed, 3 failed (gate confirmed real) |
| `bash scripts/plugin-compliance-check.sh` (unscoped) | exit 0 (was exit 1) |
| `bash scripts/lint-shell-scripts.sh` | 0 errors (9 pre-existing `macos-plugin` warnings, untouched) |
| `shellcheck` on both edited scripts | clean |
| `bash scripts/check-docs-index.sh` | `STATUS=OK ISSUE_COUNT=0` |
| `bash scripts/check-driver-freshness.sh` | OK |
| pre-commit (all matched hooks on commit) | Passed |

## Out of scope

Suggestion 3 in the issue — extracting `## The Three Pillars` out of `parallel-agent-dispatch/SKILL.md` — belongs to **#2143** and is **not** done here. That file is deliberately untouched (`git diff --stat` against it is empty); another agent may be working on it. It remains a reasonable extraction target on its own merits now that it is no longer load-bearing for the ceiling.

## Docs currency

`.claude/rules/skill-quality.md` lands in the same commit as the code, per `.claude/rules/docs-currency.md` — the stale `**characters** (`wc -c`)` parenthetical is corrected and a note records the byte/character/locale trap. `.claude/rules/regression-testing.md` gains the Known Regressions row required by that rule's checklist.

Closes #2135


laurigates:

Issue #1482, opened 2026-06-02
feedback(health-plugin): health-check Context block leaks 'bfs: error' when project has no .claude/ dir


laurigates:
## Skill

`health-plugin/skills/health-check/SKILL.md`

## Category

Bug

## Description

The `/health:check` skill's **Context** block reports project/local settings existence by running a command that errors on a missing `.claude/` directory, leaking the raw error string into the prompt context instead of gracefully reporting "absent".

In a workspace whose root has no `.claude/` directory, the Context block rendered:

```
- Project settings exists: bfs: error: .claude: No such file or directory.
- Local settings exists: bfs: error: .claude: No such file or directory.
```

The detection command (`bfs` — a breadth-first `find` replacement) is not robust to the target directory being absent; it errors rather than emitting `false`/`MISSING`.

## Evidence

Both lines above appeared verbatim in the skill's Context section when invoked from `/mnt/sabrent/comfyui-workspace` (no root `.claude/`). The error is cosmetic — the rest of the skill ran correctly — but it appears twice and reads as a failure.

## Suggested Action

Make the existence check tolerate a missing `.claude/` dir, e.g.:

- Guard with a directory test first: `test -d .claude && bfs .claude -maxdepth 1 -name 'settings.json' ... || echo MISSING`, or
- Suppress stderr and treat non-zero exit as "absent": emit `MISSING` / `false` rather than the tool's error text.

Per the skill's own Context-block notes, stderr from a Context backtick can abort the skill — so the same robustness rule that protects `git remote -v` should cover the settings-existence probe.

laurigates:
Already resolved: PR #1496 switched the health-check/health-audit Context probes to the correct `find . -maxdepth 2 -path '*/.claude/settings.json'` form, and the `find-dotclaude-start-path` lint regression rule (referencing #1482) is present. Closing as completed.

Issue #1205, opened 2026-04-29
feedback(taskwarrior-plugin): task-status Context line triggers permission prompt for `command -v task`


laurigates:
## Skill

`taskwarrior-plugin/skills/task-status/SKILL.md`

## Category

Bug

## Description

When `/taskwarrior:task-status` is invoked, the skill's Context section
(line 32) runs `!`command -v task`` as part of slash-command frontmatter
expansion. The harness rejects this with a permission-approval prompt
and the skill aborts before its body can execute.

## Evidence

Session output:

```
<command-name>/task-status</command-name>
<local-command-stderr>Error: Shell command permission check failed for pattern "!`command -v task`": This command requires approval</local-command-stderr>
```

Source (taskwarrior-plugin 1.2.1, `skills/task-status/SKILL.md` L32):

```markdown
- Task CLI available: !`command -v task`
```

The skill's `allowed-tools` declares `Bash(task *)` — but `command` is a
shell builtin, not `task`, so the wildcard does not match.

## Suggested Action

Either:

1. **Add `Bash(command *)` to the `allowed-tools` frontmatter** so the
   Context probe is pre-approved (cheapest fix).
2. **Replace the probe with a `task --version` invocation** that *is*
   covered by the existing `Bash(task *)` permission. `task --version`
   exits 0 if installed, errors out cleanly if not — same signal,
   already permitted.
3. **Move the availability check into the body** as a guarded `task ...`
   call, so missing-task is reported by the skill rather than the
   permission layer.

Option 2 is the least intrusive; option 1 is also fine if other skills
have a similar shape and a shared rule emerges.

PR #2155, opened 2026-07-27
fix(project-plugin): stop project-init aborting when gh is absent

### LINKED ISSUE #2145: fix(project-plugin): project-init aborts for any user without gh on PATH (`which gh` exits 1)
Found while verifying #2139. Pre-existing on `main`, unrelated to that PR's subject — filing separately.

## The bug

`project-plugin/skills/project-init/SKILL.md:27`:

```markdown
- GitHub CLI: !`which gh`
```

`which` exits **1** when the tool is absent. A non-zero exit from a `## Context` backtick aborts the skill before its body runs — so `/project:init` is unusable for anyone who doesn't have `gh` installed, which is exactly the fresh-machine case the skill is for.

Reproduced by the semantic backstop:

```
KIND=FAIL FILE=project-plugin/skills/project-init/SKILL.md:27 EXIT=1
CMD=which gh
STDERR=
```

This is the same abort class as #899 / #1205 / #1206 / #1351 — a Context command that exits non-zero when the thing it probes doesn't exist.

## Why the existing guards missed it

Two independent gaps, both worth noting because they'll let the next variant through too:

1. **The regex denylist has no `which` rule.** `lint-context-commands.sh` covers `test -f`, `command -v` (#1205), `git config --get` (#1206), `git rev-parse`/`git remote` (#1351) — but not `which`. It is the only `which` in a Context block repo-wide, so it never came up.
2. **The executing backstop classifies it as `FAIL`, not `ENV_MISSING`.** `check-context-command-execution.sh` decides `ENV_MISSING` from **stderr text** ("command not found", auth errors). `which` writes **nothing** to stderr — it signals purely through the exit code — so a genuinely environment-dependent probe is indistinguishable from a broken command.

Gap 2 is arguably the more interesting one: it means any silent exit-1 probe reads as a hard failure. Worth deciding whether the classifier should also consider "exit 1, empty stderr, command is a known PATH-lookup builtin" as `ENV_MISSING`.

## Constraints on the fix

The obvious alternatives are already blocked:

- `command -v gh` — blocked in Context by `lint-context-commands.sh` (#1205: `command` is a shell builtin not covered by `Bash(<tool> *)` patterns, so it triggers a permission prompt).
- `which gh || echo missing` — `||` is a blocked shell operator in Context.
- `type gh` — same exit-1 behaviour as `which`.

The house pattern for existence checks that must exit 0 is `find` (see `lint-context-commands.sh` rule `grep-hardcoded-single-file` and the #1482 fix). Either that over a fixed set of bin directories, or drop the Context line and detect `gh` in the body where a non-zero exit is survivable.

## Acceptance

- [ ] `scripts/check-context-command-execution.sh --strict` exits 0 on a clean tree
- [ ] A regression rule lands per `.claude/rules/regression-testing.md` — a `which-in-context` rule in `lint-context-commands.sh`, and/or the `ENV_MISSING` classifier gap above
- [ ] `/project:init` still reports whether `gh` is available (don't fix the abort by silently dropping the capability check)

laurigates:
Closes #2145.

## The bug, reproduced first

`project-plugin/skills/project-init/SKILL.md:27` carried ``- GitHub CLI: !`which gh` ``.
`which` signals "not found" purely through its **exit code** — it exits 1 and writes
**nothing** to stderr — and a non-zero exit from a `## Context` backtick aborts the
skill before its body runs. `/project:init` was unusable for anyone without `gh` on
PATH: exactly the fresh-machine case the skill exists for.

I did not take the diagnosis on trust. On a machine that *has* `gh`, the backstop is
green — which is itself the finding:

```
$ bash scripts/check-context-command-execution.sh --strict --files "project-plugin/skills/project-init/SKILL.md"
FILES_SCANNED=1
PASS=3
FAIL=0
STATUS=OK
EXIT=0
```

Re-running with `gh` removed from PATH (a shim dir carrying `uv`/`git`/`python3`/`which`
but not `gh`) reproduces the issue's report byte-for-byte:

```
$ PATH="$NOGH" bash scripts/check-context-command-execution.sh --strict --verbose --files "project-plugin/skills/project-init/SKILL.md"
PASS=2
FAIL=1
ISSUES:
  - KIND=FAIL FILE=project-plugin/skills/project-init/SKILL.md:27 EXIT=1
    CMD=which gh
    STDERR=
STATUS=ERROR
EXIT=1
```

**That is a third gap, and it is why this shipped.** `check-context-command-execution.sh`
only fires when *the harness box itself* lacks the tool. Every developer machine and
every CI runner has `gh`, so the abort was permanently latent behind a green gate. The
regex denylist was the only thing that could have caught it, and it had no `which` rule.

## The fix

Moved the probe out of `## Context` into the skill **body**, run via the Bash tool where
a non-zero exit is survivable, with an explicit availability report and an `--github`
fallback. The capability check is kept, not dropped — the skill now reports
`GitHub CLI: available (<path>)` or `GitHub CLI: not installed — install from
https://cli.github.com/`, and the GitHub Repository Creation section is gated on it
(scaffold locally, report the skip, never abort).

### Why not the `find`-over-bin-dirs house pattern

The issue named `find` (the #1482 fix shape) as the alternative. I rejected it, twice over:

1. **It only moves the abort.** `find` writes to stderr when a *start directory* does not
   exist — `find /usr/local/bin /opt/homebrew/bin ...` errors on Linux, where
   `/opt/homebrew` is absent. That is literally the #1482 class the pattern was invented
   to fix, reintroduced. The #1482 remedy (start from a path that always exists, match with
   `-path`) has no analogue here: there is no universally-present bin-directory set.
2. **It would report the wrong answer.** `gh` is routinely installed via mise, asdf, nix,
   cargo, or `~/.local/bin` — none of which a fixed bin-dir list can see. On this machine
   `gh` resolves to `~/.local/share/mise/installs/aqua-cli-cli/.../bin/gh`, which no
   plausible fixed list contains. A probe that reports "not installed" for a user who has
   `gh` is worse than no probe.

### Why not `gh --version` (the repo's dominant existing pattern)

30+ skills use `<tool> --version` (the #1205 fix shape: `task --version`, `jq --version`,
`k6 version`). It does **not** meet the requirement that the line exit 0 regardless: with
`gh` absent it exits 127 *and* writes `command not found` to stderr, so it still aborts a
real skill invocation. It merely relocates the failure into the harness's `ENV_MISSING`
bucket, which the gate ignores. That would have laundered the bug, not fixed it. See the
follow-up note below — this is a larger latent problem than #2145.

## Regression coverage

Per `.claude/rules/regression-testing.md`:

- **New denylist rule `which-in-context`** in `scripts/lint-context-commands.sh`, anchored
  to Context command lines so `which` in fenced code, prose, or a table cell is not
  flagged. The fix message names `type` and `hash` as sharing the exit-1 behaviour, states
  the correct fix shape (probe in the body), and records why `find` is not a substitute.
- **`scripts/tests/test-lint-context-commands.sh` — the linter's first regression test.**
  It had none, which is the deeper reason its rule set could drift. Five guards:

  | Guard | Asserts |
  |---|---|
  | A | Context `which gh` → `ERROR [which-in-context]`, exit 1 |
  | B | `which` in fenced code / prose / a table is **not** flagged (no false positive) |
  | C | a clean Context block exits 0 with no findings (guard integrity) |
  | D | the pre-existing `command-v-in-context` rule (#1205) still fires |
  | E | no `which` remains in any Context block **repo-wide** — pins this fix against revert |

- Wired into `.pre-commit-config.yaml` and the `Plugin: Lint skills` workflow (local↔CI parity).
- `Known Regressions` table row added.

**The test is a real gate, verified in both directions** against the pre-fix file:

```
$ git show origin/main:project-plugin/skills/project-init/SKILL.md > .../skills/project-init/SKILL.md

# pre-fix linter — the gap that let #2145 ship
$ bash <origin/main linter>
All context commands OK
EXIT=0

# this PR's linter
$ bash scripts/lint-context-commands.sh
ERROR [which-in-context]: ./skills/project-init/SKILL.md:27
  Found: - GitHub CLI: !`which gh`
  Fix: move the probe into the skill body and run it via the Bash tool ...
EXIT=1
```

## Decision on the `ENV_MISSING` classifier gap (issue item 2) — deliberately NOT closed

The issue asks whether `check-context-command-execution.sh` should also treat
"exit 1, empty stderr, known PATH-lookup builtin" as `ENV_MISSING`. **I recommend against
it, and did not implement it.** Three reasons:

1. **It points the wrong way.** A silent exit-1 probe *is* a hard failure — a Context
   command exiting non-zero aborts the skill whether the cause is "environment" or "bug".
   `ENV_MISSING` exists to excuse *harness-box* gaps (a CI runner lacking `jq`) where the
   skill is fine for real users. "The user doesn't have `gh`" is not a harness artifact; it
   is the exact real-user scenario `/project:init` targets. Reclassifying it would have made
   the gate **permanently blind** to #2145 rather than merely latently blind.
2. **It has no remaining subject.** With `which-in-context` in the denylist, `which`/`type`/
   `hash` cannot reach a Context block, so the proposed heuristic would classify nothing.
3. **Minimality.** Loosening the gate's own strictness in the same PR that fixes an abort
   the gate failed to catch is a net-negative risk.

### The more useful finding, for a separate issue

The classifier's real weakness runs the **other** way. `<tool> --version` probes are
classified `ENV_MISSING` and pass the gate, yet on a machine lacking that tool they abort a
real skill exactly like `which gh` did — stderr `command not found`, exit 127. This repo has
~30 of them (`task --version` x8 in taskwarrior-plugin, `jq --version` x2 in hooks-plugin,
`dotenvx --version`, `k6 version`, `cargo machete --version`, `git version`, ...). A full
repo run currently reports `ENV_MISSING=16` on this box; that is a floor, not a ceiling,
since it only counts tools *this* machine happens to lack.

Tightening `ENV_MISSING` would red-light those ~30 skills across four plugins at once. It
is a real correctness question but far outside a minimal fix for #2145, so I am flagging it
here to be filed separately rather than smuggling it in.

**Unrelated pre-existing bug noticed in passing** (also present on `origin/main`, not
introduced here): `check-context-command-execution.sh --verbose` with zero findings emits
`line 251: findings_file: unbound variable` under `set -u` (empty `declare -a` array).
Cosmetic — it does not affect the exit code or the gate.

## Acceptance

- [x] **`scripts/check-context-command-execution.sh --strict` exits 0 on a clean tree**

  ```
  $ bash scripts/check-context-command-execution.sh --strict
  === CONTEXT COMMAND EXECUTION ===
  FILES_SCANNED=413
  PASS=639
  ENV_MISSING=16
  UNPARSEABLE=1
  SKIPPED_DENYLIST=0
  FAIL=0
  STATUS=OK
  ISSUE_COUNT=0
  === END CONTEXT COMMAND EXECUTION ===
  EXIT=0
  ```

  And with `gh` absent from PATH — the case that used to fail:

  ```
  $ PATH="$NOGH" bash scripts/check-context-command-execution.sh --strict --files "project-plugin/skills/project-init/SKILL.md"
  PASS=2
  FAIL=0
  STATUS=OK
  EXIT=0
  ```

- [x] **A regression rule lands per `.claude/rules/regression-testing.md`** — the
  `which-in-context` rule plus `scripts/tests/test-lint-context-commands.sh` (5/5 pass),
  wired into pre-commit and CI, with a `Known Regressions` row. The `ENV_MISSING`
  classifier change is deliberately declined, with reasoning above.

  ```
  $ bash scripts/tests/test-lint-context-commands.sh
  ----
  PASS=5 FAIL=0

  $ just lint-context-commands
  All context commands OK
  ```

- [x] **`/project:init` still reports whether `gh` is available** — moved to a
  "GitHub CLI Availability" body step with an explicit report table for both outcomes and
  a documented `--github` fallback. Not silently dropped.

Other gates run locally: `shellcheck` (clean), `scripts/lint-shell-scripts.sh`
(0 errors), `actionlint` on the changed workflow (clean), and the full `pre-commit`
suite over every changed path (no failures).


laurigates:

laurigates:

laurigates:

Issue #2036, opened 2026-07-10
hook: demote ls→Glob exit-2 block to opt-in teach nudge in bash-antipatterns


laurigates:
## Evidence

Weekly friction-learner run (findings file: `~/.claude/friction-reports/2026-W28-frictions-2026-07-10.md`, evidence window Jul 3–10, 345 transcripts / 186 sessions / 551 events).

`ls`→Glob is the **top hook friction** this run and the highest-repeat hook signal:

| Run | events | sessions | same-session repeat-block rate |
|---|---|---|---|
| W28 re-run (2026-07-10) | 63 | 44 | **31.8%** |
| W28 (2026-07-06) | 78 | 54 | **33.3%** |

Sustained **≥30% over two consecutive readings** — the same "resistant to teaching" profile that justified demoting `find`→Glob (PR #1871, merged) and `grep`/`rg`→Grep (#1909, completed).

**It is an exit-2 hard block, not a nudge.** Live hook `hooks-plugin/hooks/bash-antipatterns.sh:250-253`:

```sh
# Check for ls used for file listing (should often use Glob)
if [ "$IS_REMOTE_EXEC" = false ] && echo "$COMMAND" | grep -Eq '^\s*ls\s+.*\*'; then
    block "REMINDER: Consider using the Glob tool for pattern-based file listing. ..."
fi
```

`block()` exits 2. Reproductions:
- `ls -1 <dir>/*.json` → **exit 2** (this run's friction-learner repro).
- **Compound false positive**: `ls -1 ~/.claude/projects/ … | head; … find … -name "*.jsonl" …` → **exit 2** — the regex `^\s*ls\s+.*\*` matches because the command *starts* with `ls` and a `*` appears *anywhere later* (here in an unrelated `find -name "*.jsonl"` clause). The blocked command was not even an `ls` glob.

**Documented behavior mismatch:** `.claude/rules/bash-tool-replacements.md` describes the `ls` row as *"Reminder only on `ls *.glob` patterns"* and states that only `cat`/`head`/`tail` remain hard blocks after the find/grep demotions. The live hook contradicts the rule — it exit-2 blocks.

## Proposal

Mirror the `find` (#1871) and `grep`/`rg` (#1909) demotions:

1. `hooks-plugin/hooks/bash-antipatterns.sh` — remove the `block()` call at lines 250-253 (the `^\s*ls\s+.*\*` branch).
[...]

Issue #1873, opened 2026-06-29
friction: long-pipeline REMINDER persists at 37.5% repeat — propose substitution table (W27)


laurigates:
## Evidence

**Source**: [`~/.claude/friction-reports/2026-W27-frictions.md`](https://github.com/laurigates/claude-plugins/) — week ending 2026-06-29.

W27 reading: **42 events / 24 sessions / 37.5% same-session repeat-block** — the **highest of any tracked friction signal** this week.

The `long_pipeline` pattern (≥5 piped stages starting from `cat`/`echo`/`printf`) was previously bundled with `echo>file` under a combined headline in W26 (~107 events). W27 splits them cleanly:

- `echo>file` resolved itself (4 ev / 0% repeat) — no further action.
- `long_pipeline` persists as its own sticky pattern (42 ev / 37.5% repeat).

The W26 "combined 107" framing masked which half persisted.

### Relationship to closed work

[Issue #1411](https://github.com/laurigates/claude-plugins/issues/1411) (closed 2026-06-01) extended `bash-antipatterns` substitution-format BLOCKED messages to git-chain, ls-glob, **long-pipelines**, and TaskOutput. The git-chain half drove its repeat rate from 9.4% → 4.8%. The long-pipelines half is the surviving 37.5%-repeat signal — the BLOCKED message exists and suggests alternatives, but the substitution is not being recalled.

This is the **same class of friction as #1470**: guidance is present, substitution recall is the gap.

## Proposal

File this issue as a research tracker mirroring #1470's shape. Before any code edit:

1. **Sample 10-15 raw evidence entries** from `~/.claude/friction-reports/2026-W27-frictions.md` and the parsed JSONL to identify the 3-5 dominant pipeline shapes. Hypothesised candidates:
   - `cat <file> | grep <pat> | awk … | sort | uniq -c | sort`
   - `gh … --json … | jq … | grep … | head`
   - `<cmd> | tr … | sort | uniq | wc -l`
   - `echo … | sed … | base64 …`
2. **Taste call** on the substitution shape:
   - (a) Inline-table common pipeline shapes into the REMINDER text — mirrors what worked for `find` (6.1%) and `cat`/`head`/`tail` (6.3%) in W27.
   - (b) Extract a new `long-pipeline-substitutions.md` rule under `.claude/rules/` — heavier-weight but reusable.
   - (c) Write only a user-global rule — minimal blast radius, no plugin code change.
[...]

PR #2191, opened 2026-07-28
fix(hooks-plugin): narrow the cat/head/tail read block to whole-command reads

### LINKED ISSUE #2148: hook: cat/head/tail read block scope widened by #2114 — same-session repeat 15% → 46%
Weekly friction-learner reading for **2026-W31** promoted `cat`/`head`/`tail` → `Read` to the run's dominant signature, with a discrete, datable cause: PR #2114 (ast-grep structural classifier) widened the block's catch surface from "a bare read as the whole command" to "a bare read *anywhere* in a compound command".

Findings file: `~/.claude/friction-reports/2026-W31-frictions.md` § "Signal A".
(Prior baseline: `2026-W29-frictions.md`. Note W30 was skipped, so the trend spans two weeks.)

## Evidence

Seven readings of the same signature — six inside a 6.3–20.0% band, then a **3.0× break**:

| Reading | Date | Events | Sessions | Same-session repeat |
|---|---|---|---|---|
| W26 | Jun-22 | 36 | 32 | 12.5% |
| W27 | Jun-29 | 34 | 32 | 6.3% (best on record) |
| W27 re-run | Jul-04 | 64 | 56 | 14.3% |
| W28 | Jul-06 | 56 | 49 | 14.3% |
| W28 re-run | Jul-10 | 42 | 35 | 20.0% |
| W29 | Jul-13 | 53 | 46 | 15.2% |
| **W31** | **Jul-27** | **70** | **41** | **46.3%** |

- Per-session prevalence: **21.6% → 68.3%** of sessions.
- **30%** of all friction events this week.
- New-day-only slice (Jul 20–27): 68 ev / 40 sess / **45.0%** — not a carry-over artifact.
- W29 characterised this signature as *"earned safety block; low repeat = teaching works."* Both halves of that claim now fail.
- The repeat rate now **exceeds every signature previously demoted** on the `hook-block-vs-nudge.md` standard: `find` 12%, `grep`/`rg` 21%, `long_pipeline` 28.1%, `ls` 31.8% — vs `cat-head-tail` at **46.3%**.

### Cause established by differential reproduction (not correlation)

The *actual command shapes recovered from this week's blocked events*, run against three installed hook versions (exit codes):

| Command shape (recovered from transcripts) | 2.5.14 | 2.6.0 | **2.7.0** |
|---|---|---|---|
| `cat tmp/krea2-skslora.yaml` (bare read) | 2 | 2 | **2** |
| `git status --porcelain \| head -30 && echo … && head -25 docs/index.md` | 0 | 0 | **2** |
| `ls tests/js/; head -20 tests/js/index.test.js` | 0 | 0 | **2** |
| `cat knip.json; echo "=== resolve ==="; ls node_modules/@x/ 2>&1 \| head` | 0 | 0 | **2** |
| `echo "=== requirements.txt ===" && cat -n requirements.txt` | 0 | 0 | **2** |
| `cd /repo && git log --oneline -3 && cat package.json && ls -R src` | 0 | 0 | **2** |

Only the bare-read case blocked before #2114. Every compound form is **newly** blocked. Pipelines remain correctly exempt in 2.7.0 (`find … | head -40` → 0, `git status | head -30` → 0, `cargo test | grep | tail -2` → 0).

**This is not a false-positive bug — the classifier is behaving exactly as written.** It is a scope expansion that was not the stated intent of #2114 (whose goal was *fewer* false positives: #1701, #1721, #1722, #1848, #1900, #2052, #2058).

Corroborating distribution: of 72 recovered blocked commands, the leading token was `cd` (22), `ls` (9), `echo` (6), `grep` (3), `git` (3), other (4) — i.e. **~47 of 72 (65%) do not begin with `cat`/`head`/`tail`** and would have passed pre-#2114. Only ~19 are the bare reads the block originally targeted.

### Why the *repeat* rate specifically explodes

For a bare `cat file.md` the substitution is clean and one-to-one: `Read(file_path=…)`. For a compound diagnostic — `cd repo && git log --oneline -3 && cat package.json && ls -R src` — there is **no single-call substitution**. The agent must decompose one Bash call into a Bash call *plus* N `Read` calls, re-plan the sequence, and re-issue. That is a workflow change, not a tool swap, so the agent frequently re-attempts a variant and is blocked again inside the same session — precisely what the same-session repeat metric measures.

### The litigation is already written — in the hook's own source

`hooks-plugin/hooks/bash-antipatterns.sh` lines 5–19 define a deliberate two-tier split:

> 1. Scoping-sensitive read/write anti-patterns (cat/head/tail reads, echo/printf/cat writes, sed -i, task-output reads) are classified STRUCTURALLY with ast-grep … These are **STYLE nudges**: when ast-grep is absent (sandboxes, subagents) they simply do not fire.
> 2. Safety / correctness blocks (curl|bash, chmod 777, git add -A, reset --hard, …) stay pure-regex and fire in EVERY context …

Line 409 restates it: *"They are STYLE nudges (no-op when ast-grep is absent), **not safety blocks**."* Yet all three read detectors call `block()` → exit 2 (lines 258, 267, 276). **The implementation contradicts the stated design.**

Against `.claude/rules/hook-block-vs-nudge.md`'s four-question test:

1. *Irreversible harm?* No — reading a file is non-destructive. → nudge.
2. *Does it exempt the dangerous variant?* It exempts pipelines, heredocs, `/dev/*`, and every ast-grep-absent context. It blocks only benign plain reads. → style; nudge.
3. *Redirect target absent in some subagents?* `Read` is widely available, and the detector fails open without `ast-grep` — so the classic subagent dead-end is **partially** mitigated. This is the one axis where cat/head/tail is genuinely better-behaved than the demoted `find`/`grep` blocks.
4. *Regression test?* Required either way.

## Proposal

Two candidate remedies; the design choice is deliberately left to human review.

**(a) Narrow — restore pre-#2114 scope (better-evidenced default).**
Add a positional constraint to the `cat-read` / `head-tail-read` ast-grep rules so they fire only when the read is the sole/leading command, not when embedded in a compound statement — e.g. `not: { inside: { kind: list } }`, mirroring the exemption pipelines already get. Keeps the earned bare-read steer (which sat happily at 6–20% for six readings) and reverts only the unintended expansion. Smallest blast radius; preserves #2114's genuine false-positive fixes.

**(b) Full demotion.**
Move all three read detectors to `bash-antipatterns-teach.sh`, mirroring `find` (#1871), `grep`/`rg` (#1909), `ls` (#2036/#2037), and `long_pipeline` (#1873). Maximally consistent with the hook's own "STYLE nudge" self-description and with four landed precedents.

(a) is preferred: the block was demonstrably *not* problematic at its pre-#2114 scope, so the minimal correct action is to revert the scope change rather than retire a steer that was working.

## Why this is RESEARCH not PR-READY

Two defensible remedies with materially different outcomes — a design call, not a mechanical substitution — and the standing precedent (set by the `find`/`grep`/`ls`/`long_pipeline` demotions) that a behavioural change to a hook block warrants the `hook-block-vs-nudge.md` litigation write-up plus a regression check before it lands.

## Acceptance criteria

- W32/W33 `cat-head-tail` same-session repeat drops **below 20%** (back inside the six-reading historical band).
- Per-session prevalence returns to ~20–25% (from 68.3%).
- Compound forms exit **0**: `cd repo && git log --oneline -3 && cat package.json`, `ls dir/; head -20 file`, `echo "===" && cat -n file`.
- Bare `cat file.md` / `head -50 file.md` / `tail -50 file.md` still exit **2** (if remedy (a); exit 0 with a nudge if remedy (b) — state which in the PR).
- Pipelines and heredocs remain exempt (exit 0), as today.
- The false positives #2114 fixed do **not** regress: #1701, #1721, #1722, #1848, #1900, #2052, #2058.
- A `scripts/check-*.sh` regression check pins the chosen scope, per `.claude/rules/regression-testing.md`.

## Related

- PR #2114 — the ast-grep structural classifier that changed the scope (merged Jul-18 11:53Z, deployed as `hooks-plugin` 2.7.0 at 14:42Z).
- #2131 — an independent over-firing detector (`awk` read-only false positive) in the *same* hook version. Two independent over-firing detectors in 2.7.0 is itself worth noting.
- Demotion precedents: #1871 (`find`), #1909 (`grep`/`rg`), #2036/#2037 (`ls`), #1873 (`long_pipeline`).
- `.claude/rules/hook-block-vs-nudge.md`, `.claude/rules/bash-tool-replacements.md`, `.claude/rules/regression-testing.md`.

<sub>Filed by the weekly friction-learner scheduled run (2026-W31). Evidence: `~/.claude/friction-reports/2026-W31-frictions.md`.</sub>

laurigates:
Closes #2148

## The regression

PR #2114 replaced the regex classifier in `bash-antipatterns.sh` with an `ast-grep` structural one. Its stated goal was *fewer* false positives (#1701, #1721, #1722, #1848, #1900, #2052, #2058) and it delivered those — but it also **widened** the `cat`/`head`/`tail` read block's catch surface, from *"a bare read as the whole command"* to *"a bare read **anywhere** in a compound command"*. Nothing in the PR intended that; the AST rules simply carried no positional constraint (`not: { inside: { kind: pipeline } }` was the only scope gate, and a compound command's arms are `list` children, not pipeline stages).

Seven friction-learner readings of the same signature — six inside a 6.3–20.0% band, then a **3.0× break**:

| Reading | W26 | W27 | W27re | W28 | W28re | W29 | **W31** |
|---|---|---|---|---|---|---|---|
| Same-session repeat | 12.5% | 6.3% | 14.3% | 14.3% | 20.0% | 15.2% | **46.3%** |

W31: **70 events / 41 sessions**, ~30% of all friction events that week; per-session prevalence **21.6% → 68.3%**. Of 72 recovered blocked commands, ~47 (**65%**) did not begin with `cat`/`head`/`tail` and would have passed pre-#2114. 46.3% exceeds every signature previously demoted on the `hook-block-vs-nudge.md` standard (`find` 12%, `grep`/`rg` 21%, `long_pipeline` 28.1%, `ls` 31.8%).

**Why the repeat rate specifically explodes.** For a bare `cat f.md` the substitution is one-to-one: `Read(file_path=…)`. For `cd repo && git log -3 && cat p.json` there is **no single-call substitution** — the agent must decompose one Bash call into a Bash call *plus* N `Read` calls and re-plan the sequence, so it re-attempts a variant and is blocked again in the same session. That is exactly what the repeat metric measures.

This was hit live during preflight for the session that produced this PR: `git fetch; git status | head -3; wc -l file; head -20 file; ls dir` was blocked because one segment was a `head`.

## Option chosen: (a) narrow, not (b) demote

**Remedy (a).** `hook-block-vs-nudge.md`'s litigation test says an exit-2 is earned only where the substitution is clean. It is clean for a bare read; it is not for a compound command. So the correct scope is "the read IS the whole command" — which is precisely the pre-#2114 scope.

Full demotion to the teach nudge — the `find` (#1871), `grep`/`rg` (#1909), `ls` (#2036), `long-pipeline` (#1873) path — was **rejected**: the block sat happily at 6–20% for six readings at its narrow scope, so it is a steer that demonstrably *works*, and demoting it would discard a real context-budget win over a defect that #2114 introduced three weeks ago. The minimal correct action is to revert the widening.

## Implementation

The three **read** rules (`cat-read`, `head-tail-read`, `task-output-read`) gain a sole-statement gate:

```yaml
    - inside: { kind: program }
    - nthChild: { position: 1, ofRule: { not: { kind: comment } } }
    - nthChild: { position: 1, reverse: true, ofRule: { not: { kind: comment } } }
```

- `inside: { kind: program }` — the read's parent is the top level, so it is not an arm of a `&&`/`||` list, not a pipeline stage (that parent is `pipeline`), not inside a subshell / substitution / body.
- The forward **and** reverse `nthChild` pair makes it the *only* statement of that program. `ofRule: { not: { kind: comment } }` excludes comment siblings, so `cat f.md  # notes` still fires rather than being smuggled past by a trailing comment.
- The now-redundant `not: { inside: { kind: pipeline, stopBy: end } }` clause is dropped — `inside: { kind: program }` subsumes it. Pipelines and heredocs stay exempt, pinned by the suite.

`task-output-read` is narrowed alongside the other two even though #2148 names only `cat-read`/`head-tail-read`: it is literally the same detector family (a `cat`/`head`/`tail` read of a `.output` path), it has priority over the generic rules, and leaving it wide would keep re-blocking exactly the compound shape under litigation.

The three **write** detectors (`cat-write`, `echo-printf-write`, `sed-inplace`) keep their original scope — they guard file *mutation*, not context budget, and neither the issue nor the friction data implicates them. Pinned by two GUARD rows.

## Verification

`bash hooks-plugin/hooks/test-bash-antipatterns.sh`:

```
Results: 171 passed, 0 failed
```

Green on **both** local `ast-grep` 0.44.1 and the CI-pinned **0.43.0** (`npm i -g @ast-grep/cli@0.43.0`) — the `nthChild` + `ofRule` construct behaves identically on both. Sibling suite `test-bash-antipatterns-teach.sh` 29/0. `shellcheck` and `scripts/lint-shell-scripts.sh` clean (0 errors).

Every acceptance criterion from the issue, run against the patched hook:

```
-- compound forms exit 0
OK  want=0 got=0  cd repo && git log --oneline -3 && cat package.json
OK  want=0 got=0  ls dir/; head -20 file
OK  want=0 got=0  echo "===" && cat -n file
-- bare reads exit 2
OK  want=2 got=2  cat file.md
OK  want=2 got=2  head -50 file.md
OK  want=2 got=2  tail -50 file.md
-- pipelines and heredocs exempt (unchanged)
OK  want=0 got=0  cat config.json | jq .key
OK  want=0 got=0  gh pr create --body "$(cat <<'EOF' ... EOF)"
-- #2114 false-positive fixes do not regress
OK  want=0 got=0  echo "hi" 2>/dev/null                                    (#1722)
OK  want=0 got=0  echo "State:" $(gh pr view 42 --json state 2>/dev/null)  (#1721)
OK  want=0 got=0  echo "use foo > bar.txt to write"                        (#1701)
OK  want=0 got=0  gh issue comment 5 --body "the sed -i rule fired here"   (#2052)
OK  want=0 got=0  ssh host <<EOF~cat /remote/file.txt~EOF                  (#1900)
OK  want=0 got=0  python3 - <<'PY'~head = "x"~PY                           (#1848)
OK  want=0 got=0  i1=$(gh issue create ... | tail -1); printf ... | tee    (#2051/#2052)
-- the actual W31 blocked shapes
OK  want=0 got=0  cat knip.json; echo "=== resolve ==="; ls node_modules/@x/ 2>&1 | head
OK  want=0 got=0  echo "=== requirements.txt ===" && cat -n requirements.txt
OK  want=0 got=0  git status --porcelain | head -30 && echo x && head -25 docs/index.md
OK  want=0 got=0  git fetch; git status | head -3; wc -l f.md; head -20 f.md; ls docs
```

## Regression guard

New `test-bash-antipatterns.sh` section **"read block fires only when the read is the WHOLE command (#2148)"** replays the issue's differential-reproduction table — six compound shapes that must exit 0 (including the preflight command that motivated the fix, and `sleep 5 && cat …/run.output`), plus GUARD rows pinning that every bare whole-command read still exits 2 (`cat`, `head -50`, `tail -50`, a task-output read, and a bare read carrying a trailing `# comment`), and that the write detectors still fire inside a compound command. Plus a Known Regressions row in `.claude/rules/regression-testing.md`.

## Two pre-existing test rows were flipped, deliberately

Both had asserted the *widened* behaviour, so they must change with it. Called out rather than quietly edited:

1. **`sleep 5 && cat …/run.output`** — used by the two task-output message assertions. Moved to the bare `cat …/run.output` form; the message assertions are unaffected (task-output-read still has priority over `cat-read` for a bare `.output` read).
2. **`cat local.txt && ssh host 'do thing'` (was exit 2, now exit 0)** — this row guarded #1900's *first-token anchoring* of the `IS_REMOTE_EXEC` guard, and that guard **no longer exists** (#2114 deleted it; the structural parse makes remote reads string/heredoc/argument nodes, so they are never detected). What it now really asserts is "a local read in a compound command blocks", which is the exact behaviour under litigation. The surviving invariant — that the AST path classifies a local `cat` as a genuine command node — is pinned by a **new adjacent row** asserting bare `cat local.txt` still exits 2.

## Docs, same commit

`.claude/rules/bash-tool-replacements.md`, `hooks-plugin/README.md`, `hooks-plugin/hooks/README.md`, and the Known Regressions row all land here (docs-currency).

Two notes on the rule file:

- **Corrected stale content.** Its "Remote-exec commands are exempt" section documented the `IS_REMOTE_EXEC` first-token suppression as still governing the safety blocks. #2114 deleted that code entirely (`grep -c IS_REMOTE_EXEC` → 0) and safety blocks were never suppressed. Rewritten to describe what actually happens now.
- **Always-loaded budget.** `bash-tool-replacements.md` is unscoped, so it is paid every turn. The first draft of this doc pass pushed `check-context-engineering.py --strict` **over** its 100,000-char budget (98,484 → 101,342, ERROR). The narrative was compressed into the hook source comment and the Known Regressions row (neither always-loaded), and the stale remote-exec section trimmed. Net: **+856 chars**, `--strict` exits 0 at 99,362 / 100,000.

## Surprise found, not fixed here

`cd /repo && echo "data" > realfile.txt` does **not** fire `echo-printf-write`, while the `;` form does. In an `&&` list tree-sitter-bash makes the whole `list` the `body` of the `redirected_statement`, so `has: { field: body, kind: command }` fails. Pre-existing #2114-era under-detection, unrelated to #2148 and in the safe (under-blocking) direction for a style nudge. Noted inline in the test file next to the guard row, which uses the `;` form for that reason. Worth a follow-up issue if the write nudge is considered load-bearing.


laurigates:

PR #2205, opened 2026-07-29
feat(rust-plugin): add cargo-generate skill and foundryvtt template pilot

laurigates:
Evaluates [cargo-generate](https://github.com/cargo-generate/cargo-generate) for this marketplace: adopts it where it's the obviously-right tool, and pilots it against the larger question of whether a template of **real files** beats a generator of **Python strings**.

## Where the impact actually is

The repo scaffolds new repos from two hand-rolled Python template engines — **4,324 lines** total:

| Surface | Generator | Size |
|---|---|---|
| `comfyui-node-scaffold` | `scaffold.py`, 4 variants × a `--widgets` switch | 3,045 lines |
| `foundryvtt-module-scaffold` | `scaffold.py`, 3 variants | 1,279 lines |
| `rust-plugin` | none — `cargo new` only | — |

Inside those generators the emitted TypeScript/JSON/YAML is Python string literals: `tsc`, `biome`, `actionlint`, and Renovate cannot see any of it, and the only proof it's valid is running the generator.

### 1. `rust-plugin/skills/cargo-generate` (new)

The gap was conspicuous — rust-plugin has a skill per cargo tool (`cargo-nextest`, `cargo-llvm-cov`, `cargo-machete`, `cargo-worktree-builds`) but nothing for the ecosystem's standard scaffolding tool. Covers generating, authoring (`cargo-generate.toml` placeholders / conditionals / Rhai hooks), and the non-interactive flags that make it usable from a skill.

Three gotchas documented because they were hit while building the pilot, not read off a docs page:
- Liquid eats GitHub Actions `${{ }}` and Handlebars `{{ }}` — both fixes spelled out
- `variable::set` rejects integers, so `system::date().year` needs `"" +`
- the project name is kebab-cased **before** hooks run, so a hook can't reject a malformed one

### 2. `foundryvtt-plugin/templates/foundryvtt-module/` (new) — the pilot

A cargo-generate port of the smaller `scaffold.py`. Chosen as the pilot deliberately: 1,279 lines and no existing test suite to re-point, versus comfyui's 3,045 lines and three tests. **`scaffold.py` remains the default** and the port isn't promoted until it has generated a real module repo end to end.

Three variants via `[conditional]`, `module_id` derived by a Rhai pre-hook (it must byte-match `module.json` `id`, the install folder, and the zip name), Liquid in path names (`styles/{{ module_id }}.css`).

### 3. `test-template-parity.sh` (new)

Runs **both** generators over the same inputs and requires byte-identical trees. Carrying two generators is only safe if divergence is caught, so the comparison is the test — not a spot-check.

The matrix includes a run with **every value non-default**, and that half earned its keep immediately: the first parity run used defaults throughout, passed, and was still hiding two hardcoded fields — the README Foundry-version line, and `renovate.yml`'s reusable-workflow org, which `scaffold.py` pins to `laurigates` regardless of `--publisher`. Defaults-only inputs cannot see that class of bug.

Verified three ways: passes clean (9/9), **SKIPs** when `cargo-generate` is absent, and **fails 7/9** when a single line is injected into the template.

## Known divergence

| Input | `scaffold.py` | cargo-generate |
|---|---|---|
| A name that is not kebab-case (`foundryvtt-Bad_Id`) | exits 2 | kebab-cases it to `foundryvtt-bad-id` and proceeds |

Both reach a valid Foundry id, but cargo-generate renames the author's repo rather than making them choose again. Asserted in the parity test so it can't change silently.

## Explicitly out of scope

**`configure-plugin`.** ADR-0006 rejected Cookiecutter/Yeoman on "one-time generation; doesn't help existing projects." cargo-generate has the same shape and that reasoning still holds — the 37 configure-* components apply standards to *existing* repos with drift re-audit. The line stays at: cargo-generate for repo birth, `configure-*` for everything after.

**This repo's own plugin scaffolding.** Three of the 7 "Creating a New Plugin" steps edit root files *outside* the generated directory (`marketplace.json`, release-please config/manifest, `enabledPlugins`), which a Rhai post-hook can't reach without `--allow-commands`. `claude plugin init` already covers the rest.

**`comfyui-node-scaffold`.** Deliberately second — only if this pilot pays.

## Costs, stated rather than buried

- **`cargo-generate` is a new dependency** for a TypeScript repo (`cargo install cargo-generate --locked`; not in the base image). This is the main cost of the port.
- **Renovate coverage** of the template's `uses:` pins: they sit at `templates/…/.github/workflows/`, matched by neither this repo's skill-markdown `customManagers` nor Renovate's root-anchored github-actions `fileMatch`. Not a regression — they were equally unmanaged as `scaffold.py` string literals — but real files at a real path are newly *able* to be managed. Noted as a follow-up in `templates/README.md`.

## Verification

- `test-template-parity.sh` — 9/9 (6 parity comparisons × 3 variants × default/non-default, plus name handling)
- `run-skill-script-tests.sh` — new test discovered and passing; the one failure (`test-bash-antipatterns.sh`) is pre-existing and environmental (`ast-grep` absent in this sandbox, so the structural rules fail open)
- `plugin-compliance-check.sh` — no new errors; new skill 7,913 chars, description 141 chars
- `check-docs-index.sh`, `check-git-sandbox-guards.sh --strict`, `check-version-pin-coverage.sh --strict`, `audit-skill-descriptions.py --strict-length`, `lint-context-commands.sh`, `lint-shell-scripts.sh` — all clean
- `pre-commit` on all changed files — clean

https://claude.ai/code/session_01VY6iFbfGYUGiPGrezm6ouB

---
_Generated by [Claude Code](https://claude.ai/code/session_01VY6iFbfGYUGiPGrezm6ouB)_


laurigates:
Triaged the compliance WARNs. Five of the seven are pre-existing on files this PR never touches (`bevy-ecs-patterns`, `mockito-http-mocking` ×2, `cargo-worktree-builds`, `cargo-llvm-cov`) — out of scope here.

Two are crossings this PR caused:

| Skill | Before | After bot run | Now |
|---|---|---|---|
| `foundryvtt-module-scaffold` | 8,831 | 10,640 | **9,930** ✅ |
| `bevy-game-engine` | 9,929 | 10,189 | 10,189 (unchanged) |

**Fixed** in 12292cd: the pilot section's detail already lived in `templates/README.md`, so per `context-engineering.md` the SKILL body keeps the pointer and the sidecar keeps the explanation. Also dropped a duplicate `cargo generate` invocation from the Agentic Optimizations table — the same command sat ten lines above it. No content lost, only relocated or de-duplicated.

**Not fixed — `bevy-game-engine`, deliberately.** It sat at 9,929 chars before this PR: **71 chars of headroom**. The change corrects genuinely outdated setup guidance (the skill said `cargo new my_game`; Bevy's own quickstart is `cargo generate --git .../bevy_new_2d`), and the template URL alone is 47 chars. I tried three progressively terser forms; the tightest that still carries a runnable command lands at 10,062, and even a prose-only pointer with no command lands at 10,003. Getting under would mean either dropping the guidance or making an unrelated cut to a skill this PR otherwise shouldn't touch.

It's a WARN against a 26,000-char ceiling, and the file was one edit away from crossing regardless of what that edit was. Happy to extract `bevy-game-engine` to a `REFERENCE.md` if you'd rather clear it, but that's its own change, not this one's.

---

PR #2189, opened 2026-07-28
refactor(agent-patterns-plugin): extract parallel-agent-dispatch material to REFERENCE.md

laurigates:
## Premise correction (read this first)

This was pulled forward from the #2174 epic analysis, which reported
`parallel-agent-dispatch/SKILL.md` as a **live build break** at **26,024
characters** against the 26,000 ERROR ceiling in
`scripts/plugin-compliance-check.sh:1497`.

**That figure is the byte count, not the character count.** Issue #2135 had
already fixed the gate to count decoded UTF-8 characters (`skill-quality.md`
documents this exact file as the motivating example: *"parallel-agent-dispatch
measured 26,024 bytes vs 25,850 characters"*). Measured on the pre-change file:

```
chars 25850
bytes 26024
```

So the gate reported **WARN, not ERROR**, and the repo was **not** failing its
own compliance check. Verified live before touching anything:

```
| agent-patterns-plugin | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ⚠️ | ✅ | ⚠️ | ⚠️ |
- ⚠️ .../parallel-agent-dispatch: SKILL.md is 25850 chars (~6462 tokens, >10000)
```

The refuted premise still marked a real gap, so the work proceeded: **150 chars
(0.6%) of headroom** on a file carrying **25 guard-pinned literal tokens** means
any routine edit adding one hazard subsection breaches the gate — and the pin
density means the breach can't be relieved by trimming freely, since most prose
is load-bearing for a `grep -qF`.

## Before / after

| | chars | headroom vs 26,000 |
|---|---|---|
| SKILL.md before | 25,850 | 150 (0.6%) |
| SKILL.md after | **23,404** | **2,596 (10.0%)** |
| REFERENCE.md before → after | 30,409 → 31,887 | n/a (no ceiling) |

## What moved

Extraction only — **no guidance was changed**. Whole coherent blocks moved to
`REFERENCE.md`, each leaving a one-line pointer behind:

| Moved | Now in REFERENCE.md |
|---|---|
| Nested-repo worktree narrative (#1838) | existing § "Nested-repo worktree isolation (#1838)" |
| `GIT_DIR`/`GIT_WORK_TREE` export narrative (#1692) | existing § "Worktree GIT_DIR-export leak (#1692)" |
| Skill-less agentType evidence (~88k/~12k injection sizes, #1549/#1550 batch-failure data) | **new** § "Skill-less agentType — evidence" |

[...]

Issue #2174, opened 2026-07-27
epic: migrate qualifying skills to Claude Code dynamic workflows


laurigates:
Tracking issue for the dynamic-workflow migration. Plan merged in #2162: [`docs/plans/dynamic-workflow-migration.md`](../blob/main/docs/plans/dynamic-workflow-migration.md).

## What was decided

47 candidate skills were evaluated by a 16-agent dynamic workflow (fan-out triage → per-group harness design → three adversarial lenses → synthesis). 18 were nominated; 21 refutations were raised; 6 were demoted by majority refutation.

**Of 47, five ship a harness.** The separating criterion:

> **The fan-out width must already be enumerable by something other than the model.**

Every mover reads its N from a deterministic source that already exists. Every rejection either invented its N, or duplicated a bound a shell script already enforces — `.claude/rules/offload-to-deterministic-substrate.md` run backwards.

Two of the blog post's six shapes — **tournament** and **loop-until-done** — have **no fit in this repo**. That is recorded as a finding, not a gap to close later.

## Sequencing

Groundwork gates everything. Nothing ships a `.js` before #2163.

```
#2163 (convention) ──┬──> #2164 (guard)
                     ├──> #2165 (fork conflict) ──> #2170 (test-analyze)
                     ├──> #2167 (verify-before-filing) <── #2166 (file-wave.sh)
                     ├──> #2168 (configure-all)
                     ├──> #2171 (blueprint-story-audit)
                     └──> #2172 (Tier B) ──> #2169 (evaluate-plugin-batch)

#2173 (Tier C) — independent, no .js, landable any time
```

### Groundwork

- [ ] #2163 — `.claude/rules/workflow-vs-skill.md` + CLAUDE.md + plugin-structure layout convention **(blocks all harnesses)**
- [ ] #2164 — `scripts/check-workflow-templates.sh` guard
- [ ] #2165 — resolve `context: fork` pin vs `skill-fork-context.md` **(blocker for #2170)**
- [ ] #2166 — extract `scripts/file-wave.sh` **(blocker for #2167)**

### Tier A — ship the harness

- [ ] #2167 — `workflow-verify-before-filing` · composite · *zero refutations*
- [ ] #2168 — `configure-all` · fan-out-and-synthesize · *zero refutations*
- [ ] #2169 — `evaluate-plugin-batch` · fan-out-and-synthesize · **must ship capped**
- [ ] #2170 — `test-analyze` · classify-and-act · *zero refutations*
[...]

Issue #2164, opened 2026-07-27
ci(scripts): add check-workflow-templates.sh guard for bundled workflow templates


laurigates:
Groundwork for the dynamic-workflow migration. Required by `.claude/rules/regression-testing.md` — several of the planned templates encode bug fixes, and **none is currently guarded**.

Source: [`docs/plans/dynamic-workflow-migration.md`](../blob/main/docs/plans/dynamic-workflow-migration.md) §Repo-level consequences, item 4 (merged in #2162).

## Scope

`scripts/check-workflow-templates.sh` + `scripts/tests/test-check-workflow-templates.sh`, wired into pre-commit and the plugin PR-checks workflow. Follows the `=== HEADER ===` / `KEY=VALUE` / `STATUS=` convention in `.claude/rules/structured-script-output.md`.

## Checks

1. **Every `workflows/*.js` is reachable.** Its sibling `SKILL.md` has a `## Workflow harness (template)` section referencing it, and that section contains the literal string `not a script to run verbatim`. An orphan `.js` fails.
2. **Worktree templates carry the resume warning.** Any template containing `isolation:'worktree'` also contains `#1868`. Mirrors the existing pin at `check-agent-failure-contract.sh:124-125`.
3. **`cold-read-gate/SKILL.md` retains `run_in_background: false` and `#2063`.** The #2063 fix currently has **no check of its own** — the batch template's verdict enum is the second fix to the same bug class, so the guard is overdue independently of whether that template ships.
4. **`execution-grounded-review/SKILL.md` retains the literal token `loop-integrity.md`.** Already required by `check-loop-integrity.sh:74` — worth noting in that skill's delta so a later "tighten the Related section" edit does not fail the build.

## Ordering note

Check 3 is worth landing **even if the `cold-read-gate` batch template never ships** (it is the weakest Tier B candidate). The #2063 guard gap is real today.

## Acceptance

- [ ] `scripts/check-workflow-templates.sh` implements checks 1–4, emits `STATUS=`
- [ ] `scripts/tests/test-check-workflow-templates.sh` covers pass and fail for each check
- [ ] Wired into `.pre-commit-config.yaml` and the PR-checks workflow, path-filtered to `**/workflows/*.js` and `**/SKILL.md`
- [ ] Passes against the repo as it stands (zero templates = vacuously green)

Blocked by the layout convention landing first (`## Workflow harness (template)` is the string it greps for).

PR #2207, opened 2026-07-29
refactor(agent-patterns-plugin): split parallel-agent-dispatch into references/

laurigates:
`Refs #2143` — **one slice**. This lands the repo's first `references/` directory and sets the pattern; the remaining ~52 flat-large bodies from the issue's ranked queue stay deferred.

## Why this skill first

`.claude/rules/context-engineering.md` (C3, judged **2.9**) asks for something the corpus has never done: *"divide it into many files and split them out"*. The measurement was **129** single-`REFERENCE.md` sidecars and **0** `references/` directories. A single sidecar defers content but is all-or-nothing — a run that needs one salvage routine loads every table in the file.

`parallel-agent-dispatch` is both the sharpest C3 case and the most dangerous one to touch, which is why it is the right place to establish the convention under real constraint rather than on an easy file.

## The split — by consumer path, not by size

`REFERENCE.md` was 31,887 chars. Splitting it down the middle would have defeated the point; each file is named for the *situation the reader is in*:

| File | Chars | Carries |
|---|---|---|
| `references/dispatch-contract.md` | 3,625 | Return Contract schema (verbatim), failure mode → schema field, Skill-less `agentType` evidence (the ~25k `skill_listing` tax) |
| `references/brief-templates.md` | 6,342 | Refactor-brief template + completion manifest (#1601), verbatim-patch discipline, agent self-verification, reviewer-agent evidence |
| `references/failure-recovery.md` | 13,209 | Commit/push stall salvage, WIP salvage (#1491), idle-without-report (#2039), `TaskStop` recovery + kill thresholds (#1424), rate-limit recovery-dispatch |
| `references/worktree-hazards.md` | 10,037 | cwd-reset (#1480), `GIT_DIR`-export (#1692), nested-repo isolation (#1838), target-branch preflight (#1969) |

Rationale per file: an orchestrator *composing a dispatch* needs the contract; one *writing a refactor brief* needs the templates; one *whose agent just failed* needs recovery; one *setting up isolation* needs the hazards. Those four are genuinely disjoint reading paths — the split is only worth anything if a scoped run loads one of them, not all four.

**Content moved verbatim.** The split was done by line-range extraction (script, not hand-editing) with no prose rewritten, so every pinned literal survives on its original single line. This is not fussiness: the guards grep `-qF` **per line**, and #2174's extraction of this same file nearly shipped a break when a re-wrap split the pinned literal `WIP at checkpoints` across a newline.

`REFERENCE.md` survives as a **thin index** (1,663 chars, one row per file). Keeping it means sibling links and the `docs`-level entry point keep resolving, and there is one obvious place to add the next reference file.

## Pinned-token audit

`check-agent-failure-contract.sh` pinned **32** literals across this skill pair, plus 2 more in `plugin-compliance-check.sh` (#1549). Enumerated before touching anything:

**SKILL.md — 23 pins, all untouched** (no content moved out of SKILL.md in this PR; only link targets changed):
`Loud-failure contract`, `Terminal.`, `#1422`, `empty vs dirty`, `WIP at checkpoints`, `#1491`, `#1480`, `git -C "$WORKTREE"`, `loses worktree isolation`, `#1546`, `re-runs succeeded worktree agents`, `#1868`, `#1692`, `env -u GIT_DIR`, `#1838`, `#1601`, `completion manifest`, `Target-branch preflight`, `git ls-remote --heads origin`, `#1969`, `Idle without report`, `idle_notification`, `#2039`.
Plus `skill_listing` and `Skill-less agentType` (`plugin-compliance-check.sh`, #1549).

**REFERENCE.md — 9 pins, all repointed in this commit:**

| Pin | Was | Now |
|---|---|---|
| `9:1`, `is_error`, `30%` (#1424) | `REFERENCE.md` | `references/failure-recovery.md` |
| `WIP salvage before re-dispatch`, `#1491` | `REFERENCE.md` | `references/failure-recovery.md` |
| `Worktree GIT_DIR-export leak` (#1692) | `REFERENCE.md` | `references/worktree-hazards.md` |
| `Nested-repo worktree isolation` (#1838) | `REFERENCE.md` | `references/worktree-hazards.md` |
| `#1601`, `Completion manifest` | `REFERENCE.md` | `references/brief-templates.md` |

Also verified: the other two files the guard reads (`custom-agent-definitions/SKILL.md` → `#1422`; `agents-plugin/agents/refactor.md` → `Completion Manifest`, `#1601`) are untouched.

## The new failure mode a thin index introduces — and its guard

A `references/` file that **nothing links to** is unreachable from the skill even though it still exists on disk, and no existing check would notice. So the guard gains an index-integrity loop: `REFERENCE.md` must still link all four files. Without it, the split would be one careless edit away from silently orphaning material the guards still (correctly) report as present.

Regression test extended (`scripts/tests/test-check-agent-failure-contract.sh`, fixture builder now copies `references/`):

- **K** — strip `9:1` from `failure-recovery.md` → exit 1. Proves the checker actually *reads* the split file; a split that quietly stopped asserting the #1424 thresholds would still exit 0 without this.
- **L** — strip `Nested-repo worktree isolation` from `worktree-hazards.md` → exit 1.
- **M** — strip a `references/` row from the `REFERENCE.md` index → exit 1.
- **F** repointed from `REFERENCE.md` to `references/brief-templates.md`.

`.pre-commit-config.yaml`'s `files:` regex extended to `references/.*\.md`, so editing a reference file actually triggers the guard (the pre-#2143 regex matched only `(SKILL|REFERENCE)\.md` — a reference file could otherwise have been edited with the guard never firing locally, the same wiring gap #2164 found on `check-loop-integrity.sh`).

## Recording the convention

`.claude/rules/skill-quality.md` gains a `references/` subsection under Supporting Files (path-scoped file — **zero** always-loaded cost). It states the prefer-single-vs-split decision, the four conventions the next 52 splits should follow (split by consumer path not size; keep a thin index; repoint guards in the same commit; cross-plugin stays a name reference), and links to `context-engineering.md` rather than restating its argument.

A `Known Regressions` row lands in `.claude/rules/regression-testing.md` (also path-scoped).

## Verification

```
$ bash scripts/check-agent-failure-contract.sh
OK: loud-failure contract, hook-thrashing heuristic, and WIP-salvage discrimination present in agent-patterns dispatch skills
exit=0

$ bash scripts/tests/test-check-agent-failure-contract.sh
check-agent-failure-contract (#1601/#1868/#2039/#2143): 13 passed, 0 failed
exit=0

$ bash scripts/plugin-compliance-check.sh agent-patterns-plugin
| agent-patterns-plugin | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ⚠️ | ⚠️ | ⚠️ | ⚠️ |
  ⚠️ parallel-agent-dispatch: SKILL.md is 24184 chars (~6046 tokens, >10000)
     — ceiling: 26000 chars / ~6500 tokens
exit=0   (no ❌; every ⚠️ pre-exists on main)
```

Plus: `shellcheck` clean on both edited scripts; `lint-shell-scripts.sh` → 0 errors; `check-loop-integrity.sh --strict` → `STATUS=OK`; `check-context-engineering.py --strict` → `C5_ALWAYS_LOADED_CHARS=99362` (**unchanged** — both edited rules are `paths:`-scoped), exit 0; a link/anchor walk over all 6 files → **29 relative links checked, 0 broken**.

### Before / after character counts

| File | Before | After |
|---|---:|---:|
| `SKILL.md` | 23,404 | **24,184** |
| `REFERENCE.md` | 31,887 | 1,663 |
| `references/*.md` (4 new) | — | 33,213 |
| `.claude/rules/skill-quality.md` | 13,067 | 15,083 |
| `scripts/check-agent-failure-contract.sh` | 12,530 | 14,179 |
| `scripts/tests/test-check-agent-failure-contract.sh` | 7,702 | 10,471 |

(Baselines read from `origin/main`, characters via decoded UTF-8 per `skill-quality.md` § Size Limits — not `wc -c` bytes, the #2135 trap.)

## One tradeoff worth flagging

**SKILL.md grew 780 chars** (23,404 → 24,184), cutting the headroom PR #2189 just created from 2,596 to **1,816**. Cause: `references/worktree-hazards.md#…` is inherently ~25 chars longer than `REFERENCE.md#…`, across 14 inline links. I clawed back ~430 chars by *not* re-enumerating the four files in SKILL.md's intro and Related section — the index already owns that list, and restating it is the C4 duplication this whole exercise is against. The remainder is the honest cost of addressable multi-file links. Still WARN, well under the 26,000 ERROR ceiling.

## Also fixed in passing

`workflow-orchestration-plugin:workflow-wave-dispatch` linked this sidecar by **relative path across plugins** (`../../agent-patterns-plugin/skills/…/REFERENCE.md`) — dead for anyone who installs one plugin without the other (`.claude/rules/skill-consolidation.md` § 2). Since I was changing the target anyway, it is now a name reference.


laurigates:

Issue #2009, opened 2026-07-06
feat(scripts): shared tree-sitter-markdown extraction helper for the five markdown-structure lint scripts


laurigates:
## Problem

Five lint/check scripts hand-roll markdown structure detection (fenced-code-block state machines, table-row skips, blockquote awareness), and two have already shipped bugs from it:

| Script | Hand-rolled structure | Shipped bug |
|---|---|---|
| `scripts/check-context-command-execution.sh` | fence toggle + table-row skip | #1744 — fenced ``!`cmd` `` executed as a Context command; table-row cell boundary misparsed as ``!` `` |
| `scripts/check-version-pin-coverage.sh` | fence toggle | #1492 — fence-state subshell rework |
| `scripts/lint-context-commands.sh` | regex over `` - …: !`cmd` `` list-item lines | — |
| `scripts/lint-mcp-tool-references.sh` | blockquote skip | — |
| `scripts/lint-package-references.sh` | blockquote skip | — |

The recurring query — *"extract inline-code spans inside list items, excluding fenced code blocks, tables, and blockquotes"* — is a one-liner against a real markdown parse and an unwinnable regex arms race otherwise.

## Proposal

Add a shared helper, e.g. `scripts/lib/extract-md-elements.py`, built on `tree-sitter` + `tree-sitter-markdown` (Python bindings), that emits structured `KEY=VALUE` / JSON rows per `.claude/rules/structured-script-output.md`:

- inline-code spans with their container context (list item / paragraph / table cell / blockquote / fence)
- fenced code blocks with language info
- file + line anchors

Then migrate the five scripts to consume the helper instead of their private state machines.

## Why a dependency is acceptable here

These scripts run in **pre-commit and CI only** — a controlled environment where a pinned Python dependency is fine. There is no fail-open problem (unlike the runtime hooks). This is `.claude/rules/offload-to-deterministic-substrate.md` applied one level down: the substrate is already deterministic, it's just using the wrong parser.

## Acceptance criteria

- [ ] Shared extraction helper with tests (fixtures: fence, table row, blockquote, nested list, `~~~` fences)
- [ ] `check-context-command-execution.sh` migrated; existing regression tests (incl. #1744 fixtures F/G) stay green
- [ ] `check-version-pin-coverage.sh` migrated; #1492 prune + fence tests stay green
- [ ] The three denylist lints migrated or explicitly deferred with rationale
- [ ] Runtime comparable to current (these are pre-commit-latency-sensitive)

[...]

Issue #1869, opened 2026-06-28
deadbranch skill: merge detection misses squash-merged branches


laurigates:
## What

The `git-plugin:deadbranch` skill (and the underlying `deadbranch` CLI) classify a branch as merged via **commit ancestry** (`git branch --merged`-style). On repos that **squash-merge** — the release-please / conventional-commit default — the squash creates a single new commit on the base branch with a fresh SHA, so the branch's own commits are never ancestors. The branch reads as **unmerged** even though its work fully landed.

## Why it matters

In a real cleanup session on a squash-merge repo, `deadbranch stats` reported only **4 of 28** actually-merged local branches as "safe to delete". The other 24 (all squash-merged, several with merged PRs) were classified unmerged, so the skill's safe path left them untouched. The user has to fall back to `--force` (which also sweeps genuinely-unmerged work) or audit by hand.

## Proposed enhancement

Add a squash-aware merge signal. Two deterministic checks that survive squash-merge **and** later file drift on the base:

1. **`merge-tree` no-op** — `git merge-tree --write-tree <base> <branch>` whose result tree equals `<base>^{tree}` proves the branch contributes nothing not already in base (git ≥ 2.38).
2. **MERGED PR for the branch head** — `gh pr list --state all --head <branch> --json state` (authoritative; the squash landed at merge time even if base drifted since).

A branch matching either is safe to delete; neither ⇒ keep/review.

Minimum viable: document the squash-merge limitation in the skill's SKILL.md so users know the "safe to delete" count is a lower bound on squash-merge repos. Better: add a `--squash-aware` (or default) pass using the two signals above.

## Evidence / reference

Worked through in a dotfiles session; the deterministic recipe is implemented as a `just -g branch-audit` recipe (laurigates/dotfiles#277) and documented as a rule (laurigates/dotfiles#278) — either could seed the skill's implementation.

PR #2249, opened 2026-07-31
fix(git-plugin): stop git-ops instructing ancestry-based branch deletion

laurigates:
## The defect

`git-plugin/agents/git-ops.md:161` shipped this in an **always-on agent prompt**:

```bash
git branch --merged main | grep -v 'main\|master' | xargs git branch -d
```

`git branch --merged` is an **ancestry** check. On a squash-merge repo — the release-please default across this portfolio — a squash collapses a branch into one fresh-SHA commit on the base, so the branch's own commits are never ancestors and a **fully-landed branch reads as unmerged**. Piping that into a delete is a delete loop keyed on a signal known to be wrong.

## Why this is a sweep, not a new finding

This is the *identical* defect already fixed for `git-plugin:deadbranch` in **#1869**, which measured only **4 of 28** actually-merged local branches flagged safe to delete. That fix landed on the **skill** and was never swept to the **agent** — where the idiom sits in the system prompt of every `git-ops` dispatch. Recorded in `.claude/rules/regression-testing.md:133`; the guard added there pins the skill body only.

`-d` refuses the unsafe deletes today, so nothing currently errors. That is what kept it invisible. The classification is still wrong, and the usual "fix" for a wall of `-d` refusals is `-D`, which deletes real work.

## The fix

Branch Cleanup now points at the encoded recipe, plus the two squash-surviving signals **in authority order**:

| Signal | Strength |
|---|---|
| `just -g branch-audit` | the encoded recipe — prints MERGED vs REVIEW + a paste-ready delete |
| `gh pr list --state all --head <branch> --json state,mergedAt` | **authoritative** — a MERGED PR settles it |
| `git cherry main <branch>` | `-` = patch-equivalent already upstream; survives squash **and** cherry-pick |

`git remote prune origin` is unaffected and kept.

## The guard

`scripts/check-agent-tool-selection.sh` gains two per-line shape rules:

- **(a)** a `git branch … --merged` **command at line start** — an instruction an agent will run
- **(b)** `--merged` composed with a `branch -d`/`-D` on one line — the xargs one-liner

It already walks `*/agents/*.md` and is already pre-commit-wired and blocking, so **no new wiring**. Per-line shape denylist is the accepted form for this class in this repo — #2009 deliberately did *not* migrate these lints to the tree-sitter helper, because they carry no fence/table state machine.

**Prose that merely names the command is not flagged.** That distinction is load-bearing: it is how the hazard gets taught, and the fixed file's own prose ("Never delete branches on `git branch --merged`") exercises it.

### Verified by mutation, not just on the fixed tree

```
=== guard on the full tree ===
All 21 agent files have a Tool Selection section and no ancestry-based branch cleanup. ✅  exit=0

=== verbatim PRE-FIX git-ops.md (git show origin/main:…) ===
❌ …: instructs ancestry-based branch cleanup.
     161:git branch --merged main | grep -v 'main\|master' | xargs git branch -d
exit=1

=== the FIXED file, whose prose still cites the command ===
exit=0
```

Allowlist prefix-matching is covered by a directory-prefix case of the same shape as the seed. `shellcheck` clean; `lint-shell-scripts.sh` reports 0 errors (the 9 warnings are pre-existing, in `macos-plugin`).

### Allowlist

`BRANCH_CLEANUP_ALLOWLIST` is seeded with `experiments/claude-probe/`, which uses the broken idiom **deliberately** as a trap fixture (`docs/trap-corpus.md`, `fixtures/trap-branch-merge/setup.sh` — the experiment tests whether a model notices that `--merged` omits a squash-merged branch). Default discovery only walks `*-plugin/agents/*.md`, so the probe is already out of reach; the entry exists so a future widening of discovery cannot silently break the experiment. Test seam: `CHECK_AGENT_TOOL_SELECTION_ALLOWLIST`.

## Notes

- `.claude/rules/regression-testing.md` gains one row in each table, **appended** — both are `merge=union`, and the diff is `2 insertions, 0 deletions`.
- Authored in an isolated worktree off fresh `origin/main`: the shared checkout carried stale working-tree copies of #2240's already-landed files, so it was left untouched rather than cleaned (`agent-coworker-detection.md`).

Refs #1869


laurigates:
## 🔧 Plugin Config Auto-Fix Applied

The plugin configuration files were out of sync. I've automatically fixed and pushed the changes.

<details>
<summary>Issues that were fixed</summary>

```
Checking plugin configuration sync...
Repository root: /home/runner/work/claude-plugins/claude-plugins

Found 1 issue(s):

  - Version mismatch for 'comfyui-plugin': manifest=1.12.0, marketplace=1.11.0

Run with --fix to automatically resolve these issues.
```
</details>

The workflow will re-run automatically. If it passes, you're good to merge!

PR #2260, opened 2026-07-31
chore(scripts): probe required-check reporting on a scripts-only PR

laurigates:
Throwaway probe, **will be closed unmerged**. Gate 2: a `scripts/*.sh`-only PR — the exact shape of Item 2's work, and the shape that received neither `compliance` nor `validate` on #2250/#2252/#2253.


laurigates:
Gate 2 verified: scripts-only PR (the exact shape of #2250/#2252/#2253, which received neither `compliance` nor `validate`) now receives all three with a conclusion. Closing unmerged; the probe commit is throwaway.

Issue #2219, opened 2026-07-29
fix(scripts): guards report OK without scanning anything — CI path gap plus worktree self-prune


laurigates:
## The class

Several `scripts/check-*.sh` guards can report `STATUS=OK` / exit 0 **without having scanned anything**. A guard that did not run and a guard that found nothing are indistinguishable in the output, so the green is a false one. Two independent mechanisms produce it, both verified 2026-07-29 at `b44da4c3`.

Same family as #2141 (a guard whose exit code CI never saw) and #2164 (checks that are vacuously green) — this is the *discovery-side* variant.

## Mechanism 1 — a guard's own change does not trigger the workflow that runs it

`.github/workflows/plugin-pr-checks.yml` runs `check-agent-model.sh`, `check-workflow-model.sh`, `check-loop-integrity.sh`, `check-agent-failure-contract.sh`, `check-blueprint-level3-templates.sh`, and `check-workflow-js-model.sh`. Its triggers:

```yaml
paths:
  - '*-plugin/**'
  - '.claude-plugin/marketplace.json'
  - '**/skills/**'
  - '.github/workflows/**'
  - '**/*.py'
  - 'ruff.toml'
```

There is **no `scripts/**`**. `**/*.py` catches Python under `scripts/`, but every guard listed above is a `.sh`. So a PR that changes only `scripts/check-*.sh` + `scripts/tests/test-*.sh` never runs them in CI.

Observed on three consecutive PRs that did exactly that — #2215, #2217, #2218 each registered **only** `conventional-commits` and a skipped `doc-audit`. Pre-commit did run the hooks locally (their `files:` regexes match), so the changes were verified — but CI was blind, and `.claude/rules/regression-testing.md` describes several of these as "wired into … `plugin-pr-checks.yml`", which reads as CI coverage it does not have for this case.

## Mechanism 2 — every model guard is a silent no-op inside an agent worktree

`check-agent-model.sh`, `check-workflow-js-model.sh`, and `check-version-pin-coverage.sh` all prune agent-worktree clones (#1492 / #1548):

```sh
find "$proj_dir" -path '*/.claude/worktrees/*' -prune -o …
```

Correct from the main checkout. But when `$proj_dir` **is** an agent worktree — `…/claude-plugins/.claude/worktrees/agent-<id>` — every descendant path contains `/.claude/worktrees/`, so the prune matches everything and the guard scans **zero** files:

```
[...]

PR #2277, opened 2026-08-06
fix(session-plugin): repair five guards that silently did nothing or reported a confident wrong answer

### LINKED ISSUE #2232: feedback(session-plugin): session-survey TASKWARRIOR section is project-scoped, so portfolio-wide sessions report OPEN_TASKS=0
## Skill

`session-plugin/scripts/session-survey.sh` — TASKWARRIOR section (and, downstream, `session-end`'s Taskwarrior-sync qualify gate)

## Category

Enhancement

## Description

The survey's taskwarrior query is scoped to the **detected** project:

```sh
# scripts/session-survey.sh:135
project_tasks_json=$("$task_bin" project:"$project" '(status:pending or +ACTIVE)' export 2>/dev/null || echo "[]")
```

`$project` comes from `DETECTION=cwd-repo-basename`. In a portfolio checkout that name is the *container* repo, not any project a task would plausibly be filed under — and a session whose work spanned many sibling repos files its tasks under those repos' project slugs. Result: `OPEN_TASKS=0`, so `session-end`'s Step 2 gate (`TASK_AVAILABLE=true AND OPEN_TASKS >= 1`) never fires and the Taskwarrior-sync pass is silently skipped in exactly the sessions that created the most tasks.

This is the same shape as #1915 (PRS section reported `PR_COUNT=0` alongside 7 open session-authored PRs, because detection was keyed to local branches rather than the remote): a survey section keyed to a local proxy for scope, reporting a confident zero.

## Evidence

Session `8aa0ba6f-1ef7-4b24-94fd-56932c42d7e6` (2026-07-29/30), an `ubuntu-slim` runner migration spanning 25 repos under `~/repos/laurigates`. `session-survey.sh --with-commits --with-blueprint` emitted:

```
=== PROJECT ===
PROJECT=repos
DETECTION=cwd-repo-basename
PROJECT_DIR=/Users/lgates/repos/laurigates
=== TASKWARRIOR ===
TASK_AVAILABLE=true
OPEN_TASKS=0
ACTIVE_TASKS=0
```

The session had created **5** tasks and closed 1, all under other projects:

```
28a8fea5  portfolio-ci  Verify ubuntu-slim works with anthropics/claude-code-action…
8854e95c  portfolio-ci  Run the ubuntu-slim runner evaluation on laurigates/dotfiles…
2731b0b2  gooho         gooho claude.yml: last 20 runs are all startup_failure…
b1c98b6c  kicad-mcp     kicad-mcp release-please.yml fails with "Bad credentials"…
```

`task export | jq '[.[]|select(.status=="pending" and .project=="repos")]|length'` → **0**. There is no project named `repos`; the top pending projects are `comfyui-nodes` (39), `skullcaps-native` (29), `mcu-tinkering-lab` (25), `dotfiles` (18), `claude-plugins` (14). So the scoped query cannot match anything from a `~/repos`-level session, and the zero is structural rather than incidental.

Nothing was actually lost here — I noticed the digest disagreed with tasks I had just created and checked the queue by hand — but the gate's whole purpose is to catch what the operator *doesn't* remember.

## Suggested Action

Two options, roughly in preference order:

1. **Widen the query when the detected project is a container/no-match**, and report both counts so the gate can use the right one:
   ```sh
   project_tasks_json=$("$task_bin" project:"$project" '(status:pending or +ACTIVE)' export 2>/dev/null || echo "[]")
   if [ "$(printf '%s' "$project_tasks_json" | jq 'length')" -eq 0 ]; then
     # fall back to session-recent tasks across all projects
     project_tasks_json=$("$task_bin" '(status:pending or +ACTIVE)' export 2>/dev/null || echo "[]")
     echo "TASK_SCOPE=all-projects-fallback"
   fi
   ```
2. **Emit an additional `RECENT_TASKS_*` block** scoped by `entry`/`modified` within the session window rather than by project, and gate Taskwarrior-sync on `OPEN_TASKS + RECENT_TASKS >= 1`. This also covers the inverse case (a single-repo session that filed a task under a different slug).

Either way, surfacing `TASK_SCOPE=` in the digest would make the narrowing visible instead of indistinguishable from an empty queue — the same fix-shape as #1915's "base it on the remote, not local branches".

https://claude.ai/code/session_01UuZtyXzArr3RU2APRD6Q31

### LINKED ISSUE #2271: session-survey.sh: project detected from directory basename reports 0 tasks when the slug differs, silently skipping taskwarrior sync
## Summary

`scripts/session-survey.sh` derives the taskwarrior project slug from the **repo directory basename**. When a repo's directory name differs from the project slug its tasks actually use, the survey reports **zero open tasks** while many exist — and because `session-end`'s qualify gate keys off that count, the Taskwarrior-sync pass is silently skipped.

## Reproduction

In the chezmoi source dir for `laurigates/dotfiles` (2026-08-04):

```
=== PROJECT ===
PROJECT=chezmoi
DETECTION=cwd-repo-basename
=== TASKWARRIOR ===
TASK_AVAILABLE=true
OPEN_TASKS=0
ACTIVE_TASKS=0
```

Reality:

```console
$ task project:dotfiles status:pending export | jq length
23
$ task project:chezmoi status:pending export | jq length
0
```

The directory is named `chezmoi` only because that's chezmoi's fixed source-dir path (`~/.local/share/chezmoi`) — the project is `dotfiles`, matching the repo name and the convention in the user-global taskwarrior-tracking rule ("one short lowercase slug per long-running area, reused consistently").

## Why this matters more than a wrong label

`session-end`'s Step 2 gate is literally:

> Taskwarrior sync — qualifies when `TASK_AVAILABLE=true` AND `OPEN_TASKS` ≥ 1

With `OPEN_TASKS=0` the pass is **silently skipped**, and the skill's own guidance is to skip non-qualifying passes without mentioning them. So the user sees nothing at all — it presents as "no tasks to sync" rather than "I looked under the wrong project". I only caught it because 23 tasks contradicting a `0` was implausible enough to check by hand.

## Who else hits this

Any repo whose directory name ≠ its task project slug:

- tool-mandated source dirs (chezmoi's `~/.local/share/chezmoi`)
- git worktrees (`myproj-feature-x`)
- monorepo subdirectories
- a repo cloned under a different local name

## Suggested direction

Offered as options, not a prescription — the maintainer knows the detection chain better:

1. **Cross-check before reporting zero.** If the basename yields 0 tasks, probe the git remote's repo name (`laurigates/dotfiles` → `dotfiles`) and prefer it if *that* yields tasks. Cheap, and only fires in the failing case.
2. **Make the uncertainty visible.** Emit something like `PROJECT_CONFIDENCE=low` or `OPEN_TASKS=unknown` so `session-end` can distinguish "genuinely nothing to sync" from "probably looked in the wrong place". This alone would remove the silence, which is the actually-harmful part.
3. **Honor a committed override.** The script already accepts `--project`, so the gap is only in *auto*-detection — a repo-local config key would close it for the deterministic cases.

Option 2 seems the highest value per unit of risk: it doesn't change detection, just stops a wrong answer from looking like a confident one.

## Context

Surfaced during a `/session-end` run. The other candidate from that session — `macos-plugin:macos-disk-usage`'s "Trust the number" `du` guidance being wrong for copy-on-write clones — is fixed in #2270 rather than filed here.

### LINKED ISSUE #2272: code-quality-preflight-cue fires mid edit-sequence, so the lint it demands runs against a half-applied refactor
## What happened

Mid-way through a planned multi-edit refactor (adding a `variant: "modal" | "inline"` option to a function in `comfyui-touch-numeric/src/index.ts`), the first `Edit` — the one introducing the option and its `const inline = variant === "inline"` binding — tripped the cue:

```
PostToolUse:Edit hook blocking error from command:
"bash ${CLAUDE_PLUGIN_ROOT}/hooks/code-quality-preflight-cue.sh":
[code-quality] Large/structural edit detected. Run /code-quality:code-lint as a pre-flight before continuing.
```

The three follow-up edits that *consume* `inline` had not landed yet. So the demanded pre-flight ran against a deliberately half-applied change and reported exactly the transient it had to report:

```
> 556 │   const inline = variant === "inline";
      │         ^^^^^^
  i Unused variables are often the result of typos, incomplete refactors, or other sources of bugs.
  i Unsafe fix: If this is intentional, prepend inline with an underscore.
```

The suggested fix (`_inline`) would have been actively wrong — three edits later the variable is used four times.

## Why this is a different axis from the prior five

`code-quality-preflight-cue` has a family of closed over-trigger reports, all resolved by gating on **what the file is**:

- #1905 — throwaway Writes under the session scratchpad
- #1825 — Signal 3 (≥50 lines) on YAML/infra-config code-lint can't lint
- #1766 — small standalone shell scripts
- #1730 — `.d2` / `.svg` diagram files

This one is not a file-type false positive. `src/index.ts` is a real TypeScript file in a real repo with biome configured — a completely legitimate lint target. The mismatch is **when**, not **what**: the cue fires on the *first* edit of a sequence, when the code is guaranteed to be in an intermediate state.

## Suggested shapes

Ordered cheapest-first; any one of them would have removed the noise here:

1. **Debounce on edit locality** — suppress the cue if another `Edit`/`Write` to the same file arrived within the last N tool calls, on the theory that a sequence is in flight. Re-arm once the file goes quiet.
2. **Defer to a Stop/idle boundary** — the cue's value is "don't pile up unlinted structural change", which is a property of the *turn*, not of an individual edit. A PostToolUse nudge on edit N of M is inherently early for N < M.
3. **Soften the message when it cannot know** — if it stays per-edit, phrase it as "run before you finish this sequence" rather than "before continuing", so a mid-sequence agent isn't pushed to lint a knowingly-incomplete tree.

## Impact

Low severity, non-blocking — I ran the pre-flight, read the expected error, and continued. The cost is a wasted lint cycle plus a suggested autofix that would have introduced a bug if applied. Worth noting the cue is otherwise *earning its keep* in this repo: the same session's later structural edits were genuinely worth linting, and the full gate caught real formatting drift before commit.

## Environment

- Session: comfyui-nodes mobile-UX wave, 2026-08-04
- Repo: `laurigates/comfyui-touch-numeric` (TypeScript + biome + bun)
- The edit sequence: one `Edit` adding the option + binding, three consuming it

https://claude.ai/code/session_01LDZzfMSnW2QLmcsNZogeVp

### LINKED ISSUE #2275: fix(hooks): updatedToolOutput emitted as string is rejected — bash-antipatterns-teach and blueprint-structural-cue are silent no-ops
## Summary

Two PostToolUse hooks emit `hookSpecificOutput.updatedToolOutput` as a **JSON string**. Claude Code validates that field against the *tool's own output shape*, which is an **object**. The harness rejects it and falls back to the original output, so **the hint is discarded every time and never reaches the model**.

These hooks are currently silent no-ops.

## Evidence

Scan of 50 transcripts / 15 project dirs / 7.5 days — every `hook_success` from these hooks is immediately followed by a `hook_error_during_execution`:

```
PostToolUse hook returned updatedToolOutput that does not match Write's output shape; using original output. [
  {
    "expected": "object",
    "code": "invalid_type",
    "path": [],
    "message": "Invalid input: expected object, received string"
  }
]
```

| Hook group | Discarded outputs in window |
|---|---|
| `PostToolUse:Bash` (`bash-antipatterns-teach.sh`) | 24 |
| `PostToolUse:Write` (`blueprint-structural-cue.sh`) | 3 |
| `PostToolUse:Edit` (`blueprint-structural-cue.sh`) | 2 |
| **Total** | **29** |

The scripts themselves are fine — measured at ~90–110 ms, well inside their 5000 ms timeouts. This is purely an output-contract mismatch.

## Root cause

Both build the envelope with `jq -n --arg out "$AUGMENTED"`, which makes `updatedToolOutput` a string:

`hooks-plugin/hooks/bash-antipatterns-teach.sh`
```bash
jq -n --arg out "$AUGMENTED" '{
    hookSpecificOutput: {
        hookEventName: "PostToolUse",
        updatedToolOutput: $out
    }
}'
```

`blueprint-plugin/hooks/blueprint-structural-cue.sh` — identical shape.

Observed tool output shapes:

- **Bash** — `{interrupted, isImage, noOutputExpected, stderr, stdout}`
- **Write/Edit** — `{content, filePath, originalFile, structuredPatch, type, userModified}`

### 1. `bash-antipatterns-teach.sh` — keep `updatedToolOutput`, emit an object

Bash has a natural free-text field (`stdout`). Merge into the original response object instead of replacing it with a string:

```bash
jq -n --argjson orig "$TOOL_RESPONSE" --arg out "$AUGMENTED" '{
    hookSpecificOutput: {
        hookEventName: "PostToolUse",
        updatedToolOutput: ($orig + {stdout: $out})
    }
}'
```

where `$TOOL_RESPONSE` is `.tool_response` read from the hook's stdin JSON as an object (not `tostring`).

### 2. `blueprint-structural-cue.sh` — switch mechanism

Write/Edit has **no** free-text field. `content` is the file's actual content; appending a `[blueprint]` banner there would corrupt what the model believes was written. So `updatedToolOutput` is the wrong mechanism for this event.

Use the block-with-reason form instead — which `code-quality-plugin/hooks/code-quality-preflight-cue.sh` already uses for the *same* event, and which works (it lands as `hook_blocking_error` with the reason delivered, no shape error):

```bash
jq -n --arg reason "[blueprint] $cue" '{"decision":"block","reason":$reason}'
```

with `continueOnBlock: true` in `hooks.json` (already set for the code-quality cue).

## Regression guard

Both plugins have hook test harnesses. Suggest asserting that `updatedToolOutput` parses as an **object** whose keys are a superset of the corresponding tool's output shape — a string-valued `updatedToolOutput` should fail the test.

## Notes

The `bash-antipatterns-teach.sh` header comments cite "Claude Code 2.1.121+" for `updatedToolOutput`. The string form may have been accepted then; it is rejected on 2.1.222. Worth pinning the contract in the test rather than the comment.

### LINKED ISSUE #2276: perf(session-plugin): session-spinup-nudge blocks session start for 2.8s p50 and times out entirely on 7% of runs
## Summary

`session-spinup-nudge.sh` blocks session start for **2.8 s at the median** and hits its 15 s timeout outright often enough to matter. When it times out the user gets **no briefing at all** — having waited the full 15 s for it.

The cost is almost entirely four **serial** `gh` network calls in `scripts/session-survey.sh`. Parallelising them and degrading gracefully on timeout addresses both the median and the tail.

## Measurements

Transcript scan, 50 sessions / 15 project dirs / 7.5 days (2026-07-28 → 2026-08-05), 42 recorded runs of this hook:

| Statistic | Value |
|---|---|
| min | 2,205 ms |
| **p50** | **2,809 ms** |
| p90 | 7,436 ms |
| max | 15,026 ms |
| runs > 5 s | 12 / 42 (29%) |
| runs > 10 s | 3 / 42 |
| **hard timeouts** | **3** (`timedOut: true`, `timeoutMs: 15000`, durations 15020 / 15022 / 15026 ms) |

For scale, in the same window this is the slowest thing in the session by an order of magnitude — the next-slowest recurring hooks are `bash-antipatterns-teach.sh` at 96 ms and `blueprint-structural-cue.sh` at 92 ms.

## Attribution

Timing each component the collector runs, on a warm connection from a normal repo:

| Call | Time |
|---|---|
| `gh auth status` | 376 ms |
| `gh pr list --author @me --state open` | 654 ms |
| `gh pr list --head <branch>` | 541 ms |
| `gh issue list --assignee @me --state open` | 744 ms |
| **network subtotal (serial)** | **2,315 ms** |
| `task project:… export` | 72 ms |
| `task +ACTIVE export` | 74 ms |
| `git status --porcelain` | 31 ms |
| **local subtotal** | **177 ms** |

2,315 ms of serial network plus ~180 ms of local work plus bash/jq startup reproduces the observed 2,809 ms p50 almost exactly. **Network is ~93% of the runtime; all of it is serial** (`session-survey.sh` lines 153, 165, 170, 187).

The tail follows from the same structure: no individual `gh` call is bounded, so one slow response consumes the whole 15 s budget and takes the entire briefing down with it.

### 1. Run the four `gh` calls in parallel

They are mutually independent. Backgrounding them into temp files and `wait`ing collapses 2,315 ms → ~750 ms (the slowest single call).

```bash
gh pr list --author @me --state open --json … > "$tmp/author" 2>/dev/null &
gh pr list --head "$git_branch"        --json … > "$tmp/head"   2>/dev/null &
gh issue list --assignee @me --state open --json … > "$tmp/assigned" 2>/dev/null &
wait
```

Single biggest win, smallest change. Note `parallel-safe-queries.md` (already cited in this script's own header) — each call is exit-0-on-empty, so a failed one degrades to `[]` exactly as it does today.

### 2. Drop the `gh auth status` probe

It costs a **network round trip purely to decide whether to make network round trips**. Issue the first real query instead and treat a non-zero exit or empty result as not-ready. Saves 376 ms and one RTT.

Preserve the `GH_READY=false` distinction in the output — the comment at line 147 is right that a false `0 issues` must never imply a clean state. The signal can be derived from the real call's exit status rather than from a separate probe.

### 3. Bound each network call individually

One hanging `gh` call currently eats the entire 15 s hook budget. A per-call bound (~4 s) means a single slow endpoint costs that section, not the whole briefing.

### 4. Degrade instead of dying — the important one

Today, timeout ⇒ **zero output** after a 15 s wait. That is the worst of both: maximum latency, no value. Three sessions in this window got exactly that.

The local sections cost ~180 ms and never touch the network. Emit them unconditionally, then enrich with GitHub data if it arrives in budget. A briefing that says "2 uncommitted files, 3 open tasks — GitHub data unavailable this run" is far better than nothing, and makes the timeout a degradation rather than a failure.

### 5. Optional: short-TTL cache

56 SessionStart firings in 7.5 days across 15 project dirs. A 60–120 s cache keyed on `repo + branch` would make rapid re-spinups (resume, re-open, worktree hop) effectively free. Lower priority than 1–4.

## Not a problem

Two things I checked that are already correct, so nobody re-investigates them:

- **Source filtering** — the hook exits early on anything but `startup|resume` (lines 19–22), so `/clear` does not pay this cost.
- **Per-session dedup** — the `~/.cache/claude-session-spinup-nudge/<session_id>` marker (lines 32–35) means it fires at most once per session.

The latency is purely per-firing cost, not over-firing.

## Suggested acceptance check

Re-measure p50 and the timeout count over a comparable window after the change. Target: p50 under 1 s, and zero runs that produce no output at all.

## Related

- #2271 and #2232 — same script, correctness axis (project detection, portfolio-wide scoping) rather than latency. No overlap in the code paths touched here.

laurigates:
Fixes a cluster of five open issues that share one failure shape: **a guard or briefing that silently does nothing, or reports a confident zero it never actually measured.** None of them errored, none showed up in any test, and all five were found by noticing that a number was implausible.

Closes #2276
Closes #2271
Closes #2232
Closes #2275
Closes #2272

## ⚠️ Release scope — needs a maintainer decision

This branch fixes code in **four** plugins, but a squash-merge lands the PR title as the single commit message, so only the scope in that title releases. As titled, `session-plugin` gets a patch bump and **`hooks-plugin`, `blueprint-plugin`, and `code-quality-plugin` do not.**

Three options, your call:
1. **Merge as-is** and cut the other three manually (or let the next change to each carry them).
2. **Retitle** on merge if you'd rather a different plugin take the bump.
3. **Ask me to split** this into four PRs — the file sets are disjoint, so it splits cleanly.

I did not pick for you because the answer changes what gets published.

## What was broken

| Issue | The silent failure | Fix |
|---|---|---|
| **#2275** | Two PostToolUse hooks emitted `updatedToolOutput` as a JSON **string**. The harness validates it against the *tool's own* output schema, rejected all 29 emissions in 7.5 days, and fell back to the original output. Both hooks were no-ops — at ~100ms each, so nothing looked wrong | Bash has a free-text field, so the teach hook merges `$orig + {stdout: …}`. Write/Edit has none (`content` is the file's real content — a banner there would make the model believe it was written to disk), so the blueprint cue moved to `decision:block` + `continueOnBlock` |
| **#2276** | `session-spinup-nudge` blocked session start **2.8s p50**, and on 3/42 runs hit the 15s timeout producing **zero output** — every `echo` sat after the network work. Four serial `gh` calls, the first of which was an auth probe: a round trip to decide whether to make round trips | Probe dropped (readiness derived from a real query), remaining calls parallel under per-call watchdogs, and all no-network sections emitted **first** so a kill truncates instead of erasing |
| **#2271 / #2232** | The taskwarrior query keyed off the repo **directory basename**, so a chezmoi source dir reported `OPEN_TASKS=0` against 23 real tasks, and a portfolio checkout structurally could not match. `session-end`'s gate then skipped the sync pass silently | One all-projects snapshot scoped in jq, with `TASK_SCOPE` / `PROJECT_CONFIDENCE` making a narrowed scope visible instead of indistinguishable from an empty queue |
| **#2272** | The lint cue fired on edit 1 of 4, so the pre-flight it demanded linted a half-applied refactor and suggested renaming a variable that three later edits use four times | Cue reworded to defer to sequence completion, plus a per-file debounce that doesn't consume the session's single cue |

## Two bugs the fixes introduced, and how they were caught

Both rounds were adversarially verified, and the verifiers earned their keep:

- **The `#2271` fix reintroduced `#2271`.** jq exact equality replaced `task project:X`, which is a **prefix/hierarchy** match — so `bluepad32.own` subproject tasks silently vanished, and outside the recency window they disappeared from the digest entirely. Exactly the silent skip the issue was filed about.
- **The `#2272` fix made the existing dedup test vacuous.** The verifier deleted the entire dedup block and the suite still reported **43 pass / 0 fail**.
- A third: switching the delimiter to US fixed an empty-column shift but `join()` escapes nothing where `@tsv` escaped by construction, so a **newline** in a description split one row into two — and in the fallback rows the task vanished. Same class, different byte.

None would have been caught by "do the new tests pass?". Each needed someone mutating the code to check the tests actually discriminate.

## Verification

- Every behavioural fix is pinned by an **executable** assertion proven to fail against pre-fix code by **mutation**, not inspection (`.claude/rules/regression-testing.md`). The field-sanitizer case was mutation-checked on both halves independently: dropping `\r\n` fails 5 assertions, dropping only ` ` fails 2.
- Guard-integrity counter-assertions throughout, so a fix can't degrade into "blank the column" or "always report low".
- `run-skill-script-tests.sh`: **85/86**. The one failure is `test-bash-antipatterns.sh`, which is **environmental and pre-existing** — `ast-grep` is absent in this container (`/usr/bin/sg` is util-linux's setgid tool), all 30 failures are `expected exit 2, got 0`, and both that hook and its test are untouched by this branch.
- `lint-shell-scripts.sh`: **0 errors** (9 warnings, all pre-existing in `macos-plugin`).
- `session-survey` suite 175/175; preflight-cue suite 56/56.

## Notes for review

- One claim in an intermediate report was **overstated and is not in this diff**: the debounce is backward-looking, so it cannot suppress edit 1 — which is the case #2272 filed. What resolves that case is the rewording, which the issue explicitly accepts as sufficient. The hook comments now say this plainly.
- The retired `blueprint_structural_cue_spec.sh` was a ShellSpec suite that **never ran anywhere** (shellspec isn't installed, isn't in CI, matches no glob) and asserted the broken shape. Replaced by an executing suite that ports all 17 detection signals.
- `#2276`'s "degrade, don't die" is verified by a **structural** assertion (the no-network section's line number precedes the network one), not a timing test — no timing test can express it.

https://claude.ai/code/session_01GFBW5k9GkWdYnwMaS8r1Sr

---
_Generated by [Claude Code](https://claude.ai/code/session_01GFBW5k9GkWdYnwMaS8r1Sr)_


laurigates:

laurigates:

laurigates:
`compliance` is red on `11f68ba`, and it **fails identically on the base branch** — I verified it on a pristine `origin/main` (`549712b`) worktree with none of this branch's changes: `Passed: 20  Failed: 1`, same assertion. Not caused by this PR; the other 7 checks are green.

It's also **a bug in the test, not in the guard.**

`scripts/tests/test-guard-worktree-root-discovery.sh` (arrived via #2290) does an unanchored substring match:

```sh
contains() { printf '%s' "$1" | grep -q -- "$2" && echo true || echo false; }
```

`check-version-pin-coverage.sh` correctly reports `FILES_SCANNED=1` from a worktree-shaped root — the #2219 fix works. But it also emits a sibling key, and `grep -q 'FILES_SCANNED=0'` matches **`TEMPLATE_FILES_SCANNED=0`**, so the negative assertion trips on a guard that is behaving correctly:

```
FILES_SCANNED=1            <- the assertion the test wants
TEMPLATE_FILES_SCANNED=0   <- what 'FILES_SCANNED=0' actually matches
```

Anchoring that one pattern fixes it (line 121):

```diff
-    "$([ "$(contains "$out" 'FILES_SCANNED=0')" = false ] && echo true || echo false)"
+    "$([ "$(contains "$out" '^FILES_SCANNED=0')" = false ] && echo true || echo false)"
```

`21 passed / 0 failed` with that change. I mutation-checked it so the anchor doesn't buy a false negative: forcing the guard to genuinely emit `FILES_SCANNED=0` still fails **both** assertions (`19 passed / 2 failed`), so detection is unweakened.

Two notes:
- `check-agent-model` asserts the full `AGENT_FILES_SCANNED=1` and `check-unused-bash-grant`'s `SKILLS_SCANNED=` has no suffix-colliding sibling, so this is the only assertion in that file with the flaw today. Any future guard emitting a `*_FILES_SCANNED=` sibling would hit it again, so anchoring the other KEY=VALUE lookups may be worth it.
- Per `structured-script-output.md` these are line-oriented `KEY=VALUE` reads, so anchoring matches the convention rather than working around it.

I haven't pushed this — it's a `ci(scripts)` fix unrelated to the five issues here, and folding it into a `fix(session-plugin)` PR would muddy the release scope. Happy to push it to this branch or open a separate PR, whichever you prefer. Until then this PR can't go green, since the failure is inherited from `main`.

The two `main` merges themselves are verified clean: the `merge=union` on `.claude/rules/regression-testing.md` added only new rows (1 then 7), left this branch's row intact, and produced no duplicates or conflict markers. Local suite on the merged tree: `session-survey` 175/175, `preflight-cue` 56/56, `lint-shell-scripts` 0 errors.

---

PR #2279, opened 2026-08-07
fix(macos-plugin): report scaling efficiency instead of gating on it

### LINKED ISSUE #2183: fix(macos-plugin): scaling-efficiency floor spuriously WARNs on asymmetric P+E Apple Silicon
`macos-plugin/skills/macos-performance-benchmark` sets a **70% multi-core scaling-efficiency floor** (`MACOS_PERF_SCALING_WARN_PCT` in `scripts/config.sh`). That floor assumes symmetric cores and is wrong for Apple Silicon's P+E topology.

## Measurement

A 10P+4E machine measured **24–27%** scaling efficiency via `openssl speed -multi 14`. That is not a hardware fault — the linear-scaling assumption simply does not hold when 4 of the 14 cores are efficiency cores running at a fraction of P-core throughput. The benchmark WARNs on every healthy asymmetric Mac.

## Fix options

1. Lower the default floor to something defensible for P+E (needs a measured baseline across a couple of machines to pick a number honestly).
2. Demote scaling efficiency from a PASS/WARN gate to an **info line** in `bench/cpu.sh`, since there is no threshold that is simultaneously meaningful for symmetric and asymmetric topologies.

(2) is likely correct: a metric whose "good" value depends on the P:E ratio of the machine under test is a reporting value, not a health gate.

laurigates:
Closes #2183

## What was broken

`macos-performance-benchmark` scored multi-core **scaling efficiency** against a 70% floor (`MACOS_PERF_SCALING_WARN_PCT` in `scripts/config.sh`, consumed by `scripts/bench/cpu.sh`):

```bash
scaling=$(( mc_mbs * 100 / (sc_mbs * NCPU) ))
if (( scaling < SCALING_WARN_PCT )); then warn ... else pass ...; fi
```

That ratio only means something when every core is equivalent. On Apple Silicon's P+E topology it isn't: 4 of a 10P+4E machine's 14 cores run at a fraction of P-core throughput, so a **healthy** machine measures 24–27% and WARNs on every run. Worse, `report/generate.sh` derives the run's Overall Status from any WARN row, so a correct machine was reported 🟡 degraded.

## The fix — option (2), demote to an info line

Option (1) (lower the floor) was rejected: no single number is simultaneously right for symmetric and asymmetric topologies, so the threshold is **unfixable, not mis-tuned**. A metric whose "good" value depends on the P:E ratio of the machine under test is a reporting value, not a health gate.

| Change | File |
|---|---|
| `report_scaling()` computes the ratio and emits `info` **only** — INFO rows already render in the report and are excluded from the PASS/WARN/FAIL counts, so the number stays visible and scores nothing | `scripts/lib/common.sh` |
| `core_topology()` annotates it with the P:E split from `hw.perflevel0/1.physicalcpu` (`… (cores: 10P+4E)`), degrading to nothing on Intel/non-Darwin, so the percentage is interpretable rather than mysterious | `scripts/lib/common.sh` |
| `MACOS_PERF_SCALING_WARN_PCT` deleted, replaced by a comment naming #2183 so an already-exported value reads as a documented no-op rather than a bug | `scripts/config.sh` |
| Delegates to `report_scaling`; the stale "single-core baseline not available" message on the unparseable-output branch was corrected too | `scripts/bench/cpu.sh` |
| Documents that the ratio is reported, never scored, with the 24–27% measurement | `SKILL.md` |

**One deliberate addition beyond a pure deletion:** removing the gate would have left the OpenSSL multi-thread section contributing no verdict at all. The *absolute* multi-core throughput is now scored via the suite's existing self-calibrating `score_hib` baseline (`bench_aes_mc_mbs`) — a per-machine, best-seen comparison that carries no cross-topology assumption — so the section keeps a real health gate on the one number that is meaningful.

## How the regression check pins it

`scripts/tests/test-scaling-efficiency-info.sh` (20 assertions, auto-discovered by `scripts/run-skill-script-tests.sh`) is **semantic, not syntactic**: it sources `lib/common.sh` into a sandbox run dir and **executes `report_scaling()`**, then asserts the emitted `summary.tsv` rows.

- 24% (the issue's real measurement) → exactly one **INFO** row, zero WARN/FAIL
- 0% (worst conceivable) → still cannot WARN or FAIL
- 92% (symmetric-like) → still cannot **PASS** — the metric is not a verdict in either direction
- degenerate inputs (empty / zero divisor / non-numeric) emit nothing and raise no `set -u` error
- **guard integrity**: a control run calls `warn`/`pass` directly and requires those rows to appear, so the "emits no WARN" assertions can't pass against a sandbox that never wrote the TSV (a bug the first draft of this test actually had)
- plus static pins that the floor stays removed

Mutation-verified twice:

| Mutation | Result |
|---|---|
| Verbatim pre-fix `config.sh` / `bench/cpu.sh` / `lib/common.sh` restored | **10 assertions fail**, exit 1 |
| A *differently spelled* floor reintroduced inside `report_scaling` (a bare `warn` when `scaling < 70`) — every static grep still green | **3 behavioural assertions fail**, exit 1 |

The second case is the point: a grep-only guard would have shipped that.

## Verification

- `bash macos-plugin/skills/macos-performance-benchmark/scripts/tests/test-scaling-efficiency-info.sh` → ALL PASS
- `bash macos-plugin/skills/macos-performance-benchmark/scripts/tests/test-macos-performance-benchmark.sh` → ALL PASS (existing suite unaffected)
- `bash scripts/run-skill-script-tests.sh` → `TOTAL=86 FAILED=0 STATUS=OK`
- `bash scripts/lint-shell-scripts.sh` → 0 errors (9 pre-existing `set -euo pipefail` warnings on this suite, deliberate per the SIGPIPE-141 note in `shell-scripting.md`)
- `shellcheck` clean; full pre-commit suite green
- `.claude/rules/regression-testing.md` Known Regressions row added


laurigates:

Issue #2255, opened 2026-07-31
chore(scripts): flip check-unused-bash-grant.sh to --strict (blocking)


laurigates:
## What

Flip `scripts/check-unused-bash-grant.sh` from advisory to blocking by adding `--strict` at its two call sites:

- `.pre-commit-config.yaml` → `entry: ./scripts/check-unused-bash-grant.sh --strict`
- `.github/workflows/plugin-pr-checks.yml` → `run: bash scripts/check-unused-bash-grant.sh --strict`

## Why it is a follow-up rather than part of #2254

#2254 landed the lint plus the triage in one pass: 5 genuinely-unused `Bash` grants removed, `ISSUE_COUNT=0`, **allowlist empty**. So the flip is a one-word change and the ratchet would hold today.

It is deliberately deferred one cycle so real use can surface a false positive I did not hit. The detector already produced **one** in its first run against the corpus (`comfy-subgraphs-app-mode` runs `<venv>/bin/python …` — a real command behind a placeholder path prefix), which is the class of miss worth waiting for. A blocking lint that false-positives blocks every commit touching a `SKILL.md`.

## Gate before flipping

```sh
bash scripts/check-unused-bash-grant.sh          # expect STATUS=OK ISSUE_COUNT=0
bash scripts/check-unused-bash-grant.sh --strict # expect exit 0
bash scripts/tests/test-check-unused-bash-grant.sh
```

Flip only while `ISSUE_COUNT=0` **and** `UNUSED_BASH_GRANT_ALLOWLIST` is still empty — a non-empty allowlist means the finding set was suppressed rather than resolved, and the ratchet would be locking in an exception instead of a clean state.

Refs #2254, #2250

PR #2281, opened 2026-08-07
fix(git-plugin): route branch containment to the authority ladder, not branch-audit

### LINKED ISSUE #2268: git-ops routes agents to branch-audit, whose REVIEW bucket measured 88-91% false on two repos
`git-plugin/agents/git-ops.md:170` tells agents **"Prefer the encoded recipe"** and routes them to `just -g branch-audit`. Ten lines below, the same file gives the correct manual fallback — MERGED PR is authoritative, then `git cherry` — and cites `~/.claude/rules/pr-merge-hazards.md` #1 on tree-containment being a one-way signal.

The recipe did not implement that ladder. **The agent's fallback was more correct than the tool it recommends.**

## Measured

Two independent repos, 2026-08-04:

| Repo | PRs | Branches | REVIEW rows | Actually landed | False |
|---|---|---|---|---|---|
| `laurigates/claude-plugins` | 1881 | 517 | 191 | 174 | **91%** |
| `ForumViriumHelsinki/infrastructure` | 1684 | 536 | 245 | 216 | **88%** |

Zero regressions in the other direction (nothing classified MERGED that had not landed).

## Two defects, both silent

1. **`merge-tree` used as the primary signal.** `pr-merge-hazards.md` #1 explicitly corrects this: once the base drifts over the same files, the trees differ for work that fully landed. The recipe's own comment claimed merge-tree "survives squash-merge AND later drift" — directly contradicting the rule. 132 branches lost this way on claude-plugins.
2. **`gh pr list --limit 500`** on repos with 1881 and 1684 PRs. Any branch whose merged PR falls outside the window reads `PR=—` → "no merged PR" → REVIEW. 32 branches on claude-plugins, e.g. #955 (Mar), #885, #1203, #1301.

Neither surfaces as an error. The failure mode is a REVIEW bucket the user cannot trust, which pushes toward either keeping hundreds of dead branches or deleting them unexamined.

## Fixed

`private_dot_config/just/git.just` (dotfiles) now checks in authority order — MERGED PR → `git cherry` → `merge-tree` (positive proof only) — with the PR window raised. Verified by diff-test against the old logic on `infrastructure`: 216 branches recovered from false REVIEW, no regressions.

## The structural point

`git-plugin` routes agents to a recipe living in a **private dotfiles repo**. claude-plugins can neither version nor regression-test its own preferred path, and this defect sat behind a comment asserting the opposite of the rule the same repo ships. Worth deciding one of:

- have `git-ops.md` state the authority ladder as the primary instruction and demote `branch-audit` to a convenience, or
- vendor an equivalent script into `scripts/` where `regression-testing.md` applies, or
- accept the split and add a caveat at the recommendation site.

`scripts/check-agent-tool-selection.sh` and `.claude/rules/regression-testing.md` also reference `branch-audit`, so the coupling is wider than the one agent file.

laurigates:
## What

`git-plugin/agents/git-ops.md:170` told agents **"Prefer the encoded recipe"** and routed them to `just -g branch-audit`. Ten lines below, the *same file* gave the correct manual ladder — a MERGED PR is authoritative, then `git cherry` — and cited `~/.claude/rules/pr-merge-hazards.md` #1 on tree-containment being a one-way signal. The recipe did not implement that ladder, so **the agent's documented fallback was more correct than the tool it recommended**.

This PR inverts the routing so the authority ladder is the primary instruction, and adds a regression guard so it cannot silently revert.

## Why

Measured 2026-08-04 (issue #2268), the recipe's REVIEW bucket was **~90% false** on two independent repos, with **zero** regressions in the other direction:

| Repo | PRs | REVIEW rows | Actually landed | False |
|---|---|---|---|---|
| `laurigates/claude-plugins` | 1881 | 191 | 174 | **91%** |
| `ForumViriumHelsinki/infrastructure` | 1684 | 245 | 216 | **88%** |

Two silent defects, neither of which surfaces as an error:

1. **`merge-tree` used as the primary signal.** Once the base drifts over the same files, merging an already-landed branch back would re-introduce its older versions, so the trees differ for work that *fully landed*. A non-match proves nothing. **132 branches** lost this way.
2. **`gh pr list --limit 500`** against repos holding 1881 and 1684 PRs. Any branch whose merged PR fell outside the page read as `PR=—` → "no merged PR" → REVIEW. **32 more branches** (e.g. #955, #885, #1203, #1301).

The failure mode is a REVIEW bucket the user cannot trust, which pushes toward either keeping hundreds of dead branches or deleting them unexamined.

## ⚠️ Out of scope: the `branch-audit` recipe itself is NOT fixed here

The `branch-audit` recipe lives in the user's **private dotfiles repo** (`private_dot_config/just/git.just`), not in this marketplace. **Nothing in this PR changes, verifies, or re-tests the recipe.** Issue #2268 reports that the dotfiles copy has already been corrected there (authority order + a raised PR window) — that work is separate from this PR and is not confirmed by anything here. Do not read a green build on this PR as evidence the recipe is fixed.

What this PR fixes is the **routing and guidance** that sends agents to it. This repo can neither version nor regression-test the recipe, which is the structural point the issue raises: any further recipe work, and any mechanism for keeping it correct, belongs in the dotfiles repo and should be tracked there.

Accordingly, `git-ops.md` still names the recipe — but only as a convenience carrying the measured caveat, so a reader learns the ~90% figure *at the mention* rather than trusting its REVIEW bucket.

## Changes

**`git-plugin/agents/git-ops.md`** — Branch Cleanup rewritten around the authority ladder:

```bash
# 1. Authoritative. A MERGED PR means the work landed, whatever the tree looks like now.
gh pr list --state all --head <branch> --json number,state,mergedAt
# 2. Patch-equivalence. Survives squash-merge AND cherry-pick.
git cherry main <branch>
# 3. Positive-containment shortcut ONLY.
git merge-tree --write-tree main <branch>
```

The internal contradiction is gone: the recipe paragraph now sits *below* the ladder and explains why the ladder is authoritative, instead of recommending a tool and then separately describing a better method with no connection between them.

**`git-plugin/skills/deadbranch/SKILL.md`** — carried the same inversion (Step 1.5 listed the `merge-tree` no-op as signal **1** and the MERGED-PR check as signal **2**). Reordered to the same ladder, with `git cherry` added as the middle rung. Both literals pinned by `plugin-compliance-check.sh` (#1869) survive.

**Pagination.** No `--limit 500` existed in this repo, but the trap is now named at the command that avoids it: `--head <branch>` is an **exact** per-branch query, unaffected by `gh pr list`'s 30-item default cap (`.claude/rules/gh-json-fields.md` § "Default list cap"). Both files say so, and the guard blocks reintroduction.

**`scripts/check-agent-tool-selection.sh`** — its remediation message also led with `just -g branch-audit`; now leads with the ladder and carries the caveat.

## New guard

`scripts/check-branch-containment-guidance.sh` (+ `scripts/tests/test-check-branch-containment-guidance.sh`, 26 assertions), six rules:

| Rule | Catches |
|---|---|
| `branch_audit_uncaveated` | a `branch-audit` mention missing both `REVIEW bucket` and `#2268` |
| `recipe_before_ladder` | `branch-audit` presented **above** the authoritative `gh pr list … --head` line |
| `merge_tree_before_pr` | `merge-tree` ranked above the MERGED-PR signal |
| `merge_tree_uncaveated` | merge-tree used for **branch** containment without the `positive-containment` token |
| `ladder_missing` | containment guidance with no ladder at all — this is what stops the two ordering rules going vacuous if that line is deleted |
| `paginated_containment` | a `gh pr list` with a bare `--limit N` **and** `state`/`mergedAt` in `--json` **and** no `--head` |
| `nothing_scanned` | plugin dirs present but zero markdown discovered — the scan misfired (#2219) |

**Deliberately narrow, so it does not become noise that gets disabled.** Rules 1–4 scan `git-plugin/**/*.md` only, so `.claude/rules/gh-json-fields.md` and `pr-merge-hazards.md` stay free to quote the broken forms while *teaching* them — the same instruction-vs-explanation discrimination `check-agent-tool-selection.sh` already makes. Rule 5 spans all plugins but fires only on a merged-state *determination*, so `gh pr list --state merged -L 30 --json title` (scope discovery) is not flagged. YAML frontmatter, blockquotes, and `# Wrong`-marked anti-examples are skipped everywhere.

**Verified by mutation on both sides.** Run against the *verbatim pre-fix* files it exits 1 naming `git-ops.md:173` and `deadbranch/SKILL.md:77`. Nine neutered rules each turn the suite red — including reverting the discovery walk to absolute paths, which (because this repo's agent worktrees live under `.claude/worktrees/`) pruned the **entire** corpus and reported a vacuous OK over zero files. That defect was found *by* the mutation sweep before the first commit, and is the same one #2290 independently fixed across four sibling guards — which is why TEST A asserts a non-empty corpus rather than only `STATUS=OK`, and why TEST I re-runs discovery from a worktree-shaped scan root.

**Complies with #2290's zero-scan standard.** Plugin dirs present but zero markdown discovered is `TYPE=nothing_scanned` / ERROR; a tree with no plugin dirs at all stays `SCANNED_EMPTY=true` / OK / exit 0, because a guard that errors on a legitimately empty corpus gets disabled.

## Rebased onto post-#2290 `main`

`.pre-commit-config.yaml` resolved **additively**, as expected — #2290's three hooks (`check-workflow-script-triggers`, `test-check-workflow-script-triggers`, `test-guard-worktree-root-discovery`) and its `--strict` flip on `check-unused-bash-grant` are all intact; this branch's diff against `main` adds only its own two hooks. `.claude/rules/regression-testing.md` union-merged per `.gitattributes`, both sides' rows surviving. Re-verified after the rebase rather than assumed.

`bash scripts/tests/test-guard-worktree-root-discovery.sh` reports **20 passed / 1 failed** on this branch — and **also on unmodified `main`**: both that test and `check-version-pin-coverage.sh` are byte-identical to `origin/main` here. It is a substring false positive in the test, not a guard defect — `contains()` is an unanchored `grep -q`, so its `does not report FILES_SCANNED=0` assertion matches inside the legitimate `TEMPLATE_FILES_SCANNED=0`. Running the guard directly against the same fixture gives `FILES_SCANNED=1 … TEMPLATE_FILES_SCANNED=0 … STATUS=OK`. Flagged as a separate follow-up; deliberately not fixed here.

## Verification

All re-run after the rebase.

| Check | Result |
|---|---|
| `scripts/check-branch-containment-guidance.sh` | `STATUS=OK`, 50 git-plugin + 714 plugin md files scanned, `SCANNED_EMPTY=false` |
| `scripts/tests/test-check-branch-containment-guidance.sh` | 38 passed, 0 failed |
| Mutation sweep (9 neutered rules) | all 9 killed the suite |
| `scripts/run-skill-script-tests.sh` | `TOTAL=86 FAILED=0 STATUS=OK` |
| `scripts/lint-shell-scripts.sh` | 0 errors (9 pre-existing `macos-plugin` warnings, untouched) |
| `shellcheck` on both new scripts | clean |
| `scripts/check-agent-tool-selection.sh` | all 21 agent files pass |
| `scripts/check-unused-bash-grant.sh --strict` (#2255) | `STATUS=OK ISSUE_COUNT=0`, 408 skills / 186 grantees |
| `plugin-compliance-check.sh git-plugin` | Body ✅ (deadbranch's `#1869` pins intact) |
| `pre-commit` full run | passed |

## CI wiring

Wired into `.pre-commit-config.yaml` **only**. `.github/workflows/plugin-pr-checks.yml` was deliberately left untouched (another agent holds that file this session) — CI wiring is deferred to #2219.

Closes #2268


laurigates:

laurigates:

Issue #2224, opened 2026-07-29
feat(comfyui-plugin): evaluate porting comfyui-node-scaffold (3,045 lines) to cargo-generate


laurigates:
The larger of the two hand-rolled Python template engines. **Gated on the foundry pilot paying off (#2223)** — deliberately second, and worth abandoning if the pilot's answer is negative.

## Why it was not the pilot

| | `foundryvtt-module-scaffold` (piloted) | `comfyui-node-scaffold` (this) |
|---|---|---|
| Generator size | 1,279 lines | **3,045 lines** |
| Variants | 3 (`basic`/`app`/`libwrapper`) | **4** (`frontend`/`backend`/`gesture`/`shim`) |
| Extra axis | none | **`--widgets`** — presence/absence emits a *structurally different* `src/index.ts` (widget-intercept vs standalone-modal) |
| Existing tests | none | **3** (`test-finishing-pass.sh`, `test-shim-variant.sh`, `test-standalone-variant.sh`) |

The `--widgets` switch is the interesting part: it is not an additive variant like foundry's, it swaps the entry point's whole shape. `[conditional]` handles it (4 variants × modal-shape is still a small matrix), but it is a genuinely harder mapping than anything the pilot exercised.

## The payoff, if it holds

Same as the pilot's thesis: the emitted TypeScript becomes real files `tsc` and `biome` can check, instead of Python string literals nothing can. That argument is *stronger* here — this generator emits more TS, and `scaffold.py` already carries a `print_finishing_pass_audit` precisely because parts of its output cannot be verified mechanically (see #1877).

## What to preserve

1. **The three existing tests are the contract**, not optional extras. #1877 exists because a CI-green pack shipped with `Icon = ""` and a bare README TODO; `test-finishing-pass.sh` is what stops that recurring. Any port keeps it passing or replaces it with something at least as strong.
2. **The finishing-pass audit** — the port must still print it, or move its checks into the template where they become structural. This is the better outcome if reachable: an emitted `icon.svg` in a template is a real file, not a generated string.
3. **A parity test first**, same shape as `test-template-parity.sh`, including the all-non-default half. That half caught two hardcoded fields in the foundry pilot that a defaults-only run passed over.

## Costs specific to this pack

- ComfyUI pack authors are Python/TypeScript people; requiring `cargo-generate` is a bigger ask than it was for the foundry repos. Weigh this before starting, not after.
[...]

Issue #2223, opened 2026-07-29
feat(foundryvtt-plugin): promote the cargo-generate template to default and retire scaffold.py


laurigates:
Tracking issue for the second half of the #2205 pilot. **Do not start this until the exit criterion below is met** — the pilot deliberately shipped both generators with `scaffold.py` as the default.

## Exit criterion

> The template has generated **at least one real FoundryVTT module repo end to end** — scaffolded, `bun install`, `just check` green, pushed, release-please wired, module loads in Foundry.

Byte-identical parity against `scaffold.py` across all three variants is already proven mechanically (`test-template-parity.sh`, 9/9). What is *not* proven is that anyone has lived with the cargo-generate flow for a real module: the install step, the `--define` ergonomics, the prompts when values are omitted, whether the Rhai pre-hook's kebab-case handling is pleasant in practice.

## Blocked on

- #2221 — the parity test does not currently run in CI. Retiring `scaffold.py` while the only thing proving equivalence is a local-only test would be backwards.

## When the criterion is met

1. Flip the default in `foundryvtt-module-scaffold/SKILL.md` — template first, `scaffold.py` as the documented fallback.
2. Delete `scaffold.py` (1,279 lines) once no skill references it.
3. `test-template-parity.sh` loses its comparison half and becomes a **template-only smoke test** — generate each variant, assert the tree shape and that `module.json` parses. Keep it; the variants are still worth pinning.
4. Update `foundryvtt-plugin/README.md` and `templates/README.md` (both currently say "`scaffold.py` remains the default" in as many words).
5. `cargo install cargo-generate --locked` becomes a hard prerequisite of the skill rather than an alternative path — state it in the skill's requirements, not a footnote.

## If the criterion is *not* met

Deleting the template is a perfectly good outcome. The pilot's value was answering the question; a negative answer is an answer. The cost of keeping both indefinitely is the parity test plus the cargo-generate dependency, and that is not free.

The most likely reason for a negative result is the toolchain cost: requiring a Rust toolchain (or a release-binary install) to scaffold a TypeScript repo is the one genuinely awkward part of the port, and it is the thing to weigh hardest.

## Related

- #2205 — the pilot
- #2221 — CI enforcement of the parity claim (blocker)
- #2224 — the comfyui port, which is gated on this one paying off
- `foundryvtt-plugin/templates/README.md` — comparison, divergences, brace-collision fixes

laurigates:
[...]

Issue #1417, opened 2026-05-26
feedback(taskwarrior-plugin): task-add should emit UUID to prevent numeric-ID-shift footgun on subsequent ops


laurigates:
## Skill

`taskwarrior-plugin/skills/task-add/SKILL.md` (and any related task-* skill that returns or surfaces the freshly-created ID)

## Category

Enhancement

## Description

When `task add` (the bare CLI or the wrapping skill) creates a new task, it emits the numeric ID assigned at creation time — e.g. `Created task 141.`. The user-global rule `.claude/rules/taskwarrior-bulk-operations.md` explicitly warns:

> Taskwarrior numeric IDs are a display index over **pending** tasks. The moment you complete one task, every higher ID shifts down by one.

But the `task-add` skill (and its bare CLI surface) doesn't include the **stable UUID** in its output. A session that:

1. Creates task X → gets numeric ID `141`
2. Does work that completes other tasks in the global queue
3. Later annotates `task 141 ...` thinking it targets the same task it created earlier

…will silently annotate the wrong task. The numeric ID has shifted to point at a different task in the global queue.

This bit me in a recent session: I created an FR-026 slice 2.5 WO and got ID 141, then 30 minutes later (during which other tasks completed in the global queue) tried to annotate task 141 — and the annotation landed on an unrelated `R4C-Cesium-Viewer` task. Had to denotate + re-annotate via UUID lookup to fix.

## Evidence

The session transcript:

```
$ task add project:skullcaps-native +wo bpid:FR-026 ... "FR-026 slice 2.5: ..."
Created task 141.
The project 'skullcaps-native' has changed.

... (subsequent dispatch/wait cycle ~30 min, other taskwarrior tasks completed in parallel sessions) ...

$ task 141 annotate "commit 7498a05 — ..."
Annotating task 141 '+follow-up Verify #784 (cross-view nav tablet a11y) clears ...'
                            ^^^^^^^^^^^^^^^^^^ wrong task — R4C-Cesium-Viewer
```

The fix required:

```
$ task project:skullcaps-native status:pending export | \
    jq -r '.[] | select(.description | test("slice 2.5: extend autoload")) | .uuid'
d14a6e5e-1c60-4cfd-9dd0-8a9fe7659b74

$ task <uuid-of-cesium-task> denotate "commit 7498a05"   # remove wrong annotation
$ task d14a6e5e-... annotate "..."                        # apply to correct task
```

[...]

PR #2282, opened 2026-08-07
fix(ci): make the template-parity gate actually run, and stop skips reading as passes

### LINKED ISSUE #2221: fix(ci): test-template-parity.sh never runs in CI — cargo-generate absent and templates/ not a path trigger
`foundryvtt-plugin/skills/foundryvtt-module-scaffold/scripts/tests/test-template-parity.sh` (added in #2205) is the acceptance gate for carrying two generators — it runs `scaffold.py` and the cargo-generate template over the same inputs and requires byte-identical trees. **It has never actually run in CI.** Two independent causes, either of which alone is sufficient:

## 1. `cargo-generate` is not installed on the runner

The test SKIPs (exit 0) when the binary is absent — deliberate, so it degrades gracefully on a contributor's machine. But `.github/workflows/test-skill-scripts.yml` installs `ast-grep` and `uv` for exactly this reason and installs nothing for cargo-generate, so the SKIP branch is the *only* branch CI ever takes.

The SKIP is silent in the aggregate: `run-skill-script-tests.sh` reports `PASS=` for a skipped test, so the CI log reads as if the parity gate ran.

## 2. The template directory is not a path trigger

`test-skill-scripts.yml` triggers on `**/skills/**/scripts/**` among others. The test matches. **The template does not** — it lives at `foundryvtt-plugin/templates/foundryvtt-module/**`, which matches no trigger in either the `pull_request` or `push` list.

So even with cargo-generate installed, a PR that edits *only* the template — precisely the change the parity test exists to catch — would not run the workflow at all.

## Why it matters

The whole justification for keeping `scaffold.py` and the template side by side is that divergence gets caught mechanically. Right now it's caught only when someone runs the test locally with cargo-generate installed. Two generators with an unenforced parity claim is worse than one generator.

## Verification

```
$ grep -c 'cargo-generate\|cargo install' .github/workflows/test-skill-scripts.yml
0
$ env PATH=/usr/bin:/bin bash foundryvtt-plugin/skills/foundryvtt-module-scaffold/scripts/tests/test-template-parity.sh
SKIP: cargo-generate not available (cargo install cargo-generate --locked)
$ echo $?
0
```

## Suggested fix

1. Add a `cargo-generate` install step to `test-skill-scripts.yml`, following the existing `ast-grep` / `setup-uv` precedent. Prefer the GitHub release binary over `cargo install` — a from-source build took **several minutes** in this sandbox, which is a real per-run cost for one test.
2. Add `foundryvtt-plugin/templates/**` (or a general `*-plugin/templates/**`) to both the `pull_request` and `push` path trigger lists.
3. Consider whether a silent SKIP should be visible in the runner summary at all — `run-skill-script-tests.sh` currently reports skipped tests as `PASS=`, which is how this stayed invisible.

Fixing (2) alone is cheap and worth doing regardless. (1) is the one that actually turns the gate on.

## Related

- #2205 — the PR that added the test and the template
- `blueprint-plugin/templates/*.workflow.yml` (#2005) is the established precedent for "a template outside the linted path needs its own guard wiring"
- `.claude/rules/regression-testing.md` — a guard that cannot fire is not a guard

laurigates:
Closes #2221. Sibling of #2219 (`a guard that reports OK without scanning anything`) — same class, different surface: this one is *a guard that reports PASS without running anything*. #2219 touches `plugin-pr-checks.yml` / `.pre-commit-config.yaml` / `scripts/check-*.sh`; this PR deliberately touches none of those. They should land as a pair.

### 1. `cargo-generate` is installed on the runner

`.github/workflows/test-skill-scripts.yml` already installs `ast-grep` and `uv` so their gates actually run; it installed nothing for `cargo-generate`, so `test-template-parity.sh`'s deliberate SKIP branch was the only branch CI ever took.

**Install method: prebuilt release tarball, not `cargo install cargo-generate --locked`.** The from-source build compiles several hundred crates — minutes on every run, for one test. Alternatives considered and rejected:

- `taiki-e/install-action` — checked its `TOOLS.md` at HEAD; **cargo-generate is not in its tool list** (0 matches).
- A third-party installer action — would add a `uses:` whose behaviour has to be audited at its pinned ref per `external-contributor-prs.md` / `composite-actions-that-need-docker.md`. A four-line `curl | tar` in a `run:` block is auditable in place and adds no action.

The version is pinned in a top-level `env:` block with a `# renovate:` annotation, and `renovate.json` gains a customManager for annotated `<TOOL>_VERSION:` pins in `.github/workflows/*.yml` — mirroring the existing `scripts/install_pkgs.sh` manager, so the pin cannot silently rot. Cached via `actions/cache` keyed on version + os + arch.

**One honest caveat:** upstream publishes no checksum file alongside these release assets (verified against `releases/latest`), so the pinned version tag is the only integrity anchor. A hardcoded sha256 would break on every Renovate bump.

### 2. `*-plugin/templates/**` is a path trigger

Added to **both** the `pull_request` and `push` lists. Kept generic rather than `foundryvtt-plugin/templates/**` — other plugins will grow template directories (#2224 evaluates the same shape for comfyui), and this is the trigger that makes "edit only the template" run the gate that exists to catch template drift.

### 3. A skip no longer reads as a pass — the real defect

`scripts/run-skill-script-tests.sh` printed `PASS=<file>` for a skipped test, so a gate that did not run and a gate that passed were indistinguishable in the log *and* in the `TOTAL=`/`FAILED=` roll-up. That is what kept both causes above invisible for weeks.

The runner now follows `structured-script-output.md`:

```
SKIP=<path>
TOTAL= / PASSED= / SKIPPED= / FAILED=
REQUIRED_DECLARED= / REQUIRED_VIOLATIONS=
STATUS=OK|WARN|ERROR
ISSUE_COUNT=
SKIPS:
  - <path> (<reason>)
ISSUES:
  - SEVERITY=ERROR TYPE=required_test_skipped TEST=... MSG=...
```

**Detection rule:** exited 0 **and** every *non-indented* non-empty line is a `SKIP:` / `SKIP -` notice. Both punctuation forms exist in the corpus (`blueprint-plugin` uses `SKIP - `), and the non-indented qualifier is load-bearing for `hooks-plugin/hooks/test-bash-antipatterns.sh`, which emits an indented `      Install: npm install -g @ast-grep/cli` continuation line. A **partial** skip inside an otherwise-executing suite correctly stays a `PASS`.

**Making CI fail on a skip that shouldn't happen** needed a declared expectation, not a blanket rule — fail-on-any-skip would be red on arrival for optional deps (taskwarrior, lizard) and get disabled, which is the `generated-fleet-drift.md` trap. So `scripts/required-to-run-tests.txt` declares the tests the workflow installs a dependency for **on purpose**; a skip there is `STATUS=ERROR` + exit 1, and so is a listed path that matches no discovered test (the manifest cannot silently drift from the corpus it declares). Everything else may skip freely and only warns.

## Does the parity gate actually pass?

**Yes — 9/9, verified by running it.** Installed `cargo-generate` 0.23.14 locally and ran the real test:

```
=== TEMPLATE PARITY ===
=== NAME HANDLING ===
=== SUMMARY ===
PASS_COUNT=9
FAIL_COUNT=0
STATUS=OK
```

The issue asserted this mechanically but nobody had watched it run, so it was treated as unverified. It holds: the gate was correct all along, it simply never executed. **#2223 (retiring `scaffold.py`) can rely on it once this lands** — but not before, since until now the parity claim was unenforced.

## Verification

| Check | Result |
|---|---|
| `test-template-parity.sh` with cargo-generate present | 9/9, `STATUS=OK` |
| `bash scripts/run-skill-script-tests.sh` | `TOTAL=85 PASSED=85 SKIPPED=0 FAILED=0 REQUIRED_DECLARED=5 REQUIRED_VIOLATIONS=0 STATUS=OK`, exit 0 |
| `bash scripts/lint-shell-scripts.sh` | 0 errors (9 pre-existing WARNs in `macos-plugin`, untouched) |
| `shellcheck` on both changed shell files | clean |
| `actionlint` on the workflow (embedded shellcheck on) | clean |
| `check-workflow-model.sh` | `STATUS=OK` |
| `pre-commit` on the commit | all hooks pass, including the new one |

### Before/after on the real test, cargo-generate removed from `PATH`

Pre-fix runner:

```
PASS=/tmp/.../test-template-parity.sh
TOTAL=1
FAILED=0
STATUS=OK           EXIT=0
```

Post-fix runner:

```
SKIP=/tmp/.../test-template-parity.sh
TOTAL=1  PASSED=0  SKIPPED=1  FAILED=0
REQUIRED_VIOLATIONS=1
STATUS=ERROR  ISSUE_COUNT=1   EXIT=1
ISSUES:
  - SEVERITY=ERROR TYPE=required_test_skipped TEST=foundryvtt-plugin/.../test-template-parity.sh MSG=cargo-generate not available ...
```

## Regression check

`scripts/tests/test-run-skill-script-tests.sh` (40 assertions) is **semantic**: it executes the real runner against planted fixture trees and asserts on its structured output. A grep for `SKIP=` inside the runner would pass against a half-fix that never classifies anything as a skip.

It pins both skip punctuation forms, the indented-continuation shape, a stderr-only skip, required-skip → ERROR, manifest drift → `required_test_missing`, a failure outranking the skip WARN, empty corpus → OK/exit 0, unknown argument → exit 2 (#2057), and `.claude/worktrees/` pruning.

**Guard integrity** — without these, every "not counted as a pass" assertion is vacuous: `PASSED=2` must hold (a runner classifying *everything* as a skip would otherwise sail through), a partial skip must stay a `PASS=` and must not appear as `SKIP=`, a required test that passes must raise nothing, and the repo's own manifest must be non-empty with every entry present on disk.

**Mutation-verified** against the pre-fix runner (`git show HEAD:scripts/run-skill-script-tests.sh`): **24 of 40 assertions fail**, including the headline `colon-form skip NOT reported as PASS=`.

Wired into `.pre-commit-config.yaml` and as its own step in `test-skill-scripts.yml` — repo-root `scripts/tests/` is not covered by the runner's own discovery globs.

## Files changed

| File | Change |
|---|---|
| `.github/workflows/test-skill-scripts.yml` | cargo-generate install + cache, `*-plugin/templates/**` triggers, new self-test step |
| `scripts/run-skill-script-tests.sh` | SKIP classification, structured output, required-test enforcement, fail-fast arg parsing |
| `scripts/required-to-run-tests.txt` | **new** — declares the 5 tests whose deps CI installs on purpose |
| `scripts/tests/test-run-skill-script-tests.sh` | **new** — 40-assertion semantic regression test |
| `.pre-commit-config.yaml` | wires the new regression test |
| `renovate.json` | customManager for annotated `<TOOL>_VERSION:` pins in workflows |
| `foundryvtt-plugin/.../test-template-parity.sh` | header comment: the SKIP is fine locally, an error in CI |
| `foundryvtt-plugin/templates/README.md`, `.../foundryvtt-module-scaffold/SKILL.md` | docs land in the same commit (`docs-currency.md`) |


laurigates:
## CI verification — the gate now demonstrably runs

The first push proved the point in the most direct way available: **the workflow failed**, because the install step 404'd. Upstream ships a musl tarball for `aarch64` only, so `cargo-generate-v0.23.14-x86_64-unknown-linux-musl.tar.gz` does not exist. Fixed to `-gnu` in 5b0ef99e.

That failure is worth noting rather than hiding — it is the first time this gate has ever executed in CI, so it is the first time a broken install could surface at all. Under the old runner a broken install was indistinguishable from a working one.

Green run ([31154344017](https://github.com/laurigates/claude-plugins/actions/runs/31154344017)):

```
Put cargo-generate on PATH            cargo generate 0.23.14
Run skill-local regression tests      PASS=./foundryvtt-plugin/skills/foundryvtt-module-scaffold/scripts/tests/test-template-parity.sh
                                      TOTAL=85  PASSED=83  SKIPPED=2
                                      REQUIRED_DECLARED=5  REQUIRED_VIOLATIONS=0
                                      STATUS=WARN
```

`PASS=` on that line is the whole point of the PR — **the parity gate ran, in CI, for the first time.**

## Two previously-invisible skips, now surfaced

`SKIPPED=2` is a new finding, not a regression. These two ran nowhere and reported `PASS=` before:

```
SKIPS:
  - taskwarrior-plugin/scripts/tests/test-task-id-stability.sh (task CLI not available)
  - taskwarrior-plugin/skills/task-add/scripts/tests/test-uuid-capture.sh (task CLI not available)
```

Correctly `STATUS=WARN` / exit 0 — the `task` CLI is not on the runner, they are not in `required-to-run-tests.txt`, and forcing them to ERROR is exactly the red-on-arrival trap the manifest exists to avoid.

Worth a follow-up, though, and deliberately **out of scope here** (it is a taskwarrior-plugin decision, not a CI-plumbing one): `test-uuid-capture.sh` is the *executable* semantic gate written for #1417/#1819 specifically because a syntactic pin had let a broken `task +LATEST _get uuid` ship. It has never run in CI either. Either install `task` on the runner and add both to the manifest, or record that they are local-only by design.

laurigates:

laurigates:

laurigates:
## Rebased onto `main` (3 commits, force-pushed by explicit SHA `5406a885`)

Both conflicts resolved **additively**. Re-verified rather than assumed — a clean `git rebase` is a claim about mechanics, not content.

### `.pre-commit-config.yaml` — additive, verified two ways

`git diff origin/main -- .pre-commit-config.yaml` is **exactly my 6 added lines**, nothing removed. Direct inspection confirms #2290's side intact: `check-workflow-script-triggers` (L134), `test-check-workflow-script-triggers` (L140), `test-guard-worktree-root-discovery` (L146) sit immediately before `check-auto-resolve-mergeability-poll` (L152), and `check-unused-bash-grant` keeps its `--strict` (L106).

### `renovate.json` — both mechanisms live, and both still *extract*

#2287's top-level `"github-actions"` key and my `customManagers[5]` are different mechanisms over different files; both kept.

**Validator:** `renovate-config-validator` at **44.14.3** → `Config validated successfully`. Worth recording why the first attempt was misleading: `npx --package renovate` resolved **37.440.7**, which predates `managerFilePatterns` entirely and rejected *all six* customManagers plus #2287's key. The control was decisive — entries `[0]`–`[4]` are untouched and already on `main`, and failed identically. A stale validator, not a broken config.

**Extract (`--platform=local --dry-run=extract`)** — because a config that parses is not a config that extracts:

```json
{
  "depName": "cargo-generate/cargo-generate",
  "currentValue": "0.23.14",
  "datasource": "github-releases",
  "extractVersion": "^v(?<version>.*)$",
  "packageFile": ".github/workflows/test-skill-scripts.yml"
}
```

`extractVersion` lands as its own field rather than being swallowed into `depName`. #2287's widening still works too — `blueprint-plugin/templates/*.workflow.yml` and `foundryvtt-plugin/templates/foundryvtt-module/.github/workflows/*.yml` all appear as `packageFile`s. **No double-extraction:** the dep appears twice in the log, but so does the untouched `casey/just` pin from `install_pkgs.sh` — Renovate prints the extract set twice. One record per pin.

### Denominator invariant — added (the cheap half)

`TOTAL=0` used to exit `STATUS=OK`, which is the same hole in the other direction: a discovery collapse looked like a clean run over 86 tests. Now emits `SCANNED_EMPTY=true|false` — **the same key name and `true`/`false` convention #2290 used** for the `check-*.sh` guards — and downgrades an empty corpus to `WARN`. Exit stays 0, preserving the documented greenfield contract.

The *hard* failure comes from the manifest rather than a blanket rule: if discovery collapses, every declared entry raises `required_test_missing`, so with 5 declared tests a `TOTAL=0` exits 1. Structurally the same as #2290's `nothing_scanned` ERROR — a population signal contradicting a zero denominator — with the manifest as the population signal.

### Verification after rebase

| Check | Result |
|---|---|
| `run-skill-script-tests.sh` | `TOTAL=86 PASSED=86 SKIPPED=0 FAILED=0 REQUIRED_VIOLATIONS=0 SCANNED_EMPTY=false STATUS=OK` |
| `test-run-skill-script-tests.sh` | 44/44 (was 40; +4 for the denominator case) |
| `shellcheck`, `actionlint` | clean |
| #2290's `check-workflow-script-triggers.sh` | `STATUS=OK` — independently confirms my trigger-list additions cover the scripts the workflow invokes |
| CI `test` (Test: Skill scripts) | SUCCESS |

## The red `compliance` check is pre-existing on `main` — not this PR

`compliance` fails on `scripts/tests/test-guard-worktree-root-discovery.sh`:

```
FAIL: check-version-pin-coverage: does not report FILES_SCANNED=0
```

Three independent confirmations that this is not mine:

1. **File identity** — `git diff origin/main` touches none of `check-version-pin-coverage.sh`, `test-guard-worktree-root-discovery.sh`, or `test-check-version-pin-coverage.sh`. The hook reads only bytes identical to `main`.
2. **Control PR** — #2289 (`fix/code-lint-unrendered-conditionals-2265`, entirely unrelated) fails with the *identical* message.
3. **Timing** — #2293's compliance run passed at 06:58 UTC; #2290 merged at 07:39 UTC. Every compliance run *after* that merge fails.

So **#2290 left one of its four guards unfixed and `compliance` is red for every PR against `main`.** `mergeStateStatus=BLOCKED` here is that, not a defect in this branch. This is #2219/#2290's surface and I was scoped out of `scripts/check-*.sh`, so I have not touched it.

## Follow-up I did *not* do (deliberately)

`scripts/run-skill-script-tests.sh` has the **same latent worktree-root defect #2290 just fixed in four sibling guards** — and `test-guard-worktree-root-discovery.sh` does not cover it:

```bash
find "$root_dir" -path '*/.claude/worktrees/*' -prune -o ...
```

When `root_dir` is itself an agent worktree, its own path contains `/.claude/worktrees/`, so the prune matches the scan root and discovery returns **zero**. Latent today: the justfile and CI both invoke it with the default relative `.`, which cannot match. An absolute-root invocation from inside a worktree would silently find nothing — though the required-manifest guard added here now converts that into an ERROR rather than a green `TOTAL=0`, so the blast radius is already reduced.

Not fixed here because #2290's fix changes discovery to relative paths, which changes the emitted `PASS=`/`SKIP=` path form and ripples through ~12 assertions, and it needs its own fixture rooted under `.claude/worktrees/`. That is a separate change with its own regression test — a natural fifth guard for `test-guard-worktree-root-discovery.sh`, not a rider on a rebase.

laurigates:

laurigates:
## Rebased onto `main` @ `76f13e1b` — `MERGEABLE` / `CLEAN`, all checks green

Head `7859568f`, 3 commits. Third rebase; resolution was pure union again.

### `.pre-commit-config.yaml` — three hook sets coexist

`git diff origin/main` on that file is **exactly my 6 added lines**, nothing removed. All three sets verified present by id:

| Source | Hooks |
|---|---|
| #2281 | `check-branch-containment-guidance` (L104), `check-branch-containment-guidance-test` (L110) |
| #2290 | `check-workflow-script-triggers` (L158), `test-check-workflow-script-triggers` (L164), `test-guard-worktree-root-discovery` (L170) |
| this PR | `test-run-skill-script-tests` (L339) |

`check-unused-bash-grant` keeps `--strict` (L118).

### Both hotfixed suites stay green on this branch

Exactly the counts `main` carries — no interaction with #2297:

```
scripts/tests/test-check-version-pin-coverage.sh     PASSED=37 FAILED=0 STATUS=OK
scripts/tests/test-guard-worktree-root-discovery.sh  Passed: 21  Failed: 0
```

### The changed zero-scan contract — no reconciliation needed, and here's why

Checked rather than assumed. My three files contain **zero** references to `check-version-pin-coverage`, `FILES_SCANNED`, `TEMPLATE_FILES_SCANNED`, or `plugin_dir` — separate script, own output block, no shared state or parsing.

More usefully: my runner never had the pre-hotfix shape to begin with. Its only early exits are `exit 2` on argument errors, *before* any scanning. The zero-scan path flows straight through to the normal summary — `SCANNED_EMPTY=` is printed, `required_test_missing` is **recorded** into the issue list, and `[ "$issue_count" -eq 0 ]` at the very end decides the exit.

That is the same shape #2297 moved `check-version-pin-coverage.sh` *to*: recorded `ERROR`, every key still emitted, exit decided last. So the runner is structurally incapable of reproducing defect 2 (a guard exiting above a later scan and suppressing its keys) — there is nothing downstream of the exit decision to suppress. The population signal differs (declared manifest entries rather than plugin dirs with a `skills/` tree), but the mechanism is identical: a population signal contradicting a zero denominator, recorded not thrown.

### Verification after rebase

| Check | Result |
|---|---|
| `test-check-version-pin-coverage.sh` | 37/0 |
| `test-guard-worktree-root-discovery.sh` | 21/0 |
| `test-run-skill-script-tests.sh` | 44/0 |
| `run-skill-script-tests.sh` (local) | `TOTAL=86 PASSED=86 SKIPPED=0 SCANNED_EMPTY=false STATUS=OK` |
| `shellcheck`, `actionlint` | clean |
| `check-workflow-script-triggers.sh` | `STATUS=OK`, `ISSUE_COUNT=0` |
| CI checks | gate, lint, validate, **compliance**, split, test — all SUCCESS |

CI log for this head confirms the gate genuinely executed rather than merely going green:

```
Put cargo-generate on PATH        cargo generate 0.23.14
Run skill-local regression tests  PASS=.../test-template-parity.sh
                                  TOTAL=86 PASSED=84 SKIPPED=2
                                  REQUIRED_DECLARED=5 REQUIRED_VIOLATIONS=0
                                  SCANNED_EMPTY=false STATUS=WARN
```

`SKIPPED=2` is still the two taskwarrior tests (`task` CLI absent) — correctly WARN, exit 0, and being addressed in #2294.

Issue #2265, opened 2026-08-02
feedback(code-quality-plugin): code-lint ships unrendered {{ if PROJECT_TYPE }} conditionals


laurigates:
## Skill

`code-quality-plugin/skills/code-lint/SKILL.md` (observed at cached version `1.22.0`)

## Category

Bug — the skill body ships templating syntax that nothing renders, so the agent receives it verbatim.

## Description

Invoking `code-quality-plugin:code-lint` via the Skill tool returns the SKILL.md body with **unrendered template conditionals**. Claude Code does not process `{{ if ... }}` / `{{ endif }}`, so all four language branches arrive at once and the agent has to pick one by hand.

The body also carries shell-parameter-expansion placeholders (`${1:-.}`, `${2:+--fix}`, `${3:+--check}`) that are documentation of intent rather than runnable commands — they only work if something substitutes `$1`/`$2`/`$3`, and nothing does.

Net effect: a skill whose entire job is "run the right linter" hands the agent four mutually-exclusive linter recipes plus placeholder syntax, and leaves both the language selection and the flag substitution as an exercise.

## Evidence

Verified directly in the cached skill source, not just in the tool output:

```
$ grep -n '{{ if\|{{ endif\|PROJECT_TYPE' \
    ~/.claude/plugins/cache/laurigates-claude-plugins/code-quality-plugin/1.22.0/skills/code-lint/SKILL.md
36:{{ if PROJECT_TYPE == "python" }}
42:{{ endif }}
45:{{ if PROJECT_TYPE == "node" }}
50:{{ endif }}
53:{{ if PROJECT_TYPE == "rust" }}
58:{{ endif }}
61:{{ if PROJECT_TYPE == "go" }}
66:{{ endif }}

$ grep -no '\${[12][:-][^}]*}' <same file>
38:${1:-.}
38:${2:+--fix}
39:${1:-.}
...
```

`PROJECT_TYPE` is never assigned anywhere in the skill — there is no step that detects the language and binds it.

The rendered Context block was also empty in this run:

```
- Package files: (Bash completed with no output)
- Pre-commit config: (Bash completed with no output)
```

so even the detection signal the conditionals would presumably branch on came back blank.

## Impact

[...]

PR #2284, opened 2026-08-07
fix(code-quality-plugin): drop unrendered template directives from code-lint

laurigates:
## What

Deletes the eight literal template-directive lines from `code-quality-plugin/skills/code-lint/SKILL.md`:

| Removed | Was at |
|---|---|
| `{{ if PROJECT_TYPE == "python" }}` … `{{ endif }}` | 36, 42 |
| `{{ if PROJECT_TYPE == "node" }}` … `{{ endif }}` | 45, 50 |
| `{{ if PROJECT_TYPE == "rust" }}` … `{{ endif }}` | 53, 58 |
| `{{ if PROJECT_TYPE == "go" }}` … `{{ endif }}` | 61, 66 |

Nothing else in the block changes — the prose and every linter command are byte-identical.

## Why

Nothing renders those lines. Claude Code performs `$ARGUMENTS` / `$N` / `${CLAUDE_*}` substitution only (`.claude/rules/skill-development.md` § String Substitutions), and `PROJECT_TYPE` is never bound by any step in the skill. So a skill whose entire job is "run the right linter" handed the agent raw template syntax as if it were instruction, with all four language branches presented at once.

Deleting them is **behaviour-preserving**: the `### Python` / `### JavaScript/TypeScript` / `### Rust` / `### Go` headings already perform exactly the language selection the conditionals were meant to gate.

## Evidence — verified in source, not measured in sessions

Being explicit about the strength of the signal:

- **0 transcript events.** This is human-reported only, via the fast-loop `session-feedback` issue. There is no quantitative corroboration from session clustering, so nothing here says how often it bit.
- **The defect is independently verified live in the repo** — 4 conditional pairs, confirmed by grep at `origin/main` before the edit and absent after.

Treat it as a verified-in-source defect, not a measured-friction one.

## Scope — the other half of #2265 stays open

`Refs #2265` deliberately, **not** a closing keyword. That issue covers two halves and only one lands here.

The `${1:-.}` / `${2:+--fix}` / `$3` shell-parameter placeholders in the same block are **untouched**. Nothing binds them either, but the repair requires a design decision — does the skill declare `args` and parse them, or detect the path itself and drop the placeholders? — which is out of scope for a behaviour-preserving deletion. **#2265 stays open for that half.**

## Same defect in four sibling skills (not fixed here)

The grep that confirmed this one found the class is wider:

| Skill | Directive lines |
|---|---|
| `tools-plugin:deps-install` | 14 (`PACKAGE_MANAGER`, 7 branches) |
| `project-plugin:project-init` | 2 (`$3 == "--github"`) |
| `typescript-plugin:bun-add` | inline `{{ if DEV }}` / `{{ if EXACT }}` |

Each is a different release-please package, so folding them in would mix scopes in one squash-merge commit. They want their own PRs. (`hooks-plugin:hooks-permission-request-hook` also matches, but that one is a legitimate prose citation of the syntax — leave it.)

## Regression check

Per `.claude/rules/regression-testing.md`, `plugin-compliance-check.sh` `check_skill_body()` gains a guard that ERRORs if `{{ if | endif | else | for | endfor }}` reappears in this skill's body.

**Deliberately skill-scoped, not repo-wide** — an unscoped ban would fail immediately on the four siblings above. The comment in the guard says to widen it (blockquote-exempt, per the existing `task +LATEST _get uuid` precedent) once that sweep lands.

Verified by mutation: re-appending a directive pair to the fixed file produces the ❌; the restored file is clean.

## Verification

| Check | Result |
|---|---|
| `plugin-compliance-check.sh code-quality-plugin` | Body ✅, exit 0 (remaining ⚠️ are pre-existing and unrelated: one long description, three oversized sidecars) |
| Guard mutation test | fires on the defect, silent on the fix |
| `check-context-engineering.py --strict` | exit 0 |
| Full pre-commit suite | passed (one advisory plugin-README nudge — no documented behaviour changed) |

Refs #2265


laurigates:

laurigates:

laurigates:

laurigates:

laurigates:

laurigates:
Closing as **superseded by the #2265 sweep**, which landed on `main` while this was open.

## The fix already landed, by a different route

This PR's payload was deleting the eight `{{ if PROJECT_TYPE == "..." }}` / `{{ endif }}` lines from `code-quality-plugin/skills/code-lint/SKILL.md`. `main` no longer contains them — the sweep rewrote all five affected skills (`code-lint`, `tools-plugin:deps-install`, `project-plugin:project-init`, `typescript-plugin:bun-add`, `testing-plugin:test-analyze`) to a detection-step-plus-lookup-table shape, and also fixed the propagation source in `.claude/rules/agentic-optimization.md` § Environment Detection.

That is a strictly better outcome than this PR's behaviour-preserving deletion: it covers the four siblings this PR explicitly scoped out, and removes the pattern from the rule that was teaching it.

Against current `main` the diff here is `+19 / −0` across two files — no SKILL.md change remains.

## The remaining guard is double-gating

What's left is the `check_skill_body()` branch scoped to `code-quality-plugin/code-lint`. The sweep shipped `scripts/check-unrendered-templates.sh`, which covers **every** skill body repo-wide. Verified by mutation on `main` — re-appending a directive pair to `code-lint`:

```
FILES_SCANNED=570
STATUS=ERROR
ISSUE_COUNT=2
  - SEVERITY=ERROR TYPE=unrendered_template_conditional FILE=code-quality-plugin/skills/code-lint/SKILL.md LINE=163 TOKEN={{ if PROJECT_TYPE == "python" }}
  - SEVERITY=ERROR TYPE=unrendered_template_conditional FILE=code-quality-plugin/skills/code-lint/SKILL.md LINE=165 TOKEN={{ endif }}
```

exit 1, against `FILES_SCANNED=570 STATUS=OK ISSUE_COUNT=0` on the clean tree. The regression this guard exists to catch is already caught, by name and line number. Adding a second, narrower gate is the stacking `.claude/rules/auto-mode.md` warns against — and this PR's own comment anticipated retiring it once the wider sweep landed. It has.

## And the rules row would now be inaccurate

The `.claude/rules/regression-testing.md` row describes the remedy as *"Deleted the eight directive lines"* with a `code-lint`-scoped guard. That is not what `main` did. `main` already carries a row for the sweep covering the same issue #2265, so merging this would leave two rows for one defect, the newer one describing a fix that was never applied.

## Scope note

#2265 stays **open** either way — the `${1:-.}` / `${2:+--fix}` / `$3` shell-parameter half was out of scope here and is unaffected by this closure.

Nothing is lost by closing: the defect is fixed, the guard is stronger and repo-wide, and the history is recorded once, accurately.

---

Issue #2214, opened 2026-07-29
fix(scripts): check-version-pin-coverage.sh lints the gitignored dist/ build output


laurigates:
## What

`scripts/check-version-pin-coverage.sh` walks the **gitignored `dist/`** rulesync build output and can hard-ERROR on a stale local artifact, blocking every local commit that touches a skill file. CI never sees it, because `dist/` is never committed.

## Evidence

Hit while committing an unrelated change (PR #2213, which touches `CLAUDE.md` and `.claude/skills/`):

```
skill version pins are a Renovate-managed shape (version-pinning.md)....Failed
- hook id: check-version-pin-coverage
- exit code: 1
STATUS=ERROR
ISSUE_COUNT=29
  - SEVERITY=ERROR TYPE=version_input_missing
    MSG=dist/opencode/skills/configure-surface/SKILL.md: 'Connorrmcd6/surface' installs its
        own binary but the step sets no 'version:' input …
```

`dist/opencode/skills/configure-surface/SKILL.md` is a **stale export predating #2175** — it still carries `Connorrmcd6/surface@004c9169… # v0.6.2` with no `version:` input. The real source, `configure-plugin/skills/configure-surface/SKILL.md:141`, already carries `version: v0.8.0` and is correct. Of the 29 reported issues, **16 are `dist/` duplicates** of findings already counted against their real sources.

`dist/` is gitignored at `.gitignore:47`.

## Root cause

`scripts/check-version-pin-coverage.sh:178` prunes `.claude/worktrees/` but not `dist/`:

```sh
done < <(find "$proj_dir" -path '*/.claude/worktrees/*' -prune -o \
```

Its sibling already gets this right — `scripts/lint-context-commands.sh:63` passes `--exclude-dir='dist'` alongside `worktrees` and `node_modules`, and `.claude/rules/regression-testing.md` records that exclusion as deliberate ("file discovery also excludes the gitignored `dist/` rulesync build output").

The `.claude/worktrees/` prune arrived under #1492 for exactly this class of defect — a walk descending into copies of the repo. `dist/` is the same class (a *generated* copy rather than a *cloned* one) and was missed.

## Why it is worth fixing rather than regenerating

- **Invisible to CI.** `dist/` is never committed, so no PR can catch it and no green build indicates health. It only bites a developer's local `git commit`, which is the same local-only-invisible shape as #2141.
[...]

Issue #1492, opened 2026-06-03
perf(scripts): check-version-pin-coverage.sh scans .claude/worktrees copies, ~4-5min/run


laurigates:
## What

`scripts/check-version-pin-coverage.sh` (added in the `ci(version-pinning)` commit on #1476, wired into `.pre-commit-config.yaml` and `.github/workflows/plugin-pr-checks.yml`) takes **~4–5 minutes per run**. It makes every local `git commit` and the `plugin-pr-checks` CI step crawl. It exits 0 / passes — this is purely a performance problem.

## Why

Two compounding causes:

1. **It scans agent worktree copies.** The file walk —
   ```sh
   find "$proj_dir" -path '*/skills/*' -name '*.md' -type f -print0
   ```
   descends into `.claude/worktrees/agent-*/` and `.claude/worktrees/<name>/`, which are **full copies of the repo** created by other concurrently-running agents. Observed in a single run: it scanned `*/skills/*.md` under `.claude/worktrees/agent-a0a0f9917b0ddc60e/`, `.claude/worktrees/competent-faraday-d463b7/`, **and** the real tree — every skill file processed 3×+ (and growing with each active worktree). The WARN output is littered with `.claude/worktrees/...` paths, confirming the duplication.

2. **A subshell fork per code-fence line.** The fence-toggle on line ~94 —
   ```sh
   in_fence=$([ "$in_fence" = true ] && echo false || echo true)
   ```
   forks a subshell + two `echo`s for every ` ``` `/`~~~` marker, across the (already inflated) set of files.

## How

1. **Prune the worktree copies from the walk** (the main win):
   ```sh
   find "$proj_dir" -path '*/.claude/worktrees/*' -prune -o \
        -path '*/skills/*' -name '*.md' -type f -print0
   ```
   (or `-not -path '*/.claude/worktrees/*'`). The guard should only ever audit the real tree, never sibling agents' checkouts.

2. **Replace the per-line fork** with a pure-bash toggle:
   ```sh
   if [ "$in_fence" = true ]; then in_fence=false; else in_fence=true; fi
   ```

3. Add a regression assertion to `scripts/tests/test-check-version-pin-coverage.sh` that a `.claude/worktrees/<x>/.../skills/*.md` fixture is **not** counted (`files_scanned` unchanged when one is added), per `.claude/rules/regression-testing.md`.

After the prune, a full run should drop from minutes to a few seconds.

## Context

[...]

PR #2289, opened 2026-08-07
fix(code-quality-plugin): replace unrendered template conditionals with a detection step and lookup table

### LINKED ISSUE #2265: feedback(code-quality-plugin): code-lint ships unrendered {{ if PROJECT_TYPE }} conditionals

laurigates:
## What

`code-quality-plugin/skills/code-lint/SKILL.md` shipped four `{{ if PROJECT_TYPE == "python" }}` / `{{ endif }}` blocks. **Claude Code has no template engine in the skill-invocation path**, so all four language branches reached the agent verbatim, `PROJECT_TYPE` was never bound anywhere in the skill, and the `${1:-.}` / `${2:+--fix}` / `${3:+--check}` shell-parameter placeholders had nothing substituting them. A skill whose entire job is "run the right linter" handed the agent four mutually-exclusive recipes and left both the language selection and the flag substitution as an exercise.

The failure is silent by construction: the skill loads, every structural lint passes, and the agent quietly picks a branch (or pastes `{{ endif }}` into a shell).

## Sweep — the same defect in 5 skills, not 1

I swept the whole corpus for `{{ if` / `{{ endif` / `{{ end }}` / `{{#if` / `{{/if` / `PROJECT_TYPE` across `*/skills/**/*.md`. **Four more skills carried the identical defect** and are fixed in this PR:

| Skill | Markers |
|---|---|
| `code-quality-plugin:code-lint` | 4 × `{{ if PROJECT_TYPE == … }}` / `{{ endif }}` (the reported one) |
| `tools-plugin:deps-install` | 7 × `{{ if PACKAGE_MANAGER == … }}` / `{{ endif }}` |
| `project-plugin:project-init` | `{{ if $3 == "--github" }}` / `{{ endif }}` |
| `typescript-plugin:bun-add` | `{{ if DEV }}` / `{{ if EXACT }}` inline inside its one command |
| `testing-plugin:test-analyze` | Handlebars `{{#if ARG2}}` / `{{else}}` / `{{/if}}` (× 3 blocks) |

`.claude/rules/agentic-optimization.md` § Environment Detection was **teaching** the pattern (`{{ if CI }}--frozen-lockfile{{ endif }}`) — that is where it propagated from, so it is fixed too (`scaffold-fix-backport`: fixing the instances without the generator leaves the next author to repeat it).

**Not defects, deliberately left alone** — verified individually, not assumed:

- `kubernetes-plugin` helm skills (`helm-chart-development` ×2, `helm-debugging`, `helm-values-management`) — Go chart templates that **Helm genuinely renders** at deploy time.
- `hooks-plugin:hooks-permission-request-hook` / `hooks-session-start-hook` `REFERENCE.md` — hook-script **generator templates** whose own text instructs the agent: *"Remove all `{{ if ... }}` / `{{ endif }}` markers after substitution."* A declared renderer exists (the agent), so these are correct.

## Fix shape

All five rewritten to the same shape — an explicit **detection step the agent performs**, then a **lookup table it reads**, with every command written against a named placeholder that `## Parameters` states the *caller* supplies:

- **`code-lint`** now leads with the bundled `detect-and-fix.sh` one-shot path (issue's suggestion 3), then falls back to marker-file detection (`pyproject.toml` → Python, `package.json` → JS/TS, `Cargo.toml` → Rust, `go.mod` → Go) plus a per-language command table with explicit `--fix` / `--format` columns. Polyglot repos run every matching row rather than picking one arbitrarily.
- **`deps-install`** detects on **lock files first** (they name the manager that actually produced the install), manifest second.

`code-lint` and `deps-install` also move to a bare `Bash` grant: their narrow `allowed-tools` patterns did not cover `uv`, `npx`, `cargo`, `go`, `npm`, or `pre-commit`, so most of their own commands would have prompted. This is the ratified house standard when the payload is shell (`.claude/rules/agentic-permissions.md`), and it is checked, not assumed — `check-unused-bash-grant.sh --strict` (made blocking by #2290) reports `SKILLS_SCANNED=408 BASH_GRANTEES=188 ISSUE_COUNT=0 STATUS=OK` with an empty allowlist. Both skills genuinely run shell, so the grant is justified rather than narrowed.

## Regression guard (mandatory per `.claude/rules/regression-testing.md`)

`scripts/check-unrendered-templates.sh` — fails on any Mustache/Jinja/Handlebars conditional in `*-plugin/skills/**/*.md`. Emits the `=== SECTION ===` / `KEY=VALUE` / `STATUS=` / `ISSUE_COUNT=` contract; exits 2 on an unknown argument (#2057); prunes `.claude/worktrees/` and `dist/` (#1492 / #2214).

Two exemption mechanisms, both deliberate:

1. **Structural** — Go/Helm chart templates (`{{-` / `-}}` trim markers, bare `{{ end }}`, a condition rooted at `.Values` or `$.`) and the `{{ if ... }}` ellipsis prose form.
2. **Declared** — a file containing the phrase `markers after substitution` is a generator template the skill renders itself. Keeping the exemption *in the file* rather than in a path allowlist makes it self-documenting and stops the allowlist drifting from the thing it describes.

It also complies with #2290's zero-scan contract: discovery runs from **inside** the scan root on relative paths (so a worktree-shaped root is not self-pruned), and plugin-dirs-present-but-zero-skills reports `STATUS=ERROR TYPE=nothing_scanned` while a genuinely empty tree stays `STATUS=OK` + `SCANNED_EMPTY=true`.

`scripts/tests/test-check-unrendered-templates.sh` — 58 assertions, semantic (it **executes** the guard, never greps it). It replays the **verbatim** pre-fix bodies of all five skills, and weights the false-positive half equally, since a guard that flags legitimate Helm charts gets disabled.

**Mutation-verified: eleven neutered rules, all killed.** The mutation pass found three defects that would otherwise have shipped:

1. The Go `$`-rooted skip was written as a bare `$`, which **silently swallowed** `project-init`'s shell positional `{{ if $3 == "--github" }}` — the discriminator is `$.` (Go's root context), not `$`.
2. The Go dot rule was `.*`, which also matched the `...` ellipsis and made the ellipsis rule dead code — narrowed to `.[A-Za-z]*`, so that test now kills its own mutation.
3. Reverting discovery to an absolute base makes the guard scan **nothing** from this worktree — the exact #2219/#2290 defect, now pinned by its own test.

## Verification

| Check | Result |
|---|---|
| `scripts/check-unrendered-templates.sh` (real repo) | `FILES_SCANNED=564 FILES_EXEMPT=2 UNRENDERED_CONDITIONALS=0 STATUS=OK` |
| Same guard vs. the 5 **pre-fix** bodies | `STATUS=ERROR`, 34 markers reported by name, exit 1 |
| `scripts/tests/test-check-unrendered-templates.sh` | 58 passed, 0 failed |
| `scripts/run-skill-script-tests.sh` | `TOTAL=86 FAILED=0 STATUS=OK` |
| `scripts/check-unused-bash-grant.sh --strict` | `STATUS=OK ISSUE_COUNT=0`, allowlist empty |
| `scripts/lint-shell-scripts.sh` | 0 errors (9 pre-existing warnings, unrelated) |
| `shellcheck` on both new scripts | clean |
| `scripts/plugin-compliance-check.sh` | exit 0, no findings for any edited skill |
| `scripts/lint-context-commands.sh` | All context commands OK |
| Skill sizes | all 5 under 9,000 chars (OK band; ceiling 26,000) |

### Pre-existing `pre-commit run --all-files` failures, inherited from `main`

Three hooks fail on this branch **and on `main`**; none is in this PR's diff (`git diff --stat origin/main HEAD` over those paths is empty), and my own two hooks pass:

| Hook | Owner | Why it is not this PR |
|---|---|---|
| `shellcheck` | 17 files across 7 plugins | Pre-existing SC2015/SC2034 findings surfaced by `--all-files`; the pre-commit hook normally only lints *staged* files (the meta-gotcha in `.claude/rules/shell-scripting.md`). Neither new script appears in its output. |
| `test-guard-worktree-root-discovery` | #2290 | Asserts `check-version-pin-coverage` reports `FILES_SCANNED=0` for a synthetic root; fails against a script #2287 changed afterwards. |
| `test-check-version-pin-coverage` | #2287 | TEST I's `TEMPLATE_FILES_SCANNED` comes back empty for its `mktemp -d` fixture. |

Both regressions point at one root cause: **#2287 (`b81a4bfa`) merged after #2290 (`5c31e252`)** and its template-scanning path did not carry #2290's zero-scan/relative-path contract. On the real repo the script is fine (`FILES_SCANNED=593 TEMPLATE_FILES_SCANNED=5 STATUS=WARN`) — only the synthetic fixtures fail, so it needs a follow-up on those two files rather than a fix here.

## Relationship to the concurrent draft #2284

#2284 ("drop unrendered template directives from code-lint") was opened ~an hour before this PR by another session against the same issue. **I have not touched it** — closing it is the maintainer's call. Its content is a strict subset of this PR with **one exception, which I have adopted**:

| #2284 has | Status here |
|---|---|
| Deletes the 8 directive lines from `code-lint` only | Superseded — this PR rewrites all five affected skills |
| Explicitly leaves the `${1:-.}` / `${2:+--fix}` half unfixed, keeping #2265 open | Both halves fixed; #2265 closes |
| Guard scoped to `code-quality-plugin/code-lint` inside `plugin-compliance-check.sh`, with a note to widen "once that sweep lands" | This PR *is* that sweep; the guard is repo-wide and standalone (`plugin-compliance-check.sh`'s CI run is wrapped in `\|\| true`, so a dedicated pre-commit hook is the stronger placement) |
| Its regex covers `for` / `endfor` | **Adopted** — an earlier draft of my keyword set missed loop forms entirely |

That last row was a real gap, credited in the regression row. Widening it also forced a decision the narrower guard did not have to make: Go's `range` stays **out** of the keyword set, because it is the one loop keyword with legitimate rendered uses here (four Helm chart examples). TEST N makes that exclusion falsifiable with a **bare** `{{ range $k, $v := … }}` — the trim-marker forms are skipped structurally whatever the keyword set is, so a naive fixture would let `range` back in unnoticed. Mutation-verified both ways: dropping `for`/`endfor`/`each` fails 5 assertions, adding `range` fails 2.

## CI wiring

Wired into `.pre-commit-config.yaml` only. **Correction to the original PR body**: it said the CI step was deferred to #2219 — #2219 has since been closed by #2290, and #2258 removed `plugin-pr-checks.yml`'s `paths:` filter entirely, so the gap that motivated the deferral is gone. A CI step for this guard is a clean follow-up; it is left out here to avoid re-conflicting with a workflow file three merged PRs have touched today.

Closes #2265


laurigates:

PR #2294, opened 2026-08-07
fix(ci): build Taskwarrior 3.x on the runner so the UUID-capture gate runs

laurigates:
Refs #2221. **No longer stacked** — #2282 has merged, and `main` is merged down, so the diff here is now just this change: 5 files, +79/−6.

## What was unrun

#2282's first CI run reported two more tests as `SKIP: task CLI not available`:

- `taskwarrior-plugin/skills/task-add/scripts/tests/test-uuid-capture.sh`
- `taskwarrior-plugin/scripts/tests/test-task-id-stability.sh`

The first one **is** the executable gate added for #1417/#1819, after a *syntactic* guard — a grep for the literal `task +LATEST _get uuid` — let a non-working command ship. `_get` is a DOM accessor; handed a tag filter it returns empty, so the skill captured no UUID and silently reverted to the numeric-ID drift #1417 was meant to fix. **An executable test that never executes is that same syntactic gate again, one level up.**

Its sibling skipped the same way, and its skip is broader than it looks: Parts 2 and 3 (the partial-conversion grep and the citation-drift guard) need no `task` binary at all, yet the early `command -v task` SKIP gates them too.

Both were correctly WARN, not ERROR — neither was declared in `scripts/required-to-run-tests.txt`. This closes that.

## Decision: (a) install it — and from source

Option (b) (document them as local-only) was rejected: the tests were written as executable gates *specifically because* the syntactic version was not enough.

**Not `apt-get install taskwarrior`.** ubuntu-24.04 universe pins **2.6.2**, and these skills document Taskwarrior 3.x. This is a stronger objection than speed: a 2.x gate could read green on guidance broken for every real user — the same "passes without testing the real thing" failure the whole change is about. And it is not hypothetical — I installed 2.6.2 and **both suites pass on it**, so the false-green was available and would have looked convincing.

**Not a prebuilt binary** — there isn't one. Checked the release rather than assuming: `v3.4.2` carries exactly **one** uploaded asset, `task-3.4.2.tar.gz` (source), plus GitHub's two auto-generated source archives. So the issue's preferred route does not exist upstream, and `cargo install` is not applicable either (Taskwarrior is C++ with a Rust `taskchampion` component, not a crate).

**So: build from source, pinned + cached**, following the #2282 cargo-generate step as the model — version in the top-level `env:` block with a `# renovate:` annotation (picked up by the workflow-env customManager #2282 added, no `renovate.json` change needed) and `actions/cache` keyed on version + os + arch.

| | |
|---|---|
| Cold build | **measured in CI: `test` job 5m03s** (C++ + the Rust taskchampion crate, 4 vCPU) |
| Warm restore | **measured: 60–73s, `Cache hit occurred on the primary key taskwarrior-3.4.2-linux-x86_64`.** One binary, 56 MB unstripped / 17 MB as a cache entry (`strip` only reaches 39 MB, so no strip step) |
| Cache seeding | `scripts/required-to-run-tests.txt` is already in the workflow's `push:` path list, so merging this seeds the `main` cache that PR runs fall back to |
| Runtime deps | `libuuid1`, `libstdc++6`, `libgcc`, `libm`, `libc` — all stock (`ldd`-checked) |
| Build deps | cmake, g++ and Rust are on the runner image; `uuid-dev` is installed explicitly rather than gambled on — it's the one header the configure step probes for |
| Integrity | the version tag, same as cargo-generate — upstream publishes no checksum file |

`git clone --depth 1 --branch v$VERSION` rather than the tarball, because that is the exact fetch I verified end-to-end and it doesn't assume the tarball's internal directory name.

## Verification

Built 3.4.2 locally and ran the real corpus **in both directions** — not a fixture.

| | Result |
|---|---|
| `test-uuid-capture.sh` on 3.4.2 | 4 passed, 0 failed |
| `test-task-id-stability.sh` on 3.4.2 | 25 passed, 0 failed |
| `run-skill-script-tests.sh`, task on PATH | both reported `PASS=` |
| `run-skill-script-tests.sh`, binary moved aside | both flip to `SKIP=`, raise `SEVERITY=ERROR TYPE=required_test_skipped … MSG=task CLI not available`, `STATUS=ERROR`, **exit 1** — where before this change they read `PASS=` / `STATUS=OK` |
| `scripts/tests/test-run-skill-script-tests.sh` | 44/44 (its guard-integrity assertion already requires every manifest entry to exist on disk) |
| `actionlint`, `check-workflow-model.sh`, `check-workflow-script-triggers.sh`, `test-guard-worktree-root-discovery.sh`, `lint-shell-scripts.sh`, `pre-commit` | clean |

**Confirmed in CI**, from the `test` job log rather than the check-run conclusion:

```
PASS=./taskwarrior-plugin/scripts/tests/test-task-id-stability.sh
PASS=./taskwarrior-plugin/skills/task-add/scripts/tests/test-uuid-capture.sh
TOTAL=85  PASSED=85  SKIPPED=0  FAILED=0
REQUIRED_DECLARED=7  REQUIRED_VIOLATIONS=0  STATUS=OK
```

`REQUIRED_DECLARED=7` (5 + these 2) with `SKIPPED=0` is the load-bearing pair. Note the structural consequence: because the runner exits 1 on a declared test that skips, **a green `test` job is now itself proof these two executed** — which is the property this change exists to create.

**Green on arrival**, so there is no red-gate-gets-disabled hazard (`generated-fleet-drift.md`).

> Note on the expected local result: the issue text anticipated `STATUS=WARN` for these. Once they are in the required manifest a local run *without* taskwarrior is `STATUS=ERROR` / exit 1, which is the point of listing them.

## No new regression check

This is a **wiring** defect, not a detection one, and the detection already exists twice over: the runner's `required_test_skipped` enforcement (pinned by `test-run-skill-script-tests.sh`) and the two tests themselves. Adding a third guard would be the double-gating `auto-mode.md` warns about.

## Files changed

| File | Change |
|---|---|
| `.github/workflows/test-skill-scripts.yml` | `TASKWARRIOR_VERSION` pin + `# renovate:` annotation, cache + build steps, PATH step renamed to cover both tools |
| `scripts/required-to-run-tests.txt` | both paths declared; header's "taskwarrior … absent by design" note corrected |
| `taskwarrior-plugin/…/test-uuid-capture.sh`, `…/test-task-id-stability.sh` | header notes — the SKIP is fine locally, an error in CI (same shape #2282 gave `test-template-parity.sh`) |

No `taskwarrior-plugin/README.md` update (the currency nudge fires): the plugin change is two comment headers, with no user-facing behavior change.

## Merge history

Three merges from `main` landed here, each verified by hand rather than trusted:

1. **Auto-resolved union merge** (#2278/#2279/#2283/#2287/#2290). Checked because a union merge on a line this PR *modified* is exactly the case it gets wrong; it was correct.
2. **Deliberate merge for #2297.** This branch carried #2290's unanchored `grep -q 'FILES_SCANNED=0'` **without** #2287's `TEMPLATE_FILES_SCANNED` key that triggers it — the two-commit gap was the only thing keeping CI green. Taking one without the other would have gone red, so both came in together.
3. **#2282's squash-merge** (this one). Conflicts resolved by hand: took `main`'s newer `run-skill-script-tests.sh` and its self-test wholesale (this branch never edited either); kept `main`'s newer manifest header including the #2255/#2290 denominator-guard paragraph, with this branch's one correction on top (taskwarrior dropped from the "absent by design" list, because it no longer is). **`merge=union` duplicated #2282's regression row verbatim** — removed, and this PR's row was then re-anchored to `main`'s ordering so the diff is a pure `+1` and the next merge cannot duplicate it again.


laurigates:

PR #2298, opened 2026-08-07
fix(scripts): silence relative-invocation stderr and anchor the positive assertions

laurigates:
> **Rescoped.** #2297 landed the same two primary fixes on `main` while this was open, with a **better** approach for the misfire. This PR is now rebased onto `main` and reduced to the two residuals #2297 did not cover. See "What was dropped" below.

## 1. Relative-invocation stderr noise

`check-version-pin-coverage.sh` resolves `script_dir` from `$BASH_SOURCE` *after* the `cd "$proj_dir"` that #2290 added, so a **relative** invocation — `bash scripts/check-version-pin-coverage.sh …`, the form pre-commit and CI use — looks for `scripts` inside the scan root:

```
$ bash scripts/check-version-pin-coverage.sh --project-dir <fixture>
scripts/check-version-pin-coverage.sh: line 244: cd: scripts: No such file or directory
```

Still reproduces on `main` today. Harmless to the verdict — the failed `cd` is confined to a subshell, and `$script_dir` is only read on the `files_scanned > 0` path where the helper resolves anyway — but it is noise on every run and a real break for any caller gating on empty stderr. Hoisted above the `cd`, with the reason recorded at both the definition and the old site.

An **absolute** invocation cannot reproduce it, which is why it survived every test in the suite. The new assertion therefore runs from `$repo_root` via a **relative** path *on purpose*; written the obvious way it is a no-op that passes against the defect.

## 2. The positive assertions were left unanchored

#2297 anchored the two *negative* scan-count assertions. The trap is symmetric, and the ten *positive* ones were left on unanchored `contains`:

> an unanchored `SCANNED=1` is satisfied by any `*_SCANNED=1` **sibling key**, so a guard that discovered nothing in its own corpus can pass on a neighbour's counter.

These guards keep gaining keys (`TEMPLATE_FILES_SCANNED`, `EXEMPTED_CALLS`, …), so the collision surface only grows. Adds `lacks_line()` beside #2297's `contains_line()` and converts **every** whole-line `KEY=VALUE` assertion — both polarities, all four guards.

`contains()` is deliberately retained where the match is genuinely a substring: prose (`No agent files found`) and tokens inside indented `ISSUES:` rows (`nothing_scanned`, `non_opus_model`).

## What was dropped, and why

Two things this PR originally carried are now on `main` via #2297 and have been **removed** rather than re-litigated:

| Originally here | Resolution |
|---|---|
| Anchoring the two negative assertions | Same fix landed in #2297. Kept theirs, incl. the here-string `contains_line` (avoids the `pipefail` SIGPIPE trap) — this PR builds on it. |
| `template_file_count` in the `nothing_scanned` predicate | **Dropped in favour of #2297's approach**, which is better: it records the misfire via `add_issue` and *falls through* (so the report stays complete) and uses "plugin dirs that have a `skills/` tree" as the population signal. Mine early-exited and used a template count. |

My TEST J went with it — it asserted my implementation's exit contract, and #2297's TEST I already covers the behaviour.

### A note on the auto-resolver

`auto-resolve-conflicts` merged `main` into this branch three times. Recording all three, because the pattern is the useful part:

| Merge | Outcome |
|---|---|
| 1 and 2 (against the **pre-rescope** branch) | **Not safe to keep.** Reverted #2297's `skills/`-tree discriminator back to a plain `*-plugin` count, and introduced a `nothing_scanned` boolean plus a terminal `if [ "$nothing_scanned" = true ]; then exit 1; fi` present **on neither side** — the signature of a union merge concatenating both halves of a conflicted hunk. Both suites passed on that merge, so tests would not have caught it. Discarded and rebased instead; that is what the force-push was. |
| 3 (against the **rescoped** branch) | **Clean, and kept.** Verified before accepting: both script files byte-identical to my rebased intent, #2297's `skills/`-tree discriminator intact, and the rules table at 207 rows = main's 206 + mine, no duplication. |

The distinction is the point: merge 3 was safe because dropping my competing `template_file_count` change removed the conflicting hunk, so there was nothing for union semantics to concatenate. Union is right for the append-only `regression-testing.md` table (which is what `.gitattributes` configures) and wrong for shell control flow — so the hazard is live for the next PR that conflicts inside a `.sh`, not something this PR fixes.

## Verification

Head `5eaa91d`, `mergeable_state: clean`. Diff is 3 files, +40/−13.

| Check | Result |
|---|---|
| `test-guard-worktree-root-discovery.sh` | **22 / 0** (was 21 on main) |
| `test-check-version-pin-coverage.sh` | **37 / 0** (unchanged from main) |
| Real-repo run | `FILES_SCANNED=593 PLUGIN_DIRS=45 TEMPLATE_FILES_SCANNED=5 STATUS=WARN ISSUE_COUNT=14` — main's documented baseline exactly, now with **0 bytes on stderr** |

Mutation-verified in both directions, so neither change makes the suite vacuous:

- Moving `script_dir` back below the `cd` fails exactly the new stderr assertion (21/1).
- Reinstating the pre-#2219 absolute-base prune (`find "$PWD"`) still fails the **anchored positive** assertion (20/2) — tightening a matcher risks making it unfalsifiable, and this shows it did not.


laurigates:

laurigates:

laurigates:

PR #2299, opened 2026-08-07
fix(ci): discover tests relative to the root so a worktree root is not pruned

laurigates:
Refs #2219, Refs #2221.

## ⚠️ Stacked on #2282 — read this first

**#2282 is open, not merged.** Its head is where `norm_path()`, `scripts/required-to-run-tests.txt`, `SCANNED_EMPTY=`, and `scripts/tests/test-run-skill-script-tests.sh` actually live — none of them are on `main`. This PR is branched off `fix/template-parity-ci-coverage-2221`, so **until #2282 merges the diff below shows its commits too**; afterwards it collapses to the single commit `d98dd1c`.

Base is deliberately `main`, not #2282's branch: a PR based on a branch that gets deleted on merge is auto-closed and **cannot be reopened** once its base ref is gone — the #2049 failure this repo already has a scheduled sweep for. Merge #2282 first, then this.

## The defect

`scripts/run-skill-script-tests.sh` carried the **identical** absolute-base prune defect #2290 fixed in four sibling guards, and #2290's own class test did not cover it:

```bash
find "$root_dir" \
  -path '*/.claude/worktrees/*' -prune -o \
  \( … \) -type f -print0
```

When `$root_dir` is itself an agent worktree, the root's own absolute path contains `/.claude/worktrees/`, so the bare prune matches the scan root, the whole walk is pruned, and discovery returns **zero** tests. Reproduced before touching anything:

```
$ bash scripts/run-skill-script-tests.sh --root /…/repo/.claude/worktrees/agent-f00d
TOTAL=0
SCANNED_EMPTY=true
STATUS=WARN        # exit 0
```

It fails **worse** than the four `check-*.sh` siblings. A collapse in a guard means that guard checked nothing; a collapse *here* means every skill-local and hook regression suite in the tree silently does not run — the substrate meant to catch #2290's class is itself blind.

**Latent, not active.** The `justfile` recipe and `test-skill-scripts.yml` both pass the default relative root `.`, and a relative root cannot match the prune. Only an absolute-root invocation from inside a worktree breaks — which is an agent verifying its own change, this repo's normal way of doing plugin work. #2282's non-empty `required-to-run-tests.txt` already makes a collapse loud (every declared entry raises `required_test_missing`), so it is no longer *silent*; the discovery bug was still there.

## The fix

Discovery runs from **inside** the root against **relative** paths, per `check-agent-model.sh` / `check-subagent-types.sh`:

```bash
done < <(cd "$root_abs" && find . \
  -path '*/.claude/worktrees/*' -prune -o \
  …
```

The root becomes `.`, so its absolute prefix cannot match, while worktree copies nested anywhere below it still prune correctly.

**Anchoring the prune to `"$root/.claude/worktrees/*"` was rejected** for the same reason as in #2290: it silently stops pruning a worktree nested inside a plugin directory. The relative-path form keeps both properties.

### Ripple

| Concern | Outcome |
|---|---|
| `norm_path()` | Simplified to a `./` strip — every path now arrives as `./<rel>` regardless of how `--root` was spelled, so manifest matching is *more* uniform, not less |
| Emitted `PASS=`/`SKIP=`/`FAIL=` form | **Default `--root .` output is byte-identical before and after** (`find .` already emitted `./…`); `diff` of the full 86-test run confirms, as does the CI log |
| Test invocation | `bash "$root_abs/$rel"`, so the caller's cwd is unchanged (the `cd` lives in the process-substitution subshell) |
| Unresolvable `--root` | Now exits 2 instead of reporting `SCANNED_EMPTY=true` / `WARN` / exit 0. A typo'd root reading as "found nothing" is the exact failure this class is about |

## Regression coverage

**`scripts/tests/test-run-skill-script-tests.sh`** — 40 → 56 assertions.

**CASE 9** plants the fixture tree UNDER a `.claude/worktrees/agent-…` path — the assertion that actually pins this bug. Guard integrity throughout, because `TOTAL=2` alone would hold for a runner that listed files and executed none:

- `PASSED=1` and the per-file `SKIP=` classification must hold → the tests really ran
- required-manifest matching must still resolve from that root (pre-fix it raised `required_test_missing` for every entry, since nothing was discovered)
- a worktree clone **nested below** the worktree-shaped root must **still be pruned** → proves the fix did not simply disable pruning

**CASE 10** pins the unresolvable-root exit 2.

**Mutation-verified twice.** Against the pre-fix runner, and — to attribute the new assertions to the discovery change *alone* — against a copy with only the `find` line reverted:

```
FAIL: case 9: a worktree-shaped root still discovers its tests
FAIL: case 9: discovery did not collapse to an empty corpus
FAIL: case 9: the discovered test actually ran (guard integrity)
FAIL: case 9: classification still works from a worktree-shaped root
FAIL: case 9: required-test matching survives a worktree-shaped root
FAIL: case 9: no phantom manifest drift
FAIL: case 9: a nested worktree clone is still pruned
```

All 7 fail; every other assertion still passes.

**`scripts/tests/test-guard-worktree-root-discovery.sh`** — the runner is added as the **fifth guard**, since that file is the repo's home for this bug class (5 new assertions). 4 of the 5 fail against a reverted `find` line; the 5th (`exits 0`) passes trivially, which is precisely why the other four exist. The hook's pre-commit `files:` regex is extended to `run-skill-script-tests\.sh` so editing the runner triggers it — the allowlist-drift lesson from #2164.

## Verification

**All CI green** on `c784bf8`: `test`, `lint`, `validate`, `compliance`, `split`, `gate`, `conventional-commits` ✅ (`doc-audit` skipped).

| Check | Result |
|---|---|
| `scripts/tests/test-run-skill-script-tests.sh` (CI) | `PASS_COUNT=56 FAIL_COUNT=0 STATUS=OK` |
| `scripts/run-skill-script-tests.sh` (CI) | `TOTAL=86 PASSED=84 SKIPPED=2 FAILED=0 REQUIRED_DECLARED=5 REQUIRED_VIOLATIONS=0 SCANNED_EMPTY=false STATUS=WARN ISSUE_COUNT=0`, exit 0 |
| Emitted path form (CI) | `PASS=./blueprint-plugin/hooks/…` — unchanged from pre-fix |
| `scripts/run-skill-script-tests.sh` (local) | **byte-identical** to the pre-fix run (`diff` clean) |
| `scripts/tests/test-guard-worktree-root-discovery.sh` | 26 passed, 0 failed |
| `bash scripts/lint-shell-scripts.sh` | 0 errors (9 pre-existing `macos-plugin` WARNs, untouched) |
| `shellcheck` on all three changed shell files | clean |

`STATUS=WARN` rather than `OK` on the runner is **correct and green**: two optional taskwarrior tests skip because the runner has no `task` CLI, and neither is in `required-to-run-tests.txt`. Per #2282's contract a skip is never silently `OK`, so `WARN` with `ISSUE_COUNT=0` and exit 0 is the expected clean result. (An earlier revision of this body predicted `STATUS=OK`; corrected against the actual log.) The CI run also confirms `test-template-parity.sh` and both `test-rules-project.sh` gates now genuinely **execute** rather than skip, which is #2282's whole point.

### Notes on two earlier red items, both now resolved

1. **`check-version-pin-coverage: does not report FILES_SCANNED=0`** in the class test — a gap left by #2290 that was red on `main`, and red here on the first push. **Fixed on `main` and picked up by the auto-merge**: #2290's assertion used an unanchored substring test, so `FILES_SCANNED=0` matched inside `TEMPLATE_FILES_SCANNED=0` and fired on correct behaviour. Main added a whole-line `contains_line()` helper. Not my change, but it is why `compliance` is now green — the class test went 25/1 → 26/0.

2. **`foundryvtt-module-scaffold` size ⚠️** in the compliance bot comment is a WARN at 11466 chars against a 26000 ERROR ceiling, on a file this commit does not touch (it belongs to #2282). No action.

The auto-resolve merge (`c784bf8`) was verified by hand: my five files are intact, `regression-testing.md` has no duplicated or reordered rows (the `merge=union` driver behaved), and there are no real conflict markers anywhere in the tree — the `<<<<<<< HEAD` strings in `hooks-plugin/hooks/test-task-completeness.sh` are deliberate heredoc fixtures for the conflict-marker detection test.

## Files changed

| File | Change |
|---|---|
| `scripts/run-skill-script-tests.sh` | relative-path discovery, `root_abs` resolution + fail-fast, simplified `norm_path()` |
| `scripts/tests/test-run-skill-script-tests.sh` | assertions repointed to the `./<rel>` form; CASE 9 (worktree-shaped root) + CASE 10 (unresolvable root) |
| `scripts/tests/test-guard-worktree-root-discovery.sh` | the runner added as the fifth guard |
| `.pre-commit-config.yaml` | class-test `files:` regex extended to the runner |


laurigates:

Issue #1616, opened 2026-06-13
feat(blueprint-plugin): widen structural-cue detection beyond manifests + export lines


laurigates:
## What

Extend `blueprint-plugin/hooks/blueprint-structural-cue.sh` detection signals beyond plugin/marketplace manifests and `export` / `pub` lines. Follow-up to #1611; tracked under [ADR-0017](../blob/main/docs/adrs/0017-hook-based-behavioral-cues-for-plugin-utilization.md).

Part of #1599.

## Why

The MVP started deliberately narrow to avoid false positives. The ADR roadmap calls for widening the detection surface once the cue has proven useful in practice.

## How

- Add signals: new public symbols in more languages, route / handler / schema registration, public type changes.
- Preserve the once-per-session dedup and the ADR token-budget contract.
- Extend `hooks/spec/blueprint_structural_cue_spec.sh` with a case per new signal (`.claude/rules/regression-testing.md`).

Issue #1592, opened 2026-06-11
bash-antipatterns: false positive on grep -l/-c over /tmp scratch files


laurigates:
## What

The `bash-antipatterns` PreToolUse hook BLOCKED a legitimate multi-file `grep -l` over downloaded scratch files in `/tmp`:

```
grep -lE 'NOTARY' /tmp/onb-*.js /tmp/ur-common.js /tmp/pu-common.js; cd /tmp && curl -sk -o ... && jq ...
```

It fired the `'grep -rn pattern src/' → use Grep tool` block, although the command was neither recursive nor a codebase search — it was file-list filtering (`-l`) over artifact bundles downloaded to `/tmp` during FE bundle reverse-engineering. The hook's own message says "Pipelines (… | grep …) and boolean checks (grep -q pattern file && do_thing) are still allowed", and `-l` mode is the same class of structured, non-search usage.

Workaround was splitting into two Bash calls — friction, not breakage, but it recurs whenever a session works with downloaded artifacts outside the repo.

## Suggested fix

Exempt from the codebase-search heuristic any of:

- `grep -l` / `grep -c` (file-list and count modes — these are filters, not searches the Grep tool replaces)
- targets resolving outside the repo / additional working dirs (`/tmp`, `/private/tmp`)

## Bonus (minor, same session)

The pipe-count REMINDER claimed "8 pipes" for a command with 2 real pipes — it appears to count `|` characters inside a `curl -w '%{http_code} (redirect: %{redirect_url})\n'`-style format string / quoted arguments. Worth excluding quoted strings from the pipe count if cheap.

Session: 2026-06-11, ForumViriumHelsinki/infrastructure (simpl-eval applicant onboarding).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2337, opened 2026-08-09
fix(blueprint-plugin): gate structural-cue code signals on file extension

### LINKED ISSUE #2336: blueprint-structural-cue: prose containing "export " trips the public-API cue — file_ext is computed but never used
## What

`blueprint-plugin/hooks/blueprint-structural-cue.sh` fires its structural-change
cue on **prose**. Signals 2–6 are source-code patterns, but they are grepped
against the payload of *any* file the Edit/Write touched, with no file-type
guard — so an English sentence containing "export " (or a line beginning `pub `)
trips the "public API changed" cue in a Markdown file.

Observed while drafting a GitHub issue body in a scratchpad directory outside any
repo:

```
[blueprint] Structural change detected (manifest / public API). Consider /blueprint:derive-plans or /blueprint:adr-validate to keep PRDs/ADRs current.
```

## Repro

```
export BLUEPRINT_STRUCTURAL_CUE_CACHE_DIR=$(mktemp -d) && printf '%s' '{"tool_name":"Write","session_id":"repro","tool_input":{"file_path":"/tmp/notes.md","content":"needs a test before the export can be called correct"}}' | bash blueprint-plugin/hooks/blueprint-structural-cue.sh
```

Emits the `{"decision":"block","reason":"[blueprint] Structural change detected …"}`
payload. The offending substring is `export ` inside ordinary prose — in the real
case, a Markdown **table cell** reading *"needs the same consumer-contract test
before the export can be called correct"*.

Control (a genuine structural edit) still fires correctly, so the detection
itself is fine — only its scope is wrong:

```
printf '%s' '{"tool_name":"Write","session_id":"ctl","tool_input":{"file_path":"/tmp/x.ts","content":"export interface Foo { a: string }"}}' | bash blueprint-plugin/hooks/blueprint-structural-cue.sh
```

## Root cause

The hook computes the extension and then never uses it:

```sh
base_name="${file_path##*/}"
file_ext="${file_path##*.}"    # assigned at line 50, referenced nowhere else
```

`grep -c 'file_ext'` over the hook returns **1**. `base_name` is used (the
`plugin.json` and `openapi.*` checks); `file_ext` is dead. That reads like an
intended file-type guard that was dropped, most likely when detection was widened
in #1616 from manifests + export lines to signals 2–6.

## Suggested fix

Gate the *code-shaped* signals (2–6) on a source extension, leaving the
path/basename signals (1, 7 — manifests, `.proto`, `openapi.*`) extension-free
since they already identify themselves:

```sh
case "$file_ext" in
    js|jsx|mjs|cjs|ts|tsx|rs|py|go) ;;   # code signals apply
    *) is_structural=0 ;;                # prose, config, data: signals 2-6 do not
esac
```

Two smaller things worth considering in the same pass:

1. **Scratchpad/out-of-repo paths.** The cue suggests `/blueprint:derive-plans`,
   which is meaningless for a file outside any blueprint repo. The session
   scratchpad (`/private/tmp/claude-*/…/scratchpad/`) is a known-inert location;
   an early exit there would suppress a class of cues that can never be
   actionable. This is the same lesson as #2052 and #1592 for
   `bash-antipatterns` — scratch files need their own carve-out.
2. **The `pub ` pattern is anchored but still prose-prone** —
   `^[+[:space:]]*pub ` matches a Markdown list item or table row starting with
   the word "pub". Extension gating fixes this too.

## Impact

Low severity by design — it is a cue, not a gate, and it is deduped to once per
session, so the cost is one spurious nudge. Worth fixing anyway because a
false-positive cue on prose trains the reader to skim past it, which is exactly
the signal-erosion the once-per-session dedup exists to prevent.

## Environment

- `blueprint-plugin` 3.44.0
- Fired during a session doing GitHub-issue authoring in `loractl`; no repo file
  was touched by the triggering Write.

laurigates:
## The defect

`blueprint-plugin/hooks/blueprint-structural-cue.sh` fired its **"public API changed"** cue on ordinary **prose**.

Signals 2–6 are source-code greps (`export `, `module.exports`, `export interface|type`, Go/Rust exported types, route registration), but they ran against the payload of *any* file an `Edit`/`Write` touched. So an English sentence containing `export ` tripped the public-API signal in a Markdown file — observed on a table cell reading *"…the same consumer-contract test before the export can be called correct"* — and the anchored `^[+[:space:]]*pub ` matched a Markdown list item beginning with the word "pub".

The dropped guard was visible in the source but not in the behaviour:

```sh
base_name="${file_path##*/}"     # used by the plugin.json / openapi.* checks
file_ext="${file_path##*.}"      # assigned at line 50, referenced NOWHERE else
```

`grep -c 'file_ext'` over the hook returned **1**. It was most likely lost when detection widened in #1616 from manifests + export lines to signals 2–6.

The failure is quiet by construction — a cue is not a gate and dedups to once per session — which is exactly why it survived. The cost is one spurious nudge that trains the reader to skim past the cue, eroding the signal the dedup exists to protect.

*(Amusingly, the hook fired on the very edit that added this fix: a `.sh` file whose new comments contain the word "export". Post-fix, that edit is silent.)*

## Verified repro — before and after

**Before** (`git show HEAD~1`), both prose and the genuine control fired identically:

```
export BLUEPRINT_STRUCTURAL_CUE_CACHE_DIR=$(mktemp -d) && printf '%s' '{"tool_name":"Write","session_id":"repro","tool_input":{"file_path":"/tmp/notes.md","content":"needs a test before the export can be called correct"}}' | bash blueprint-plugin/hooks/blueprint-structural-cue.sh
```
→ `{"decision":"block","reason":"[blueprint] Structural change detected …"}`  ❌

```
export BLUEPRINT_STRUCTURAL_CUE_CACHE_DIR=$(mktemp -d) && printf '%s' '{"tool_name":"Write","session_id":"ctl","tool_input":{"file_path":"/tmp/x.ts","content":"export interface Foo { a: string }"}}' | bash blueprint-plugin/hooks/blueprint-structural-cue.sh
```
→ `{"decision":"block","reason":"[blueprint] Structural change detected …"}`  ✅

**After**, the same two commands: prose is **silent, exit 0**; the `.ts` control still emits the cue. The only difference between the two payloads is the extension — which is precisely the invariant.

## What changed

**1. Signals 2–6 are gated on a source extension** (`js jsx mjs cjs ts tsx mts cts rs py pyi go`) via an `is_code_file` flag. Signals 1 and 7 — manifests (`plugin.json`, `marketplace.json`), `*.proto` / `*.graphql` / `schema.prisma`, `openapi.*` — stay **extension-free**: they identify themselves by path or basename.

The issue's suggested shape was **not** taken verbatim:

```sh
case "$file_ext" in
    js|jsx|…) ;;
    *) is_structural=0 ;;   # would clobber signal 1
esac
```

Placed after signal 1 this zeroes the `plugin.json` manifest verdict, since `.json` is not a source extension. The guard has to gate the **individual code signals**, not zero the accumulated result. A regression case pins this (`plugin.json manifest still fires though .json is not a code ext`).

**2. Early exit for the session scratchpad** (issue's secondary suggestion (a)) — `{/tmp,/private/tmp,/var/folders/*}/claude-*/…`, the harness's per-session scratch tree, where `/blueprint:derive-plans` can never be actionable. Deliberately **narrower** than `bash-antipatterns`' all-of-`/tmp` carve-out (#1592, #2052) so that a plain `/tmp/foo.ts` still gets the cue — the issue's own control uses that path, and a `assert_fires` case pins the narrowness.

**3. The `pub ` prose-proneness** (secondary suggestion (b)) needed no separate change — extension gating covers it, as the issue predicted. Pinned by its own assertion (`a Markdown list item starting with the word 'pub'`).

## How the regression test pins it

`blueprint-plugin/hooks/test-blueprint-structural-cue.sh` gains **26 assertions** (auto-discovered by `scripts/run-skill-script-tests.sh` via the `*/hooks/test-*.sh` glob).

They are **semantic**, not a grep for the string `is_code_file` — the #1417 lesson. Each case feeds a payload that is byte-for-byte code-shaped and asserts the **verdict turns on the extension alone**: `export interface Foo { a: string }` in `.md` / `.txt` / `.json` / `.yaml` / no-extension is silent, while the *same string* in `.ts` still fires.

**Guard integrity** — the suite cannot be satisfied by a hook that simply went silent everywhere:

| Group | Assertions | Pins |
|---|---|---|
| Prose stays silent | 9 | the defect itself, incl. the issue's verbatim repro and the `pub ` list item |
| Genuine source still fires | 7 | `.ts .js .go .rs .py .mjs .tsx` with the same payloads |
| Path/basename signals unaffected | 5 | manifests, `openapi.yaml`, `.proto`, `schema.prisma` — none swept up by the gate |
| Scratchpad inert + narrow | 4 | 3 silent under `claude-*` temp roots, 1 asserting `/tmp/x.ts` still fires |

**Mutation-verified** against `git show HEAD:<path>` (the pre-fix hook, same test file):

| | PASS | FAIL |
|---|---|---|
| Pre-fix hook | 48 | **12** |
| Post-fix hook | **60** | 0 |

All 48 pre-existing assertions stay green in both directions, so the 12 failures isolate to the new behaviour.

## Verification run

- `bash blueprint-plugin/hooks/test-blueprint-structural-cue.sh` → `PASS=60 FAIL=0 STATUS=OK`
- `bash scripts/run-skill-script-tests.sh` → `TOTAL=90 PASSED=90 FAILED=0 STATUS=OK`
- `shellcheck` on both scripts → exit 0; `scripts/lint-shell-scripts.sh` → 0 errors (9 pre-existing unrelated warnings)
- `pre-commit run --files <the five paths>` → no failures

## Docs landed in the same commit

Per `.claude/rules/docs-currency.md` — `blueprint-plugin/README.md`, `blueprint-plugin/docs/behavioral-cue-registry.md` (trigger summary), and a `.claude/rules/regression-testing.md` Known-Regressions row, as that rule requires for every fixed bug.

Closes #2336


laurigates:

laurigates:

Issue #2335, opened 2026-08-08
fix(scripts): check-taskcompleted-verification passes on a fully reverted verifier — a vacuous rule and union evaluation


laurigates:
## What

`scripts/check-taskcompleted-verification.sh` (added in #2308 to stop the #2301 fix silently reverting) has two holes that let a reverted verifier pass. Both are demonstrated below with executable mutations, not read off the source.

The guard is the regression check for #2301 — the whole point is that it fires when the committed-work-is-evidence behaviour is removed. Today it does not, in two distinct ways.

## Defect 1 — `worktree_still_inspected` passes vacuously (live now)

The rule (`scripts/check-taskcompleted-verification.sh:248`) accepts either spelling:

```sh
require_any "worktree_still_inspected" \
    "prompt no longer inspects the working tree at all" \
    'git status' \
    'working tree'
```

Both alternatives occur throughout the prompt in prose that is *not* the inspection step — `"Evidence of the work may live in COMMITS, not only in the working tree"`, `"fall back to the working tree"`, `"not raw git status"`. So the rule matches whether or not the prompt still inspects anything.

**Proof.** Delete the prompt's only working-tree inspection step — `3. Inspect the working tree: git status --porcelain and git diff --stat.` — and the guard is still clean:

```
$ bash scripts/check-taskcompleted-verification.sh --plugin-json=mutant.json --strict
STATUS=OK
ISSUE_COUNT=0
```

## Defect 2 — rules are evaluated over the *union* of all TaskCompleted prompts (latent)

The extractor joins every `type: "agent"` prompt into one blob before any rule runs (`:131`):

```python
sys.stdout.write("\n".join(prompts))
```

Every `require_any` then greps that blob, so a token carried by *any* hook satisfies the rule for *all* of them. A reverted verifier passes as long as some sibling hook mentions the words.

**Proof.** Replace the verifier prompt with the #2301 bug verbatim — *"Judge whether the task is complete by inspecting the working tree only… If the tree is clean, no work was done"* — and append a second, unrelated agent hook carrying the original text:

```
STATUS=OK
ISSUE_COUNT=0
```

The guard reports a clean bill of health on a verifier that has been fully reverted to the bug it exists to prevent.

This one is **latent, not currently exploitable**: `TaskCompleted` has exactly one `type: "agent"` prompt today, so the union is a single element. It becomes live the moment a second agent hook is added to that event — and nothing warns you at that point.

## Why this is the same class of defect the guard was written for

[...]

Issue #2301, opened 2026-08-07
feedback(hooks-plugin): TaskCompleted verification judges only the working tree, so committed work reads as no evidence


laurigates:
## Skill

`hooks-plugin` — **TaskCompleted → Implementation Verification** (`type: "agent"`), documented in `hooks-plugin/README.md` §"TaskCompleted — Implementation Verification (agent)".

## Category

Bug — the hook returns a factually wrong verdict, not merely a narrow one.

## Description

The verifier judges task completion by inspecting the **working tree**. Work that has been committed (and pushed) therefore presents as no work at all, and the hook blocks `TaskUpdate status=completed` with *"Agent hook condition was not met"*.

This inverts the incentive it should create: it penalises the workflow it exists to encourage — commit early, keep the tree clean — and pushes toward leaving work uncommitted so the check can see it.

## Evidence

Two fires in one session (`ForumViriumHelsinki/thelma`, 2026-08-07).

**1. Task "Fix Unit Tests job"** — verdict:

> The only modification in the working tree is to .mcp.json (whitespace reformatting of MCP server args arrays, plus adding an unrelated 'pal-mcp-server' entry). No source files, test files, or test/build configuration were touched. This does not address the task […] There is no evidence of a local repro, root-cause isolation, or an actual fix.

At that moment the fix was commit `67d36aa5` on `fix/ci-green-after-dep-pin`, already pushed and open as PR #1259 — a 20-file change that took the suite from *hanging with a heap OOM* to **214 files / 4,470 tests passing in 27s**. The working tree was clean precisely *because* the work was committed. The `.mcp.json` edit it fixated on was pre-existing user-owned state deliberately kept out of every commit.

**2. Task "Fix release-please conflicts workflow"** — the hook confirmed the workflow edit itself was correct and complete, then withheld completion because 22 *other* modified files were present. Those were a concurrent lint sweep plus the same pre-existing `.mcp.json`. Scope-checking the working tree is reasonable in principle, but it cannot distinguish "this agent's unrelated changes" from "state that was already there".

## Suggested Action

Before concluding no evidence exists, have the verifier consider **committed** work on the current branch, e.g.:

```sh
git log --oneline origin/<base>..HEAD          # commits this branch adds
git diff --stat origin/<base>...HEAD           # their cumulative effect
```

[...]

PR #2338, opened 2026-08-09
fix(scripts): evaluate TaskCompleted verification rules per prompt

laurigates:
## What

`scripts/check-taskcompleted-verification.sh` — the regression guard added by #2308 so the #2301 fix cannot silently revert — reported `STATUS=OK ISSUE_COUNT=0` on a **fully reverted verifier**, through two independent holes. Both are closed here, per items (1) and (2) of #2335.

- **Defect 1 (live):** `worktree_still_inspected` accepted the bare tokens `git status` and `working tree`, both of which occur throughout the prompt in prose that is *not* the inspection step. It now requires the imperative instruction.
- **Defect 2 (latent):** the extractor joined every `type: "agent"` TaskCompleted prompt into one blob before any rule ran. Prompts are now extracted individually and the rule set runs against the **verifier alone**.

---

## Proof 1 — deleting the working-tree inspection step

Mutation: remove the prompt's only inspection step, `3. Inspect the working tree: git status --porcelain and git diff --stat.` Everything else is untouched — including the words `working tree` (still on four other lines) and `git status` (still present, inside the *negation* `not raw git status`).

**Before**

```
$ bash scripts/check-taskcompleted-verification.sh --plugin-json=mutant1.json --strict
PROMPT_FOUND=true
RULES_CHECKED=15
STATUS=OK
ISSUE_COUNT=0
EXIT=0
```

**After**

```
$ bash scripts/check-taskcompleted-verification.sh --plugin-json=mutant1.json --strict
AGENT_PROMPT_COUNT=1
VERIFIER_ID=agent[0]
VERIFIER_SOURCE=sole-agent-hook
STATUS=ERROR
ISSUE_COUNT=1
ISSUES:
  - SEVERITY=ERROR PROMPT=agent[0] RULE=worktree_still_inspected MSG=prompt no longer issues a working-tree inspection command (e.g. git status --porcelain / git diff --stat of the uncommitted state)
EXIT=1
```

The rule now accepts either a working-tree-scoped `git status` invocation carrying a machine-readable flag (`--porcelain` / `--short` / `-s` / `-uall` / `--untracked` — forms prose about the tree does not use), or an imperative inspection verb bound to a `git status` / `git diff` call with **no sentence boundary** (`[^.;]*`) between them. The scope paragraph's `— not raw git status;` clears neither: it carries no flag, and its nearest preceding verb is `judge`.

## Proof 2 — reverted verifier with a token-carrying sibling hook

Mutation: replace the verifier prompt with the #2301 bug verbatim (*"Judge whether the task is complete by inspecting the working tree only… If the tree is clean, no work was done"*) **and** append a second, unrelated `TaskCompleted` agent hook carrying the original shipped text.

**Before**

```
$ bash scripts/check-taskcompleted-verification.sh --plugin-json=mutant2.json --strict
PROMPT_FOUND=true
PROMPT_CHARS=3748          <- the two prompts, concatenated
RULES_CHECKED=15
STATUS=OK
ISSUE_COUNT=0
EXIT=0
```

**After**

```
$ bash scripts/check-taskcompleted-verification.sh --plugin-json=mutant2.json --strict
AGENT_PROMPT_COUNT=2
VERIFIER_ID=agent[0]
VERIFIER_SOURCE=statusMessage
PROMPT_CHARS=401           <- the verifier alone
STATUS=ERROR
ISSUE_COUNT=11
ISSUES:
  - SEVERITY=ERROR PROMPT=agent[0] RULE=branch_log_inspection ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=committed_work_is_evidence ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=base_ref_resolution ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=base_ref_merge_base ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=base_ref_unresolved_path ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=clean_tree_is_success ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=worktree_fallback_preserved ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=scope_against_branch_diff ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=scope_not_raw_status ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=preexisting_files_not_violations ...
  - SEVERITY=ERROR PROMPT=agent[0] RULE=still_blocks_without_evidence ...
EXIT=1
```

11 of 15 rules fire, each naming which prompt failed. (`worktree_still_inspected` correctly does **not** fire — the reverted prompt genuinely says "Run git status and git diff --stat"; inspecting the tree was never the bug, judging *only* from it was.)

### Identifying the verifier, rather than inferring it

`plugin.json`'s agent hook already carried `"statusMessage": "Verifying task implementation quality..."` — a real, harness-honoured agent-hook field (`.claude/rules/prompt-agent-hooks.md`), not a synthetic key — so **no `plugin.json` change was needed**. Resolution:

| Agent hooks on `TaskCompleted` | Verifier |
|---|---|
| 1 | that hook (`VERIFIER_SOURCE=sole-agent-hook`; no marker required) |
| ≥2, exactly one `statusMessage` matching `/verif/i` | that hook (`VERIFIER_SOURCE=statusMessage`) |
| ≥2, **none** marked | `RULE=verifier_unidentified`, ERROR, exit 1 |
| ≥2, **several** marked | `RULE=verifier_ambiguous`, ERROR, exit 1 |

The guard never picks a prompt by position and never unions. New output keys: `AGENT_PROMPT_COUNT`, `VERIFIER_ID`, `VERIFIER_SOURCE`; every finding carries `PROMPT=<id>`.

## Proof 3 — a behaviour-preserving reword still passes

The existing whole-prompt reword (TEST C, unchanged) still passes. Three new cases target the newly anchored rule specifically, replacing step 3 in place:

| Case | Replacement for step 3 | Result |
|---|---|---|
| **N1** | `3. Now survey the uncommitted state — git status --short, then git diff --stat.` | `STATUS=OK` |
| **N2** | `3. Read git status and git diff to see what is still uncommitted.` | `STATUS=OK` |
| **N3** | `3. Remember that the working tree may also carry changes; git status output is not authoritative here.` | fires `worktree_still_inspected` |

N2 carries no machine-readable flag at all and still passes, so the anchor did not collapse into a flag grep. N3 is the discrimination proof: prose *about* the working tree, in the same position, carrying both old tokens, must still fail — without it, N1/N2 would pass just as well against a rule quietly relaxed back to the old vocabulary match.

---

## Audit of the other 12 `require_any` rules

Mechanical: for every rule, every alternative was matched against the shipped prompt line-by-line and the satisfying line classified as *the instruction the rule names* vs *prose about it*.

| Rule | Satisfied by | Verdict |
|---|---|---|
| `branch_log_inspection` | `git log --oneline BASE..HEAD` on the instruction line only | OK — anchored on a command + range |
| `committed_work_is_evidence` | the assertion lines | OK — the concept **is** a statement; prose is the correct anchor |
| `base_ref_resolution` | `symbolic-ref` / `refs/remotes/origin/HEAD` / `rev-parse --verify`, instruction line only | OK — command tokens |
| `base_ref_merge_base` | `merge-base` on the instruction line | OK today; see finding B |
| `base_ref_unresolved_path` | `no base ref` on two genuine handling lines | **FIXED** — see finding A |
| `clean_tree_is_success` | co-occurrence of "clean tree" + success/negation vocabulary | OK — statement concept, requires two terms on one line |
| `worktree_fallback_preserved` | `fall back to the working tree` on the instruction line | OK — the phrase *is* the instruction |
| `worktree_still_inspected` | `working tree` on 5 lines (4 prose), `git status` on a negation | **FIXED** — the reported defect |
| `scope_against_branch_diff` | co-occurrence of `scope` + `...HEAD` on the instruction line | OK — see finding C |
| `scope_not_raw_status` | a negation immediately preceding `git status` | OK — the only rule already anchored on polarity |
| `preexisting_files_not_violations` | `not (a )?scope violation`, a specific claim phrase | OK |
| `still_blocks_without_evidence` | `no supporting change` / `neither commits … nor` | OK |
| `blocking_response_preserved` | literal `"ok": false` | OK |

**Finding A — fixed.** `base_ref_unresolved_path` accepted `no upstream`. That names a *different* condition — the branch having no `@{u}`, which this prompt explicitly forbids using as a base — so a reword of the `@{u}` anti-pattern sentence ("if there is no upstream…") would satisfy the rule while the actual unresolved-base handling had been deleted. The spelling is currently unmatched, and the surviving alternatives (`neither resolve`, `no base ref`, `(cannot|…) resolve`) cover both the shipped prompt and the reword fixture, so removal is behaviour-neutral here and strictly narrowing. Dropped.

**Finding B — reported, not fixed.** `base_ref_merge_base` accepts `divergence point`, pure prose vocabulary; `"BASE is **not** the divergence point"` would satisfy it. It is currently unmatched (`merge-base` carries the rule), so a tightening cannot be validated against real evidence, and removing it narrows reword tolerance for a concept whose natural English spelling that is. Left as-is deliberately.

**Finding C — reported, not fixed; a property of the design, not a bug.** Nine rules are **polarity-blind**: they require a token or co-occurrence but cannot tell an assertion from its negation, so a prompt stating the *opposite* of an invariant matches the same words. #2335 names this directly. It is not repairable within a keyword design — closing it is exactly what deferred item (3) is for. `scope_not_raw_status` is the one rule that already anchors on polarity, and it does so only because the invariant it pins is itself a negation.

---

## Deferred item (3) — an executable gate

**Not attempted here, and #2335 should stay open for it.** Items (1) and (2) are the "small and independently landable" pair; item (3) is a different kind of artifact — sandbox repos built in the three base-ref states the prompt must handle (branch ahead of base, clean tree with commits, no base ref resolvable), asserting the *verdicts* rather than the wording. It is the only version that cannot be satisfied by a prompt stating the inverse, and so the only thing that closes finding C above.

This PR therefore says **`Refs #2335`, not `Closes`**. What lands here removes both proven false negatives and tightens one latent third; what remains open is the design change from a textual gate to an executable one, plus finding B as a small rider.

---

## Tests

`scripts/tests/test-check-taskcompleted-verification.sh`: **36 → 61 assertions**, all green; the 36 pre-existing ones are unmodified.

- **K** — mutation 1. K4/K5 are the fixture-validity half and are load-bearing: the mutant must **still contain** `working tree` and `git status`, so the pass proves the rule stopped accepting vocabulary rather than the vocabulary having vanished.
- **L** — mutation 2. L9 asserts the sibling genuinely carries the shipped tokens (else the union case is vacuous). **L10 is the attribution control**: the same two-hook shape with an *intact* verifier must still pass, so L1's ERROR is the reverted verifier's and not merely a consequence of having two hooks.
- **M** — verifier identification, weighted on the cases a naive fix gets wrong. M4/M5 put the intact verifier **second**, behind a reverted one — a guard that took index 0, or unioned, reports ERROR. M6/M7 pin that a sole hook needs no marker, so dropping `statusMessage` cannot turn the guard into a false alarm.
- **N** — reword tolerance and its discrimination proof, above.

**Mutation-verified twice**, per `.claude/rules/regression-testing.md`:

1. Against the **verbatim pre-fix guard** (`git show origin/main:scripts/check-taskcompleted-verification.sh`) with the new suite: `PASSED=44 FAILED=17` — every new behavioural assertion fails, while the guard-integrity and fixture-validity ones (K4, K5, L9, L10, M5, M7, N1, N2) pass on **both** versions, which is what an unchanged-behaviour control should do.
2. Against a copy with **only** the two `worktree_still_inspected` patterns reverted, per-prompt evaluation intact: `PASSED=57 FAILED=4` — exactly K1–K3 and N3. Each half of the fix is attributed to its own change.

Also run clean: `shellcheck` (both files, exit 0), `scripts/lint-shell-scripts.sh` (0 errors), `scripts/check-git-sandbox-guards.sh --strict`, `scripts/tests/test-guard-worktree-root-discovery.sh` (27/0). No new wiring needed — the guard and its self-test are already in `.pre-commit-config.yaml` and `plugin-pr-checks.yml`.

`.claude/rules/regression-testing.md` gains the Known Regressions row.

Refs #2335


laurigates:

Issue #2359, opened 2026-08-10
fix(session-plugin): session-end-nudge bypasses the collector's scope ladder with a raw basename project query


laurigates:
## What

`session-plugin/hooks/session-end-nudge.sh:77-82` decides whether to mention a taskwarrior sync pass by running its own query, built from the bare directory basename:

```sh
repo_root_tw=$(git -C "$cwd" rev-parse --show-toplevel 2>/dev/null || echo "$cwd")
proj=$(basename "$repo_root_tw")
task_count=$("$task_bin" project:"$proj" '(status:pending or +ACTIVE)' export 2>/dev/null \
    | jq 'length' 2>/dev/null || echo 0)
[ "${task_count:-0}" -gt 0 ] && has_open_tasks=1
```

This is the exact detection `session-survey.sh` exists to do correctly, reimplemented without any of its guards. The collector carries a full scope ladder — exact slug → `PROJECT_RESOLVED` from the git remote → `all-projects-fallback` — plus `TASK_SCOPE`, `PROJECT_CONFIDENCE`, `PROJECT_EXACT_TASKS`, `PROJECT_AMBIGUOUS`, and `PROJECT_PREFIX_SIBLINGS`. The hook has none of it.

## Two failure directions, both silent

**Under-count → the nudge stays quiet.** Any repo whose basename ≠ taskwarrior slug returns 0 and the cue is dropped, so the user is not offered the sync pass precisely when their tasks are somewhere the naive query cannot see. This is the class of #2271 (basename ≠ slug), #2304 (nested repo, tasks on the parent), #2232 (portfolio checkout) and #2355 (separator variant, open) — all fixed *in the collector*, none reaching this hook.

**Over-count → the cue fires on someone else's backlog.** `task project:X` is a **prefix** match (#2323), so in a `comfyui` checkout the query sweeps in all 60 `comfyui-nodes` tasks and the nudge claims open work that belongs to a sibling project.

The hook is offer-only and never blocks, so neither is dangerous — but the cue is either missing or wrong exactly where the collector already knows the answer.

## Second instance in the cue text

The `task_cue` string the hook injects (line 103) instructs the agent to run:

```
task project:<name> '(status:pending or +ACTIVE)' export | jq '.[]'
```

propagating the same prefix-match hazard into the session, this time as guidance the agent follows. It also duplicates the per-task done/update/leave loop that #2356's design replaces with a single item-level confirmation, so this line likely wants deleting rather than fixing.

[...]

Issue #2358, opened 2026-08-10
fix(session-plugin): blueprint auto-drain gate ignores task_registry .enabled, so a disabled task still auto-runs


laurigates:
## What

`session-end`'s blueprint auto-drain gate reads two manifest fields and skips a third that the schema declares and that means "do not run this task".

`session-plugin/skills/session-end/SKILL.md:103`:

```sh
jq -r 'if ((.automation.autonomy_level // 0) >= 1) and (.task_registry["feature-tracker-sync"].auto_run == true) then "auto" else "ask" end' docs/blueprint/manifest.json 2>/dev/null
```

`blueprint-plugin/schemas/manifest.schema.json` → `definitions.taskRegistryEntry` declares `enabled` (boolean) alongside `auto_run`. The gate never reads it.

## Effect

A manifest with:

```json
"feature-tracker-sync": { "enabled": false, "auto_run": true }
```

evaluates to `auto`, so `session-end` **auto-confirms the pass, runs it without asking, and reports a receipt line** — for a task the repo owner switched off. `auto_run` is the *how* (without asking); `enabled` is the *whether*. Reading only the former inverts the owner's opt-out into an unattended write.

This is the one pass `session-end` deliberately excludes from its Step 3 confirmation, so there is no second gate behind it. Blast radius is bounded — the drain is a `git`-tracked edit to `docs/blueprint/feature-tracker.json` and `git restore`-able — but it is unattended and silent.

## Scope

Affects only repos that set `enabled: false` while leaving `auto_run: true` at `autonomy_level >= 1`. That combination is unusual, which is why this is a latent bug rather than a live incident. This repo is at `autonomy_level: 1` and is a consumer of the gate, so it is worth confirming its own manifest before/after.

## Suggested fix

```sh
jq -r 'if ((.automation.autonomy_level // 0) >= 1)
       and (.task_registry["feature-tracker-sync"].enabled == true)
       and (.task_registry["feature-tracker-sync"].auto_run == true)
       then "auto" else "ask" end' docs/blueprint/manifest.json 2>/dev/null
```

Decide deliberately whether a **missing** `enabled` reads as enabled or disabled. `== true` treats absent as disabled, which is the safe reading and consistent with the `// 0` default on `autonomy_level` — but it changes behaviour for any manifest that omits the key, so it is the part worth a second look rather than the `.enabled` addition itself.

[...]

Issue #2355, opened 2026-08-10
feedback(session-plugin): session-survey misses separator-variant project slugs (fvh-cost-attribution vs fvh.cost-attribution)


laurigates:
## What

`session-survey.sh` recovers from a wrong detected project slug in two ways today: it degrades to `TASK_SCOPE=all-projects-fallback` + `PROJECT_CONFIDENCE=low` (#2271), and it surfaces `PROJECT_AMBIGUOUS` when a **named ancestor** slug owns tasks (#2304). Neither catches a slug that differs from the detected one only by **separator**.

## Repro

Real session, 2026-08-10, in `~/repos/ForumViriumHelsinki/fvh-cost-attribution`:

```
$ session-survey.sh
OPEN_TASKS=0
TASK_SCOPE=all-projects-fallback
PROJECT_CONFIDENCE=low
PROJECT_REMOTE_NAME=fvh-cost-attribution

$ session-survey.sh --project fvh.cost-attribution
OPEN_TASKS=10
TASK_SCOPE=project
PROJECT_CONFIDENCE=high
```

The repo is `fvh-cost-attribution` (dash) — the directory basename *and* the git remote name agree. The taskwarrior project is `fvh.cost-attribution` (dot). Ten pending tasks, invisible.

`fvh.cost-attribution` is not an ancestor of the cwd, so `PROJECT_AMBIGUOUS` cannot fire; it is not a prefix of the detected slug either, so #2323's check does not apply.

## Why the existing guards were not enough in practice

They worked exactly as designed — `PROJECT_CONFIDENCE=low` was correct, and the `/session-end` skill instructs the agent to treat a low-confidence zero as an unqueried project. In this session that instruction was followed and the tasks were found.

But recovery cost a manual `task export | jq 'group_by(.project)'` sweep and a guess about which of 40+ slugs was the right one. The information needed to skip that was already in hand: the detected slug and the real one differ by one character class.

## Suggestion

Extend the `PROJECT_AMBIGUOUS` candidate set beyond ancestors with a **separator-normalized** lookup — fold `.`, `-`, `_` (and case) before comparing, then emit any real slug that collides:

```
PROJECT_AMBIGUOUS=fvh.cost-attribution
PROJECT_AMBIGUOUS_TASKS=10
PROJECT_AMBIGUOUS_REASON=separator-variant
```

[...]

PR #2365, opened 2026-08-11
fix(session-plugin): session-end and session-wrap request GITHUB_DRIFT via --with-dedup

### LINKED ISSUE #2357: fix(session-plugin): session-end and session-wrap build on GITHUB_DRIFT but never pass --with-dedup, so the section is always absent
## What

`session-survey.sh` emits the `GITHUB_DRIFT` section **only** under `--with-dedup` (`session-plugin/scripts/session-survey.sh:1012`). `session-spinup` is the only skill that passes the flag. `session-end` and `session-wrap` never pass it — but both build a redundancy rule on the section's contents, so that rule silently runs on no data.

```
$ for f in session-plugin/skills/*/SKILL.md; do \
    printf '%-16s dedup=%s\n' "$(basename $(dirname $f))" "$(grep -c -- '--with-dedup' $f)"; done
session-distill  dedup=0
session-end      dedup=0
session-spinup   dedup=2
session-wrap     dedup=0

$ grep -rln GITHUB_DRIFT session-plugin/skills/
session-end/SKILL.md
session-wrap/SKILL.md
session-wrap/REFERENCE.md
session-spinup/SKILL.md
session-spinup/REFERENCE.md
```

Two of the three skills that *read* `GITHUB_DRIFT` never *request* it.

## Why it matters

Both call sites are the "don't duplicate an existing tracker" guard — the rule that stops a wrap filing a taskwarrior task for work an assigned GitHub issue already tracks:

- `session-wrap/SKILL.md` — *"the Step 1 digest's `PRS` section carries every open PR … and `GITHUB_DRIFT` every assigned-but-untracked issue"*, with a decision-table row **"Assigned GitHub issue, no local task → nothing — `GITHUB_DRIFT` surfaces it at spinup"**, and an explicit instruction to *"check a candidate's PR/issue number against `PRS` / `GITHUB_DRIFT` before adding."*
- `session-end/SKILL.md` Step 4.1 — *"an open PR or assigned issue in the Step 1 digest's `PRS` / `GITHUB_DRIFT` sections is already its own tracker — annotate, never mirror it into a task."*

With the section absent, the assigned-issue half of that check has nothing to match against. It fails in the **duplicate-creating** direction: an assigned issue is invisible, so wrap adds a task mirroring it, which is exactly the behaviour the section was added to prevent. The `PRS` half still works (that section is unconditional), which is likely why this went unnoticed — the guard appears to function.

## Repro

In a repo with an open GitHub issue assigned to you and no matching local task:

```sh
bash session-plugin/scripts/session-survey.sh --with-commits | grep -c '=== GITHUB_DRIFT ==='   # 0 — what session-end/session-wrap run
bash session-plugin/scripts/session-survey.sh --with-commits --with-dedup | grep -c '=== GITHUB_DRIFT ==='   # 1
```

## Suggested fix

Add `--with-dedup` to the collector invocation in `session-end/SKILL.md` Step 1 and `session-wrap/SKILL.md` Step 1 (both the Execution step and the Agentic Optimizations row).

Note this makes the section's `GH_READY=` gate load-bearing for two more skills: when `gh` is unavailable the section is present but empty, and neither skill may then treat "no drift" as "no assigned issues" — that is the standard suppress-don't-conclude handling those skills already apply to `PROJECT_CONFIDENCE=low`.

## Regression guard

`plugin-compliance-check.sh` `check_skill_body()` for `session-end` and `session-wrap`: assert each body's `session-survey.sh` invocation carries `--with-dedup`. A semantic pairing is better than a bare string pin — assert that any skill body referencing `GITHUB_DRIFT` also passes the flag that emits it, so the next skill to read the section cannot repeat this.

## Provenance

Found while grounding the proposal-record design in #2356 (design doc: `docs/plans/session-end-proposal-records.md`). Distinct from the collector-side slug work in #2355 / #2271 / #2304 / #2323 — nothing here is about project scoping; the section is simply never requested.

Plugin: `session-plugin` · `skills/session-end/SKILL.md`, `skills/session-wrap/SKILL.md`

laurigates:
Fixes #2357

Summary

session-survey.sh only populates its GITHUB_DRIFT section when invoked with --with-dedup. Both session-end/SKILL.md and session-wrap/SKILL.md build a dont duplicate an existing GitHub tracked issue into a taskwarrior task guard on that sections contents, but neither passed --with-dedup when invoking the collector, so the guard silently ran against an always empty section. Only session-spinup/SKILL.md passed the flag before this change.

Changes

- session-end/SKILL.md: added --with-dedup to the Step 1 Execution invocation of session-survey.sh and to the matching Agentic Optimizations table row. Added a clause noting that GH_READY=false means GITHUB_DRIFT is present but unqueried, mirroring the existing PROJECT_CONFIDENCE=low handling, so an empty section under GH_READY=false is never treated as evidence there is nothing to dedup against.
- session-wrap/SKILL.md: same flag addition to its Step 1 Execution invocation and Agentic Optimizations row, plus the matching GH_READY=false clause next to the existing Dont duplicate an existing tracker section (which already reads GITHUB_DRIFT and PRS).
- scripts/plugin-compliance-check.sh (check_skill_body): added a semantic regression guard scoped to session-end and session-wrap. It asserts that any SKILL.md body referencing GITHUB_DRIFT also carries --with-dedup on every session-survey.sh invocation LINE, not merely as a substring anywhere in the file (a body that only mentions --with-dedup in explanatory prose still fails, proving this is a semantic pairing check and not a bare grep).
- scripts/tests/test-plugin-compliance-check.sh: extended with a fixture case named literally session-wrap (the check keys off the skill directory basename). Three cases: (1) GITHUB_DRIFT referenced with no --with-dedup on the invocation line raises the new finding, (2) --with-dedup present only in prose (not on the invocation line) still raises it, proving the semantic pairing, and (3) --with-dedup on the actual invocation line clears the finding and the fixture exits 0.

This issue is distinct from #2355 (already separately handled) — nothing here is about project-slug scoping, only the missing flag.

Repro (per the acceptance criteria)

Verified locally in this sandbox, where gh is not installed:

bash session-plugin/scripts/session-survey.sh --with-commits | grep -c '=== GITHUB_DRIFT ==='
→ 0

bash session-plugin/scripts/session-survey.sh --with-commits --with-dedup | grep -c '=== GITHUB_DRIFT ==='
→ 1

The GITHUB_DRIFT section header appears exactly once --with-dedup is passed, confirming the issue's claim and confirming this does not require gh to be present — with gh absent, the section still emits with GH_READY=false, ASSIGNED_ISSUES=0, DRIFT_COUNT=0, STATUS=OK rather than being omitted. This matches the acceptance criteria's expectation.

Testing

- bash scripts/tests/test-plugin-compliance-check.sh — 17/17 pass, including the three new session-wrap fixture cases. Mutation-verified: with the new check block removed, exactly the three positive-detection assertions fail (14 pass / 3 fail) while the fix and negative-control assertions still pass, confirming the guard is load-bearing.
- just lint-compliance — exit 0, zero ❌ issues repo-wide; session-plugin's Body column is ✅ confirming the real session-end/session-wrap bodies pass the new guard after the fix (WARN-level size/description recommendations for those skills are pre-existing and unrelated).
- bash session-plugin/scripts/tests/test-session-survey.sh — PASS=322 FAIL=0 (unaffected by this change; the collector script itself was not modified).
- just test-skill-scripts — session-plugin and scripts/tests/test-plugin-compliance-check.sh all PASS. The one FAILED entry in that run (hooks-plugin/hooks/test-bash-antipatterns.sh) and several SKIPs (ast-grep/cargo-generate/task CLI not installed in this sandbox) are pre-existing environment gaps, verified by re-running that same test against origin/main with these changes stashed — it fails identically (138 passed, 37 failed) with no changes present.
- just lint-shell — 0 errors, 9 pre-existing warnings in an unrelated plugin (macos-plugin), unaffected by this change.

Branched from a fresh git fetch origin && git switch -c … origin/main, so the diff is against current main content (the two sibling PRs for #2358/#2359 mentioned in the task description had not yet landed on origin/main at the time this branch was cut — git log shows no commits referencing #2358 or #2359 on origin/main, so nothing needed to be rebased onto).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EJqsYc8voEa6AA5Tib3Htu)_

### DIFF
--- a/scripts/plugin-compliance-check.sh
+++ b/scripts/plugin-compliance-check.sh
@@ -488,6 +488,29 @@ check_skill_body() {
       fi
     fi
 
+    # Regression: session-end and session-wrap both read the collector's
+    # GITHUB_DRIFT section as a dedup guard (redundant-tracker test / Step 4
+    # taskwarrior-sync), but session-survey.sh only populates GITHUB_DRIFT when
+    # invoked with --with-dedup — without the flag the section is always empty
+    # and the guard silently runs against nothing (issue #2357). Semantic
+    # pairing check, not a bare string search: a body referencing GITHUB_DRIFT
+    # must carry --with-dedup on EVERY session-survey.sh invocation LINE, so a
+    # skill that merely mentions --with-dedup in explanatory prose (without
+    # putting it on the actual command) still fails.
+    if [ "$skill_name" = "session-end" ] || [ "$skill_name" = "session-wrap" ]; then
+      if grep -q "GITHUB_DRIFT" "$skill_file"; then
+        local invocation_lines missing_dedup_lines
+        invocation_lines=$(grep -v '^[[:space:]]*>' "$skill_file" | grep "session-survey\.sh" || true)
+        if [ -n "$invocation_lines" ]; then
+          missing_dedup_lines=$(printf '%s\n' "$invocation_lines" | grep -v -- '--with-dedup' || true)
+          if [ -n "$missing_dedup_lines" ]; then
+            issues+=("❌ ${plugin}/${skill_name}: SKILL.md references GITHUB_DRIFT but at least one session-survey.sh invocation line is missing --with-dedup (issue #2357)")
+            has_errors=true
+          fi
+        fi
+      fi
+    fi
+
     # Regression: session-distill's mechanical substrate is distill-survey.sh —
     # the read-only collector that mines the session transcript for recipe
     # candidates / hot files / process groupings so the skill judges instead of
--- a/scripts/tests/test-plugin-compliance-check.sh
+++ b/scripts/tests/test-plugin-compliance-check.sh
@@ -241,30 +241,6 @@ make_session_wrap_fixture() {
     printf -- '---\n'
     printf 'name: session-wrap\n'
     printf 'description: Fixture session-wrap. Use when exercising the compliance self-test.\n'
-# --------------------------------------------------------------------------
-# Regression test for the session-end blueprint auto-drain gate (issue #2358).
-#
-# The jq gate that auto-confirms the Blueprint tracker-sync pass must require
-# ALL THREE of autonomy_level >= 1, task_registry[...].enabled == true, and
-# task_registry[...].auto_run == true — not just the first and third. A task
-# the owner disabled (`enabled: false`) must never auto-run unattended just
-# because `auto_run: true` and `autonomy_level >= 1`.
-#
-# THE SEMANTIC INVARIANT UNDER TEST: the two-field form (autonomy_level +
-# auto_run, missing the enabled check) must be REJECTED (❌, exit 1), and the
-# three-field form must PASS clean. A syntactic pin on "the string 'enabled'
-# appears somewhere in the body" would be insufficient — prose that merely
-# mentions "enabled" without wiring it into the actual jq filter must still
-# fail, so the fixture below carries prose naming "enabled" beside a gate line
-# that omits the check from the filter itself.
-make_session_end_skill() {
-  local gate_line="$1"
-  local dir="$root/$PLUGIN/skills/session-end"
-  mkdir -p "$dir"
-  {
-    printf -- '---\n'
-    printf 'name: session-end\n'
-    printf 'description: Fixture session-end. Use when exercising the compliance self-test.\n'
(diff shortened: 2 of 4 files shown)


laurigates:

Issue #2323, opened 2026-08-08
session-survey: PROJECT_CONFIDENCE=high can report a prefix-matched sibling's task count


laurigates:
## What

`scripts/session-survey.sh` reports a confident task count that can belong to a **different project than the slug it names**, because taskwarrior `project:` filters are **prefix matches**.

Observed this session in `~/repos/laurigates/comfyui-nodes`:

```
PROJECT=comfyui-nodes
TASK_SCOPE=project
PROJECT_CONFIDENCE=high
OPEN_TASKS=60
```

All four are true and consistent. But the repo's own `CLAUDE.md` documented the slug as `project:comfyui`, and the actual distribution was:

```
60 tasks   comfyui-nodes
 1 task    comfyui
```

`task project:comfyui list` returns all 60, so the wrong slug looked verified. A prior commit had enshrined it in `CLAUDE.md` on exactly that evidence ("the slug that exists"), and this session then filed four follow-ups into the near-empty sibling before the survey's own per-project view exposed the split.

## Why the current signal doesn't catch it

`PROJECT_CONFIDENCE` answers *"is the detected directory-basename slug trustworthy?"* It cannot answer *"do these tasks actually carry that slug?"* — and a prefix hit makes those two questions look like one. The failure is silent in the direction that matters: a **misfiled backlog reads as a verified one**.

## Suggestion

Emit an exact-value count alongside the filtered one, and flag divergence:

```sh
task export | jq -r --arg p "$PROJECT" '[.[] | select(.project == $p)] | length'
```

Then either add `PROJECT_EXACT_TASKS=` next to `OPEN_TASKS=`, or downgrade `PROJECT_CONFIDENCE` to `low` when the prefix count exceeds the exact count — plus a `PROJECT_PREFIX_SIBLINGS=` list so the session can surface the split rather than pick a side silently.

## Impact

Low frequency, but it corrupts the artifact the wind-down pass exists to produce: follow-ups land in a project nobody reads. It also survives across sessions, since the wrong slug gets written into `CLAUDE.md` and the next session trusts it.

Environment: session-plugin 1.9.3, taskwarrior 3.x, macOS.

Issue #2304, opened 2026-08-07
session-plugin: cwd-basename project detection reports OPEN_TASKS=0 for a nested repo whose tasks live under the parent workspace


laurigates:
## What

`session-survey.sh` derives the taskwarrior project from the cwd repo basename. In a portfolio where a **nested** repo's tasks are filed under the **parent workspace's** project slug, that reports `OPEN_TASKS=0` when the real backlog is non-empty — so `/session-end` silently disqualifies its own Taskwarrior-sync pass.

## Reproduction

Same machine, same taskwarrior store, two cwds — one nested inside the other:

```
$ cd ~/repos/laurigates/comfyui-nodes/comfyui-gallery-loader && session-survey.sh
PROJECT=comfyui-gallery-loader
DETECTION=cwd-repo-basename
OPEN_TASKS=0                       ← reported

$ cd ~/repos/laurigates/comfyui-nodes && session-survey.sh
PROJECT=comfyui-nodes
OPEN_TASKS=54                      ← reality for the same work

$ task project:comfyui-nodes status:pending count
54
```

Both are git repos. `comfyui-gallery-loader` is one of eleven packs nested under the `comfyui-nodes` workspace, and `~/.claude/rules/taskwarrior-tracking.md` asks for "one short lowercase slug per **long-running area**, reused consistently" — so all eleven packs file under `project:comfyui-nodes`. There is no `project:comfyui-gallery-loader`, and shouldn't be; per-pack projects would fragment one backlog into eleven.

## Impact

Hit during a real `/session-end` today. The session's own follow-up task was filed under `project:comfyui-nodes`; the survey ran from the pack directory, reported `OPEN_TASKS=0`, and by the skill's own gate (`TASK_AVAILABLE=true AND OPEN_TASKS >= 1`) the Taskwarrior-sync pass did not qualify. I only noticed because I'd created the task earlier in the same session and knew it existed.

That's the bad failure mode: not a wrong answer, an **empty** one — which is exactly what the "silently skip passes that don't qualify" instruction is designed to look like. A fresh session would have accepted the skip without a second thought.

## Suggested fix

In rough order of preference:

1. **Walk up for the nearest ancestor with tasks.** Having derived `PROJECT` from the basename, if it yields zero open tasks, try each ancestor repo basename up to `$HOME` and use the first with a non-empty count. Reports the effective slug plus a `DETECTION=cwd-repo-basename-ancestor` marker so the skill can say which it used.
[...]

PR #2377, opened 2026-08-13
fix(session-plugin): catch separator-variant project slugs in session-survey

### LINKED ISSUE #2355: feedback(session-plugin): session-survey misses separator-variant project slugs (fvh-cost-attribution vs fvh.cost-attribution)

laurigates:
## Problem

`session-survey.sh` recovers from a wrong detected project slug two ways today: it degrades to `TASK_SCOPE=all-projects-fallback` + `PROJECT_CONFIDENCE=low` (#2271), and it names an ancestor slug that owns the work via `PROJECT_AMBIGUOUS` (#2304, with the prefix-sibling split from #2323). Neither can see a slug that differs from the detected one only by **separator character**.

The repro from #2355 (real session, `~/repos/ForumViriumHelsinki/fvh-cost-attribution`):

```
$ session-survey.sh
OPEN_TASKS=0
TASK_SCOPE=all-projects-fallback
PROJECT_CONFIDENCE=low
PROJECT_REMOTE_NAME=fvh-cost-attribution

$ session-survey.sh --project fvh.cost-attribution
OPEN_TASKS=10
TASK_SCOPE=project
PROJECT_CONFIDENCE=high
```

The directory basename **and** the git remote name agree on `fvh-cost-attribution` (dash); the taskwarrior project holding 10 pending tasks is `fvh.cost-attribution` (dot). It is not an ancestor of the cwd, so `PROJECT_AMBIGUOUS` could not fire; it is not a prefix of the detected slug, so #2323's check did not apply. The guards behaved exactly as designed — the low-confidence zero was correct — but recovery still cost a hand-rolled `task export | jq 'group_by(.project)'` sweep and a guess about which of 40+ slugs was the right one, when the information needed was already in hand.

## Change

Extends the **existing** `PROJECT_AMBIGUOUS` mechanism rather than bolting on a second pass. On the fallback path, after the #2304 ancestor probe finds nothing, the collector folds `.`, `-`, `_` and case over the project slugs already present in the **one** all-projects snapshot and names any *different real* slug that collides:

```
PROJECT_AMBIGUOUS=fvh.cost-attribution
PROJECT_AMBIGUOUS_TASKS=10
PROJECT_AMBIGUOUS_REASON=separator-variant
```

Verified end to end against the issue's exact shape (a repo with `origin` also named `fvh-cost-attribution`):

```
=== TASKWARRIOR ===
TASK_AVAILABLE=true
OPEN_TASKS=0
TASK_SCOPE=all-projects-fallback
PROJECT_CONFIDENCE=low
PROJECT_REMOTE_NAME=fvh-cost-attribution
PROJECT_AMBIGUOUS=fvh.cost-attribution
PROJECT_AMBIGUOUS_TASKS=1
PROJECT_AMBIGUOUS_REASON=separator-variant
```

Design notes worth reviewing:

- **The fold REPLACES, it does not DELETE.** Each separator becomes one canonical character, so `ab` and `a-b` stay distinct and only a genuine separator/case substitution collides. A deletion-style fold would collide `fvhcostattribution` with `fvh-cost-attribution`; AL3 pins that case explicitly.
- **`PROJECT_AMBIGUOUS_REASON` is a new field** — the existing mechanism carried no reason. It is emitted for both rungs (`ancestor` for #2304, `separator-variant` for this one), because the consumer's phrasing differs ("the workspace above holds N" vs "a slug spelled with other separators holds N") and the slug alone cannot tell them apart.
- **The ancestor keeps precedence.** A containing workspace is a stronger claim than a name collision, so when both exist the ancestor is named and #2304's behaviour is byte-for-byte unchanged.
- **Nothing is adopted or rewritten**, and no second `task ... export` is issued — candidates are read out of the snapshot already in hand (`parallel-safe-queries.md`). Candidate slugs are store content, so one carrying a control character is skipped rather than allowed to break a `KEY=VALUE` row.
- **`--project` / `.claude/session.json` behaviour is unchanged**: an asserted slug keeps its scope, its count and its `high` confidence; only the additive advisory key appears, exactly as the #2304 ancestor signal already does there.

The `elif` chain in the fallback path was restructured into a nested `if` so the separator lookup can run in *both* the asserted and non-asserted cases; the resulting scope/confidence degradation is logically identical to the previous chain.

## Tests

TDD: TEST AL was written first and confirmed failing (14 assertions) against the unmodified collector, then passing after the fix. Every case EXECUTES the collector against stub binaries — no greps of the source.

Guard integrity is weighted equally with the positive case, since "always name the nearest slug" would satisfy AL1 while destroying every correct per-repo scope in the corpus:

| Case | Pins |
|---|---|
| AL1 | The repro: `PROJECT_AMBIGUOUS` / `_TASKS=10` / `_REASON=separator-variant`, slug never rewritten, variant's tasks never scoped in, exactly one of each row |
| AL2 | **Guard integrity** — a slug that DOES match raises no ambiguity at all, keeps `TASK_SCOPE=project` and `high` |
| AL3 | **No false positives** — `fvhcostattribution` (separators removed) and `fvh-cost-attribution-legacy` are NOT variants |
| AL4 | Case and `_` differences do collide (`FVH.Cost_Attribution`) |
| AL5 | Ancestor precedence + `PROJECT_AMBIGUOUS_REASON=ancestor` |
| AL6 | An asserted `--project` keeps scope/count/confidence and is only additively annotated |
| AL7 | The signal reaches the `--summary` hook path |
| AL8 | Still exactly one `task ... export`, still no `project:` filter |
| AL9 | Degradation: empty store, no `jq`, no `task` binary — nothing claimed |

```
$ bash session-plugin/scripts/tests/test-session-survey.sh
---
PASS=365 FAIL=0          # 351 pre-existing + 14 new, 0 regressions
```

Sibling suites, unchanged and green:

```
$ bash session-plugin/scripts/tests/test-distill-survey.sh      → PASS=36 FAIL=0
$ bash session-plugin/hooks/test-session-spinup-nudge.sh        → 18 passed, 0 failed
$ bash session-plugin/hooks/test-session-end-nudge.sh           → 23 passed, 0 failed
$ bash scripts/lint-shell-scripts.sh                            → 0 errors (9 warnings, all pre-existing in macos-plugin)
```

## CI: the one red check reproduces on `main`

`compliance`, `validate`, `gate` and `conventional-commits` are green. **`test` is red, and not because of this diff** — the runner reports `PASSED=135 / FAILED=1`, and the single failure is `scripts/tests/test-check-docs-index.sh`, which this PR does not touch. This branch's own suite passes in that same run (`PASS=./session-plugin/scripts/tests/test-session-survey.sh`).

Reproduced on a clean `origin/main` checkout:

```
$ bash scripts/check-docs-index.sh
STATUS=WARN
ISSUE_COUNT=2
  - SEVERITY=WARN TYPE=doc_count_drift MSG=README.md:103 states software-design-plugin has 5 skills but 6 exist on disk
  - SEVERITY=WARN TYPE=doc_count_drift MSG=PLUGIN-MAP.md:50 states software-design-plugin has 5 skills but 6 exist on disk
```

Cause: #2368 (`feat(software-design): promote prefer-diy-over-heavy-dependency to a skill`) added a sixth skill without updating the two hand-maintained count tables. `git diff --name-only origin/main` on this branch lists only the two `session-plugin/scripts/` files, so nothing here can affect it. Left alone deliberately — it wants a one-line fix in each doc, on its own PR (or `/docs-refresh`), not a drive-by in a session-plugin change.

## Follow-up left out of scope

The consumer docs describe `PROJECT_AMBIGUOUS` as "the ancestor slug" (`session-spinup/REFERENCE.md`, `session-spinup/SKILL.md`, `session-end/SKILL.md`, `session-plugin/README.md`) and none mention `PROJECT_AMBIGUOUS_REASON`. The new key is additive so those consumers keep working unchanged, but per `.claude/rules/docs-currency.md` they should be widened to "a slug that actually holds the work (ancestor or separator variant)" and taught to read the reason. Deliberately not touched here — this PR was scoped to the collector and its test — and worth a small follow-up.

Fixes #2355

---
_Generated by [Claude Code](https://claude.ai/code/session_01XauNJSTERATYqQsAxdEi1b)_

### DIFF
--- a/session-plugin/scripts/session-survey.sh
+++ b/session-plugin/scripts/session-survey.sh
@@ -409,6 +409,10 @@ project_confidence="high"
 project_resolved=""
 project_ambiguous=""
 project_ambiguous_tasks=0
+# Which rung named it: `ancestor` (#2304) or `separator-variant` (#2355). The
+# consumer's phrasing differs — "the workspace above holds N" vs "a slug spelled
+# with other separators holds N" — so the slug alone is not enough.
+project_ambiguous_reason=""
 # The prefix-sibling split (#2323). Integers so consumers can do arithmetic on
 # them unconditionally, exactly like OPEN_TASKS — the "was this even queried"
 # question is answered by TASK_SCOPE (`none`/`unknown`), not by a sentinel here.
@@ -489,21 +493,66 @@ else
         project_tasks_json="$anc_json"
         open_tasks="$anc_n"
         project_confidence="low"
-      elif [ "${anc_n:-0}" -gt 0 ]; then
-        # (3) A zero we are not allowed to adopt away must never look
-        # authoritative: name the ancestor slug that does hold the work, so
-        # session-end can say "0 here, N under <slug>" instead of skipping.
-        project_ambiguous="$anc_slug"
-        project_ambiguous_tasks="$anc_n"
+      else
+        if [ "${anc_n:-0}" -gt 0 ]; then
+          # (3) A zero we are not allowed to adopt away must never look
+          # authoritative: name the ancestor slug that does hold the work, so
+          # session-end can say "0 here, N under <slug>" instead of skipping.
+          project_ambiguous="$anc_slug"
+          project_ambiguous_tasks="$anc_n"
+          project_ambiguous_reason="ancestor"
+        else
+          # (3b) #2355: a REAL slug differing from the detected one only by
+          # SEPARATOR CHARACTER (or case). The reported shape: the repo is
+          # `fvh-cost-attribution` (basename AND git remote name agree) while
+          # the taskwarrior project holding the backlog is
+          # `fvh.cost-attribution`. It is no ancestor of the cwd, so (3) cannot
+          # see it, and it is not a prefix of the detected slug, so the
+          # prefix-sibling split below cannot either — yet the two names differ
+          # by one character class, which is enough to say so.
+          #
+          # The fold REPLACES each separator with one canonical character
+          # rather than deleting it, so `ab` and `a-b` stay distinct. Ranked
+          # only behind the ancestor: a containing workspace is a stronger
+          # claim than a name collision. Candidates come from the ONE snapshot
+          # already in hand — no second `task ... export`
+          # (parallel-safe-queries.md).
+          sep_slug=""
+          sep_n=0
+          if [ -n "$project" ]; then
+            while IFS= read -r sep_cand; do
+              [ -n "$sep_cand" ] || continue
+              # A slug is store content; a control character in one would break
+              # a KEY=VALUE row. Real taskwarrior slugs never contain one.
+              case "$sep_cand" in *[[:cntrl:]]*) continue ;; esac
+              cand_json=$(scope_of "$sep_cand")
+              [ -n "$cand_json" ] || cand_json="[]"
+              cand_n=$(printf '%s' "$cand_json" | jq 'length' 2>/dev/null || echo 0)
+              [ -n "$cand_n" ] || cand_n=0
+              if [ "$cand_n" -gt 0 ] 2>/dev/null; then
+                sep_slug="$sep_cand"
+                sep_n="$cand_n"
+                break
+              fi
+            done < <(printf '%s' "$all_tasks_json" | jq -r --arg p "$project" '
+              def norm: ascii_downcase | gsub("[._-]"; "-");
+              ($p | norm) as $np
+              | [ .[] | (.project // "")
+                  | select(. != "" and . != $p and (norm == $np)) ]
+              | unique | .[]' 2>/dev/null)
+          fi
+          if [ "${sep_n:-0}" -gt 0 ]; then
+            project_ambiguous="$sep_slug"
+            project_ambiguous_tasks="$sep_n"
+            project_ambiguous_reason="separator-variant"
+          fi
+        fi
         if [ "$asserted" = false ]; then
+          # (4) #2232: no single slug wins (portfolio checkout, renamed dir, …).
+          # Widen, and make the widening visible rather than reporting a clean 0.
           task_scope="all-projects-fallback"
           project_confidence="low"
         fi
-      elif [ "$asserted" = false ]; then
-        # (4) #2232: no single slug wins (portfolio checkout, renamed dir, …).
-        # Widen, and make the widening visible rather than reporting a clean 0.
-        task_scope="all-projects-fallback"
-        project_confidence="low"
       fi
     fi
   fi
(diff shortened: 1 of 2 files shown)


laurigates:
CI's `test` job failed on `scripts/tests/test-check-docs-index.sh` ("real repo is clean" — `STATUS`/`ISSUE_COUNT` assertions). I verified this is **pre-existing on `main`**, unrelated to this PR's diff: I ran the same test script both against this branch and against a clean checkout of `origin/main` (commit `fbeec576`) and got an identical `FAILED=2` in both — `session-survey.sh` and its test are the only files this PR touches, and this test doesn't cover either. Filing a separate issue to track the repo-wide docs-index drift itself; leaving it here per this run's guardrails (no unrelated scope creep in this PR).

---

laurigates:
Correction/update: rather than filing a separate issue, I fixed the root cause directly since it was a trivial 2-line count correction — see #2379. This PR's `test` check will go green once #2379 merges and this branch is updated against `main` (or once GitHub re-runs checks after a rebase); no further action needed on this PR for that specific failure.

---

PR #2414, opened 2026-08-16
docs(tools-plugin): warn that bare hf download fetches the whole repo

### LINKED ISSUE #2409: tools-plugin:hf-downloads — bare `hf download <repo>` pulls the entire repo; skill never says so
## What

`tools-plugin:hf-downloads` never states what `hf download <repo>` does when the
`FILENAMES` argument is omitted: it downloads the **entire repo**, not a listing
and not a no-op.

## Why

Every example in SKILL.md passes a file:

```
hf download <repo> <file> --local-dir /big/disk/staging
HF_HOME=/big/disk/hf-cache hf download <repo> <path/in/repo> --local-dir /big/disk/staging
```

So a reader following the skill is fine. The gap is that nothing marks the
filename as **load-bearing**, and the skill's whole premise is avoiding
surprise multi-GB writes to a disk that cannot take them. Omitting it is
exactly the mistake the skill exists to prevent, and it is silent — no prompt,
no size estimate, no confirmation.

Hit live on 2026-08-15 pulling four ~26–604 MB projection matrices from
`NicoLab28/ClipProj-MiniMax-H3`: the bare form fetched **3.2 GB** (the repo also
carries a `demo/` folder of reference videos and an `obsolete/` directory)
onto a disk at 78% use. Harmless there; on the ENOSPC-prone setups this skill
targets it would not have been.

The adjacent trap: reaching for `hf download` as a way to *list* a repo's files
is what produced the mistake. There is no listing subcommand in the examples, so
the natural-but-wrong move is to drop the filename and see what happens.

## How

Two small additions to `skills/hf-downloads/SKILL.md`:

1. A line where the download form is introduced — the `FILENAMES` argument is
   required in practice; omitting it fetches the whole repo, including any
   `demo/`, `examples/`, or `obsolete/` directories.
2. A listing recipe, so there is a documented right answer for "what's in this
   repo?" — either `hf download <repo> --include '<glob>'` for a filtered pull,
   or a `HfApi().list_repo_files(...)` snippet for a pure listing.

The `--include` / `--exclude` glob options already exist in `hf download --help`
but appear nowhere in the skill; they are the correct tool for "several files,
not all of them" and would have avoided this entirely.

laurigates:
## Summary

- `tools-plugin:hf-downloads` never stated what a bare `hf download <repo>` (no file argument) does: it silently fetches the **entire repo** — including any `demo/`, `examples/`, or `obsolete/` directories — with no listing step, size estimate, or confirmation. Every example in the skill happened to pass a file, so a reader following the skill was fine, but the skill's whole premise is avoiding surprise multi-GB writes and this is exactly that mistake with no guardrail.
- Hit live 2026-08-15: a bare-form pull grabbed 3.2 GB pulling four projection matrices, because the repo also carried a `demo/` folder and an `obsolete/` directory.
- Added a warning at the point the download form is introduced (the `<file>` argument is load-bearing, not optional), plus a "Listing a Repo's Files" section offering `HfApi().list_repo_files()` for a pure listing and `--include`/`--exclude` globs for a filtered pull — so dropping the file argument is never the natural move for "what's in this repo?".
- Added a `plugin-compliance-check.sh` regression guard (`check_skill_body()`) asserting the warning and the listing recipe survive future bulk edits.

## Test plan

- [x] Mutation-verified the new compliance guard: stripping the warning token makes it fire `❌ ... must retain token 'required in practice'`; restoring the file clears it.
- [x] `./scripts/plugin-compliance-check.sh tools-plugin` — Body column ✅, no new errors (pre-existing WARNs on description length / When-to-Use / size are unrelated).
- [x] `bash scripts/tests/test-plugin-compliance-check.sh` — 21/21 pass.
- [x] `bash scripts/lint-shell-scripts.sh scripts/plugin-compliance-check.sh` — 0 errors, 0 warnings.

Fixes #2409

---
_Generated by [Claude Code](https://claude.ai/code/session_01QQnWDk8E73mypNc1yY2L1W)_

### DIFF
--- a/scripts/plugin-compliance-check.sh
+++ b/scripts/plugin-compliance-check.sh
@@ -843,6 +843,23 @@ check_skill_body() {
       done
     fi
 
+    # Regression (#2409): hf-downloads must document that the FILENAMES
+    # argument is load-bearing — a bare `hf download <repo>` with no file
+    # silently fetches the whole repo (demo/, examples/, obsolete/ included,
+    # no confirmation), which is exactly the surprise-disk-write mistake the
+    # skill exists to prevent. It must also offer a listing recipe so
+    # dropping the file argument is never the natural move for "what's in
+    # this repo?". A bulk edit that trims the warning or the listing recipe
+    # would silently restore the trap, so assert both survive.
+    if [ "$skill_name" = "hf-downloads" ]; then
+      for token in "required in practice" "list_repo_files"; do
+        if ! grep -qF "$token" "$skill_file"; then
+          issues+=("❌ ${plugin}/${skill_name}: SKILL.md must retain token '${token}' (bare 'hf download <repo>' fetches the whole repo — see #2409)")
+          has_errors=true
+        fi
+      done
+    fi
+
     # Regression: github-actions-auth-security must document the GitHub Actions
     # script-injection mitigation (distinct from Claude *prompt* injection):
     # untrusted run-context values bound to an intermediate `env:` variable and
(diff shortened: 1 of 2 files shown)


laurigates:
CI green (gate, conventional-commits, validate, test, compliance, lint, split all passed; `doc-audit` skipped as expected), no merge conflicts, no unresolved review comments. Ready for review/merge.

---

Issue #1868, opened 2026-06-28
Workflow resumeFromRunId re-runs already-succeeded worktree agents → duplicate PRs


laurigates:
## What

`Workflow({resumeFromRunId})` re-executes `isolation: "worktree"` agents that had **already succeeded** in the prior run, instead of returning their cached results. Resuming a workflow to recover a few rate-limited worktree agents therefore re-runs the successful ones too.

## Observed (2026-06-28, issue-sweep workflow)

- A waved pipeline opened PRs from worktree agents; the final wave's 3 impl agents hit the burst rate limit and errored.
- Resuming with `{scriptPath, resumeFromRunId}` (same script, unchanged) to retry those 3 **re-ran `impl:#1806`**, which had already opened PR #1857 — producing a **duplicate PR #1858** for the same issue.
- The 3 originally-failed agents still rate-limited on resume.

Non-worktree `agent()` calls (the scope stage) cached correctly on the same resume; only the `isolation: "worktree"` agents re-ran.

## Why it matters

The documented resume contract is "completed `agent()` calls with unchanged `(prompt, opts)` return cached results instantly." Worktree agents appear to violate that, and the side effect is **outward** (a real PR on the remote), so it's not silently absorbed — it creates duplicate PRs / branches that need manual cleanup.

## Workaround (adopted)

Recover failed worktree agents with a **fresh sequential dispatch** (one Opus agent doing the remainder one-at-a-time), not a whole-workflow resume — which also dodges the burst limit. Distilled into `.claude/rules/agent-coworker-detection.md` (PR #1867).

## Possible fixes

- Make `resumeFromRunId` treat a succeeded worktree agent's recorded result (incl. its opened-PR URL) as cacheable, same as non-worktree agents; or
- At minimum, document that worktree-isolation agents are not resume-cacheable so callers don't resume-to-retry them.

## Related

- Task tracking note: Workflow `resumeFromRunId` also pins the original run's `args` (separate prior finding).
- #1857 / #1858 — the duplicate-PR instance.

PR #1858, opened 2026-06-28
feat(comfyui-plugin): standalone-modal skeleton for no-widgets packs

laurigates:
## What

`comfyui-node-scaffold`'s `scaffold.py` emitted the per-widget intercept stub
(`TARGET_WIDGETS` / `openPicker` / `enhanceNode` / `widget.onPointerDown`) for
**every** modal variant — including a `--variant backend` pack with no
`--widgets` whose UI is a standalone modal opened from a toolbar button /
command (a node manager, dashboard, gallery-actions panel). That widget framing
is wrong for such a pack and gets re-authored from scratch, breaking the
scaffold's "fill-in-a-stub" promise (observed building `comfyui-touch-manager`).

## How (no-widgets heuristic — the issue's primary suggestion)

- `scaffold.py` computes `standalone = modal and not widgets` and emits a new
  `INDEX_TS_STANDALONE`: a `registerExtension` skeleton with a
  `commands`/`menuCommands` entry whose command opens an exported `openShell()`
  over `openModalShell(...)` — **no** `TARGET_WIDGETS` / `openPicker` /
  `enhanceNode` / `onPointerDown`.
- Ships a jsdom modal-mount smoke test (`JS_TEST_STANDALONE`) that asserts
  `modal.bodyEl` is populated — the empty-modal trap is exactly where standalone
  bugs hide (the issue author had to hand-add this after the build).
- The vein-conditional CLAUDE/README fragments (`WHAT_DESC`, `EXT_ROW_DESC`,
  `HOOK_RULE`, `VEIN`) gain a standalone branch. `DEPENDENCIES_BLOCK` /
  `KIT_RULE` stay on the `modal` flag — standalone still depends on the modal
  kit (`modal == True`).
- SKILL.md variant table documents that omitting `--widgets` on a modal variant
  yields the standalone-modal skeleton; passing `--widgets` keeps the
  widget-intercept stub.

The heuristic covers both `backend` and `frontend` variants scaffolded without
`--widgets` ("no widget to hook ⇒ the widget vein is provably inapplicable").
The `commands` / `menuCommands` / `ComfyCommand` shapes were verified against
`@comfyorg/comfyui-frontend-types` so the generated `index.ts` typechecks.

## Design decision to confirm

The issue offered two routes: **(a)** the no-widgets heuristic
(`backend && no --widgets ⇒ standalone`), or **(b)** an explicit
`--ui standalone|widget` flag / 4th variant. This PR implements **(a)** per the
[...]

PR #2450, opened 2026-08-19
feat: absorb nine always-loaded user rules as skills and two hooks

laurigates:
Marketplace half of [laurigates/dotfiles#353](https://github.com/laurigates/dotfiles/issues/353). Paired with laurigates/dotfiles#378, which trims the rules these skills replace — **merge this one first**, since those stubs point here.

## What lands

**Seven new skills**, each the verbatim body of a rule that previously loaded into every session of every project:

| Skill | Promoted from | Trigger |
|---|---|---|
| `git-plugin:git-merge-hazards` | `pr-merge-hazards.md` | merging a PR, a stacked chain, or over red CI |
| `git-plugin:git-issue-scoping` | `read-issue-thread-before-contributing.md` | scoping a contribution from a GitHub issue |
| `git-plugin:git-upstream-fix-check` | `verify-upstream-before-patching.md` | patching vendored / tarball-installed / forked code |
| `git-plugin:git-repo-delete-check` | `repo-deletion-safety.md` | about to `rm -rf` a git checkout |
| `documentation-plugin:docs-verify-machine-facts` | `verify-machine-facts-before-publishing.md` | org docs quoting values read off your own host |
| `documentation-plugin:docs-single-source` | `documentation-authoring.md` | writing or restructuring docs |
| `documentation-plugin:docs-fetch-fallbacks` | `tool-use-patterns.md` (WebFetch section) | a doc or GitHub URL fetch failed |
| `code-quality-plugin:code-scaffold-backport` | `scaffold-fix-backport.md` | fixing something a generator produced |

**Two merges into existing skills** — the rule was already largely covered, so nothing new was created:

- `taskwarrior-plugin:task-add` / `task-claim` ← the `project:` prefix-match trap, the no-rc bootstrap, read-annotations-first
- `agent-patterns-plugin:parallel-agent-dispatch` ← the fan-out burst-limit and usage-limit recovery sections

**Two PreToolUse hooks** (`hooks-plugin`), turning prose into deterministic enforcement:

| Hook | Tier | Guards |
|---|---|---|
| `repo-deletion-safety.sh` | **block** (exit 2) | `rm -rf` of a checkout whose history exists nowhere else; self-extinguishing once you push or write the tar backup |
| `branch-base-guard.sh` | **nudge** (`ask`) | cutting a branch from a local default branch that is ahead of its remote (`git-hazards.md` trap #2) |

The block/nudge split is argued rather than assumed, per `.claude/rules/hook-block-vs-nudge.md`. Branch-base is a nudge because nothing it prevents is irreversible, branch creation is high-frequency, and local-main-ahead is the *normal* steady state in main-branch-dev repos — a hard block there would be permanently disabled.

## Why "cannot have drifted" is checkable, not hoped

Every body was moved by script, never retyped, then diffed against a pre-change snapshot. The only edits inside a moved body are sibling-rule path requalifications (`git-hazards.md` → `~/.claude/rules/git-hazards.md`), since the text no longer sits beside those rules. Each is listed in its commit.

One paragraph was **deliberately not** copied verbatim: the resident `tool-use-patterns.md` claimed `resumeFromRunId` replays completed agents from cache at zero cost, which contradicts `agent-coworker-detection.md` — a succeeded `isolation: "worktree"` agent is re-executed on resume and re-fires its side effects (duplicate PR #1858 of #1857, issue #1868). Promoting it verbatim would have carried a live bug into the skill.

## Regression the promotion would otherwise have caused

`git-plugin:git-issue` Step 2 fetched `--json title,body,state,assignees,labels` — body only, no comments. That is verbatim the trap `read-issue-thread-before-contributing` prevented on every turn *while it was resident*. Demoting it to a pull-loaded skill without fixing the highest-traffic issue entry point would have been a net regression, so `comments` is added and the step points at the scoping skill.

## Verification

| Check | Result |
|---|---|
| `check-docs-index.sh --strict` | `STATUS=OK ISSUE_COUNT=0` (was WARN/9) |
| `check-feature-flags-catalog.sh` | `STATUS=OK MISSING_COUNT=0` (was **ERROR/6** — the new hooks' flags were uncatalogued) |
| `check-plugin-readme-currency.sh` | `STATUS=OK ISSUE_COUNT=0` |
| `check-branch-containment-guidance.sh` | `STATUS=OK ISSUE_COUNT=0` |
| `test-repo-deletion-safety.sh` | 57 passed, 0 failed |
| `test-branch-base-guard.sh` | 59 passed, 0 failed |
| `lint-shell-scripts.sh` | 0 errors; 9 warnings, all pre-existing |
| `shellcheck` (both new hooks + tests) | clean, no suppressions |

Both hook suites include mutation checks — hardcoding `main` as the default branch fails 3 assertions, and disabling the suggested-fix exemption fails 11 — so the guards are provably load-bearing rather than vacuously green.

## Known-red, pre-existing, not from this change

`hooks-plugin/hooks/test-bash-antipatterns.sh` fails 19 assertions in this container. Cause: the hook probes `command -v sg` as an ast-grep fallback and finds `/usr/bin/sg`, which is shadow-utils' "run a command with a different group ID" — so structural detection silently no-ops and the blocks fail open. `bash-antipatterns.sh` and its test are byte-identical to `origin/main`. Worth a separate issue: the `sg` probe should verify the binary is actually ast-grep.

## Left for a follow-up

- `docs/diagrams/plugin-relationships.svg` not regenerated — `d2` is not installed here. The `.d2` source is correct; the committed `.svg` still renders the old counts.
- `documentation-plugin/.claude-plugin/plugin.json` still carries a `knowledge-graph` keyword for a skill that has no directory (the README row is removed in this PR).
- `repo-deletion-safety.sh`'s block message still inlines the three-option remediation that also lives in `git-repo-delete-check`. Trimming it collides with two things — option 2's tar command is what makes the block self-extinguishing, and the test suite pins all three options by literal — so it needs a hook+test change, not a text edit.

Refs laurigates/dotfiles#353

### DIFF
--- a/.claude/rules/agent-coworker-detection.md
+++ b/.claude/rules/agent-coworker-detection.md
@@ -239,7 +239,8 @@ agents that had already succeeded**, opening a **duplicate PR** (#1858 dup of
 worktree agents. Recover them with a **fresh, sequential dispatch** (one Opus
 agent doing the remainder one-at-a-time, or a small re-run waved ≤3) — which also
 dodges the burst rate-limit that caused the original failures
-(`~/.claude/rules/tool-use-patterns.md`). Before any recovery dispatch, check
+(`agent-patterns-plugin:parallel-agent-dispatch` § Concurrent Rate-Limit Risk).
+Before any recovery dispatch, check
 `gh pr list --search "issue-<N>"` so you don't open a duplicate.
 
 ## Limitations
--- a/.claude/rules/generated-fleet-drift.md
+++ b/.claude/rules/generated-fleet-drift.md
@@ -15,7 +15,7 @@ eventually audits the gap. The failure this rule prevents is the audit's *first
 move*: reading "differs from the template" as "wrong". It usually isn't, and
 acting on that reading is how an audit becomes a regression.
 
-Sibling of `~/.claude/rules/scaffold-fix-backport.md` (fix the class, not the
+Sibling of `code-quality-plugin:code-scaffold-backport` (fix the class, not the
 instance). This is its inverse: **don't push the class over an instance whose
 divergence was deliberate.**
 
@@ -99,7 +99,7 @@ sceptically; the script is the easy part.**
 
 ## Related
 
-- `~/.claude/rules/scaffold-fix-backport.md` — the inverse (fix the generator, not just the instance)
+- `code-quality-plugin:code-scaffold-backport` — the inverse (fix the generator, not just the instance)
 - `.claude/rules/drift-detection-triggering.md` — *triggering* a sweep; this rule is about classifying what it finds
 - `.claude/rules/offload-to-deterministic-substrate.md` — why the sweep is a script and the classification is not
 - `comfyui-plugin/skills/comfyui-node-scaffold/fleet-policy.toml` — the reference policy, with per-entry rationale
--- a/.claude/rules/gh-json-fields.md
+++ b/.claude/rules/gh-json-fields.md
@@ -94,11 +94,11 @@ wall-clock deadline.** Observed 2026-07-30: two such loops burned a 6m40s and a
 ### Gating on an expected check *count* is brittle
 
 The companion advice "wait for nothing-pending **and** at least N checks" (see
-`~/.claude/rules/pr-merge-hazards.md`) guards a real race — zero-pending is
-trivially true before jobs register. But a hardcoded `N` breaks on
-**path-filtered** workflows: the same repo yields 7 checks for a PR touching
-`*-plugin/**` and 5 for one touching only `.claude/rules/**`, so `-ge 6` waits
-forever on a perfectly green PR.
+`git-plugin:git-merge-hazards`, which carries the CI check-*registration*
+race) guards a real race — zero-pending is trivially true before jobs
+register. But a hardcoded `N` breaks on **path-filtered** workflows: the same
+repo yields 7 checks for a PR touching `*-plugin/**` and 5 for one touching
+only `.claude/rules/**`, so `-ge 6` waits forever on a perfectly green PR.
 
 Derive the expectation instead of hardcoding it — poll until the check count is
 **stable across two consecutive reads** with nothing pending, and bound the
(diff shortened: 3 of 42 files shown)


laurigates:
## CI green. One flake on the first run, root-caused — worth its own fix

`test` failed once (`PASS=404 FAIL=1`), passed on re-run. All checks now green. It is **not** this PR's, and it is not really a flake either — it's a latent bug in the test harness that this run happened to expose.

`session-plugin/scripts/tests/test-session-survey.sh` sets `set -uo pipefail` (line 8), and its `check()` helper is:

```bash
if printf '%s' "$haystack" | grep -qF "$needle"; then
```

`grep -q` exits the instant it matches, closing the pipe. If the haystack is large enough that `printf` is still writing, `printf` takes EPIPE and exits 141 — and `pipefail` promotes the whole pipeline to non-zero **even though the needle was found**. So a *successful* match is reported as a failure, and which assertion loses is a scheduling race. The CI log shows the mechanism directly:

```
test-session-survey.sh: line 17: printf: write error: Broken pipe
FAIL: AM11: session-end maps the 'unknown' reason
  expected to find: `unknown`
```

Not mine: `test-session-survey.sh`, `session-survey.sh`, and `session-end/SKILL.md` (the file AM11 asserts against) are all byte-identical to `origin/main` here. This PR's only session-plugin change is two reference repoints in `session-wrap/SKILL.md`. Passes 3/3 locally (405 assertions).

Fix, whenever someone touches that file — drop the pipe so there is nothing to break:

```bash
if [[ "$haystack" == *"$needle"* ]]; then
```

A pure-bash substring test, no subprocesses, no SIGPIPE, and faster across 405 assertions. (`grep -qF <<<"$haystack"` also works — a here-string is a temp file, not a pipe.) Not doing it here: it is unrelated to the rules triage and belongs in its own change.

For the record, the two hooks this PR adds both passed in CI on the first run:

```
PASS=./hooks-plugin/hooks/test-branch-base-guard.sh
PASS=./hooks-plugin/hooks/test-repo-deletion-safety.sh
```

`test-bash-antipatterns.sh` also passes in CI, which confirms the local failure noted in the PR description was purely the `/usr/bin/sg` name collision — CI installs `@ast-grep/cli` explicitly, so the structural blocks work there.

---

PR #2474, opened 2026-08-21
ci(workflows): workflow health check — mechanical fixes, label provisioning, standards conformance

laurigates:
## Summary

Health check and remediation of all 27 GitHub Actions workflows (3,972 lines). Built in stages — **Stages 1 and 4 have landed; 2 and 3 are queued.**

Full report: https://claude.ai/code/artifact/89bc2e8f-5ea6-42c0-8d6a-8e830ad68aa1

## Headline findings

Three problems, each verified against primary sources rather than inferred:

1. **The audit layer has been inert for five months.** `scheduled-audits.yml` (0/6 green since 2026-03-01), `workflow-model-audit.yml` (0/6) and `fleet-drift-audit.yml` (0/3) had never had a successful run. Root cause confirmed from job logs: seven GitHub labels were attached but never created, and `gh issue create --label "<unknown>"` exits 1 **on the last line of the job**, after all the analysis has already succeeded. **Fixed in this PR.**

2. **The declared Claude tool sandboxes do not bind.** Eight workflows pass a newline-separated tool list to `additional_permissions:`, which upstream `claude-code-action@v1` documents as *"Additional GitHub permissions to request (e.g. 'actions: read')"* — a GitHub permissions map, not a tool gate. Verified directly against the upstream `action.yml`. No `--allowedTools` boundary is applied anywhere in the repo. **Stage 2.**

3. **Nothing was bounded.** Zero of 27 workflows set `timeout-minutes` (360-minute default everywhere); 17 had no `concurrency` group. **Fixed in this PR.**

Alongside these, four paths reported success while doing nothing — including a required check whose auto-fix push could not succeed by construction, and a gate job that was *skipped, and therefore reported as passing*, precisely when the check it guards failed.

## Staged plan

| Stage | Scope | Status |
|---|---|---|
| 1 | Mechanical fixes — timeouts, concurrency, cron spread, silent-green repairs, dead code | ✅ Landed |
| 4 | Dead-audit remediation — guard + label provisioning | ✅ Landed |
| 2 | Standards + security — `additional_permissions` sweep, env-binding injection sites, data framing, CODEOWNERS | Queued |
| 3 | Consolidation — sync-workflow matrix collapse, `script-to-issue` composite, discovery-runner self-tests | Queued |

## What landed

**Stage 1 — 78 mechanical changes, 0 skipped.** Every audit premise was verified before acting on it and every one held. `timeout-minutes` on all 32 `runs-on` jobs; 14 new `concurrency` groups; all 11 cron schedules now distinct (seven previously sat at `0 9`, so on any 1st/8th/15th falling on a Monday four fired in the same minute, three of them opus jobs).

Silent-green repairs, each reproduced in a scratch shell before being fixed:
- `validate-plugin-configs.yml` — the auto-fix push could **never** succeed (bare `git push` from a detached-HEAD SHA checkout) *and* the retry loop swallowed the failure, so a required check reported green on out-of-sync configs. Now `git push origin "HEAD:$HEAD_REF"` with a `PUSHED` sentinel.
- `enforce-conventional-commits.yml` — the `gate` job had no `if: always()`, so it was **skipped, and reported as passing**, exactly when the check it guards failed. Now asserts `needs.*.result`.
- `release-please.yml` — same retry-loop shape, plus three `gh pr list` calls missing `--limit` (default 30, so the release PR silently falls off the page).
- `github-workflow-auto-fix.yml` — both infinite-loop guards were inert (`fix(scope):` matches neither `fix:` nor `startsWith('fix(auto):')`), which matters because it pushes with a PAT and PAT pushes *do* re-trigger `workflow_run`.

**Stage 4 — labels provisioned, three audits revived.** Six idempotent `gh label create --force` steps. `--force` is preferred over the older `2>/dev/null || true` form used elsewhere in the repo: that also swallows genuine auth failures, deferring the error to a much more expensive step. `ci-failure` is provisioned but `bug` deliberately is **not** — `--force` *updates* an existing label, and `bug` is a GitHub default whose colour and description shouldn't be overwritten.

**`scripts/check-workflow-labels.sh` + a 28-assertion regression test.** Two scoping decisions carry its signal-to-noise, both learned by getting them wrong first:
- Only `create` counts. `gh issue list --label <unknown>` returns empty and exits 0, so counting reads flags call sites that cannot fail — and an earlier revision also parsed the prose *"omit the `--label` flag"* as a label named `flag`.
- Backslash continuations are folded before matching, since every real call site puts `--label` several lines below the command.

Together those took the repo from 12 reported findings to exactly **7**, matching an independent label-API probe. The guard also caught a label the log-based diagnosis missed — `ci-failure`, masked because that prompt instructs Claude to drop the flag when the label is absent, so the workflow degraded silently instead of failing.

## Notes for review

- `timeout-minutes` is **not valid** on a job that calls a reusable workflow — GitHub rejects the file outright. The six reusable-caller jobs therefore get `concurrency` only. Verified exhaustively by parsing every job, not by grepping the diff.
- The `gate` fix is a real behaviour change: PRs whose titles were silently passing will start failing. That is the job's stated purpose, and release-please depends on it.
- **One commit is not this PR's work.** `docs: refresh plugin catalog counts` fixes count drift that arrived on `main` with the `tool-result-traps` skill. It reproduces against a pristine export of `origin/main` and passes on this branch in isolation, so it is a base-branch failure — but it blocks this PR from reaching green, so it is fixed here. Trivially separable.
- One scope addition beyond the agreed Stage 1 list: `release-please.yml`'s `continue-on-error: true` on the core release action, which let the entire release pipeline no-op with a green checkmark. Same defect class as the two retry loops; easy to drop if unwanted.
- Stage 3 will be constrained to changes that preserve every job id and check name, since branch-protection required contexts cannot be read from this environment. Folding `plugin-enablement-drift` into `compliance` is **dropped** for that reason — it would remove a check context.
- Three pre-existing concurrency inconsistencies were found and deliberately **not** fixed here, to keep this diff mechanical: `plugin-pr-checks` missing its `|| github.ref` fallback, and `validate-plugin-configs` / `skill-splitter` carrying `cancel-in-progress: true` while they commit and push. Deferred to Stage 2.

## Test plan

- [x] `scripts/tests/test-check-workflow-labels.sh` — 28/28, TEST A green
- [x] `scripts/check-workflow-labels.sh` — `UNPROVISIONED_COUNT=0`
- [x] Every changed workflow parses (`yaml.safe_load`, all 27)
- [x] No `timeout-minutes` on any reusable-workflow call (verified by job-level parse)
- [x] `scripts/check-workflow-model.sh` and `scripts/check-workflow-script-triggers.sh --strict` both exit 0
- [x] `scripts/check-docs-index.sh --strict` exits 0
- [x] No job id, job `name:` or workflow `name:` changed (programmatic base-vs-head comparison)
- [x] All 11 cron expressions distinct
- [x] CI green on `726f952`

https://claude.ai/code/session_01CKuntEiJoE9kSTap3jZrLX

---
_Generated by [Claude Code](https://claude.ai/code/session_01CKuntEiJoE9kSTap3jZrLX)_


laurigates:
CI is red on `test` with **two** failures (137/139 passing). Neither is a surprise, but one of them is not this PR's:

### 1. `test-check-docs-index.sh` — pre-existing on `main`, not caused by this PR

Verified by running the checker against a pristine export of `origin/main` (083b224):

```
STATUS=WARN  ISSUE_COUNT=3
  README.md:72 states agent-patterns-plugin has 20 skills but 21 exist on disk
  PLUGIN-MAP.md:28 states agent-patterns-plugin has 20 skills but 21 exist on disk
  plugin-relationships.d2 states agent-patterns has 20 skills but 21 exist on disk
```

`main` landed `agent-patterns-plugin/skills/tool-result-traps/SKILL.md` without refreshing the catalog counts. It passes on this branch in isolation (22/22) because the branch predates that skill; it fails on the merge commit CI actually builds.

Normally a base-branch failure is the one legitimate "not mine" and I'd leave it. I'm going to fix it here anyway, because it blocks this PR from *ever* going green and there's no indication anyone is watching `main`. It'll land as its own commit, using the repo's own `docs-refresh` path rather than hand-editing the three numbers. Happy to split it out if you'd rather it went in separately.

### 2. `test-check-workflow-labels.sh` — the intended ratchet

`TEST A` asserts no workflow attaches an unprovisioned label. That is currently false — seven labels are attached but were never created:

`blueprint-health`, `infra-compliance`, `docs-index`, `agentic-audit`, `fleet-drift`, `workflow-model-audit`, `ci-failure`

This is the root cause of the three never-green scheduled audits, so the test is meant to be red until the provisioning steps land. Fixture cases B–J all pass, so the guard is proven to parse rather than trivially return OK.

### Next push

1. Merge `main` (needed for the docs fix, and this branch is behind)
2. Catalog count refresh → clears failure 1
3. Label provisioning in the four affected workflows → clears failure 2
4. Remainder of the Stage 1 mechanical fixes + the verification pass

---

PR #2486, opened 2026-08-22
feat(code-quality-plugin): cover term-set derivation and exclusion justification in bulk-sweep-classify

### LINKED ISSUE #2479: feat(code-quality-plugin): bulk-sweep-classify should cover deriving the term set and justifying scope exclusions
## What

`bulk-sweep-classify` starts at "you have a pattern — classify its matches." Two failures from a real sweep sit *outside* that frame: one before Step 1, one inside Step 5. Both produced a clean-looking result over an incomplete scope.

Filing as an enhancement to that skill rather than a new one — it already owns sweep methodology.

## Gap 1 — derive the term set from what the mechanism *claims*, not just what it is *named*

Removing a mechanism means removing its **assertions** as well as its **identifiers**, and those share no tokens.

Concrete case (`ForumViriumHelsinki/podio-mcp`, PRs #160 / #163): a build-time credential injector was removed. The sweep terms were the identifiers — `BUILD_DEFAULTS`, `inject-build-defaults`, `postbuild`. That correctly found the script, the npm hook, the constants module and the workflow passthrough, and the PR shipped.

It missed this, in a file the sweep had already edited:

```
src/index.ts:134
 * - PODIO_CLIENT_ID: Your Podio app's client ID (baked into published package)
```

The comment asserts exactly what the mechanism was supposed to do, and contains none of the three identifiers. It survived, and typedoc renders it into the API reference — so the claim outlived the thing that made it true. Caught only incidentally, one PR later.

The generalization: after listing the identifier terms, ask **"what did this mechanism promise, and in whose words?"** Add those. For a removal, terms usually come in two families:

| Family | Example terms | Where they live |
|---|---|---|
| Identifiers | function, file, script, env var, config key | code, config, workflows |
| Claims | "baked in", "injected at build time", "automatically", "the published package …" | comments, README, docstrings, error and warning strings |

The claim family is where user-facing damage concentrates: an identifier left behind is dead code, but a claim left behind is documentation that is now false.

## Gap 2 — an exclusion in the verification scope needs its own justification verified

Step 5 already says verify against the preserved set rather than literal zero. This is the adjacent trap: the sweep's *scope* carried an exclusion, and the exclusion's rationale was wrong, so a clean result was clean over the wrong tree.

Same case. The final sweep excluded `docs/api/` — committed typedoc output — reasoning that CI regenerates and publishes it, so the committed copy did not matter. Both halves were false: GitHub Pages was never enabled on the repo (`/repos/.../pages` → 404) and the deploy workflow had failed every run for ten days. So the excluded directory was the *only* rendered copy, and it still carried the stale claim.

The report said "no matches remain," which was true of what it searched and false of the repo.

The generalization: **every exclusion in a verification scope is a claim about the world, and needs the same treatment as a negative result.** "Generated, so it doesn't count" assumes the generator runs — check that it does. Cheap test: for each excluded path, state why in one line, then verify that line. It pairs naturally with the existing rule that a zero-match verdict must be control-tested.

## Suggested shape

- A short subsection under Step 1 (or a Step 0.5) on deriving the term set, with the identifiers-vs-claims table
- Two or three sentences in **The Verification Trap** on justifying exclusions, since that section already exists and this is the same failure at a different level

Happy to draft the SKILL.md edit if the framing looks right.

laurigates:
## What

`bulk-sweep-classify` starts at "you have a pattern — classify its matches." Two failures from a real sweep sit *outside* that frame: one before Step 1, one inside Step 5. Both produced a clean-looking result over an incomplete scope.

Fixes #2479.

## Gap 1 — derive the term set from what the mechanism *claims*, not just what it is *named*

Removing a mechanism means removing its **assertions** as well as its **identifiers**, and those share no tokens.

In `ForumViriumHelsinki/podio-mcp` (#160 / #163) a build-time credential injector was removed. The sweep terms were the identifiers — `BUILD_DEFAULTS`, `inject-build-defaults`, `postbuild` — which correctly found the script, the npm hook, the constants module and the workflow passthrough. It missed this, in a file the sweep had **already edited**:

```
src/index.ts:134
 * - PODIO_CLIENT_ID: Your Podio app's client ID (baked into published package)
```

The comment asserts exactly what the mechanism was supposed to do and contains none of the three identifiers. typedoc renders it into the API reference, so the claim outlived the thing that made it true. Caught only incidentally, one PR later.

A new subsection under Step 1 adds the derivation question — *"what did this mechanism promise, and in whose words?"* — plus the identifiers-vs-claims table, and the consequence that decides severity: an identifier left behind is dead code, but **a claim left behind is documentation that is now false**.

## Gap 2 — an exclusion in the verification scope needs its own justification verified

Step 5 already says verify against the preserved set rather than literal zero. This is the adjacent trap: the sweep's *scope* carried an exclusion, and the exclusion's rationale was wrong, so a clean result was clean over the wrong tree.

Same case. The final sweep excluded `docs/api/` — committed typedoc output — reasoning that CI regenerates and publishes it, so the committed copy did not matter. Both halves were false: GitHub Pages was never enabled (`/repos/.../pages` → 404) and the deploy workflow had failed every run for ten days. The excluded directory was the *only* rendered copy, and it still carried the stale claim.

> The report said "no matches remain," which was true of what it searched and false of the repo.

**The Verification Trap** gains the rule: every exclusion in a verification scope is a claim about the world and needs the same treatment as a negative result — state why in one line, then verify that line. It pairs with the existing rule that a zero-match verdict must be control-tested.

## Regression check

Per `.claude/rules/regression-testing.md`, a `check_skill_body()` block in `scripts/plugin-compliance-check.sh` scoped to `code-quality-plugin/bulk-sweep-classify`, pinning one token per addition:

| Token | Marks |
|---|---|
| `what did this mechanism promise, and in whose words?` | the term-derivation subsection |
| `a claim left behind is documentation that is now false` | the identifiers-vs-claims distinction |
| `every exclusion in a verification scope is a claim about the world` | the exclusion-justification rule |

All three mutation-verified individually: each replaced in isolation produces a ❌ naming that token; the restored file reports `Body ✅`.

**The guard caught a real defect in this PR's own first draft.** `grep -qF` is line-scoped, and two of the three phrases had been split across an ~80-col wrap, so the draft failed its own check — the same trap #2174 records for `WIP at checkpoints`. Fixed by reflowing each pinned phrase onto one line.

The Known Regressions row lands in this PR (`.claude/rules/regression-testing.md` carries `merge=union`, so parallel PRs each append without conflicting).

## Test plan

- [x] `just lint-compliance code-quality-plugin` — exit 0, `Body ✅`; only pre-existing size/description ⚠️
- [x] `python3 scripts/audit-skill-descriptions.py --strict-all` — exit 0
- [x] `scripts/check-unrendered-templates.sh` — `UNRENDERED_CONDITIONALS=0 STATUS=OK`
- [x] `scripts/check-unused-bash-grant.sh --strict` — `SKILLS_SCANNED=419 ISSUE_COUNT=0`
- [x] `scripts/lint-context-commands.sh` — clean
- [x] `scripts/check-context-command-execution.sh --strict` — `PASS=628 FAIL=0`
- [x] `scripts/run-skill-script-tests.sh` — **FAILED=0**
- [x] `plugin-compliance-check.sh` self-tests (21/0, 8/0, 3/0) and `lint-shell-scripts.sh` — run because this PR edits that script
- [x] `pre-commit run --files <changed>` — all hooks passed

SKILL.md 11,901 → 14,516 decoded UTF-8 chars — same WARN band, no `references/` split needed. The size and 236-char description WARNs on this skill are pre-existing on `main`.

`run-skill-script-tests.sh` exits 1 on 5 `required_test_skipped` ERRORs for tooling absent from this sandbox (`ast-grep` ×2, `cargo-generate`, `task` CLI ×2) — environmental and pre-existing.

## Deliberately unchanged

`code-quality-plugin/README.md`. Its catalog row already reads "the four-category classify-then-transform discipline with allowlist-aware verification", which still describes the skill accurately — these additions extend Step 1 and an existing section rather than changing what the skill is *for*. Flagging it in case you'd rather it be rewritten.

Opened ready for review rather than draft, per the scheduled-maintenance routine that produced it.

https://claude.ai/code/session_01Lndu4u6Dc4sJ9Z6jKMdHZu

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lndu4u6Dc4sJ9Z6jKMdHZu)_


laurigates:

laurigates:

Issue #2483, opened 2026-08-21
feedback(scripts): widen check-delegation-reachability.sh beyond git-plugin


laurigates:
## Follow-up from #2442

#2442 fixed the reported defect (`git-pr-watch` presenting the
`disable-model-invocation` sibling `git-pr-feedback` as an action the agent can
take) and shipped `scripts/check-delegation-reachability.sh` as the guard.

That guard's `DELEGATION_SCOPE_DEFAULT` is currently **git-plugin only** — the
five skills that reference a gated sibling from an action section. Its output
names the audit set explicitly (`SCOPE=`, `AUDITED=`, `SCOPE_IS_REPO_WIDE=false`)
so `STATUS=OK` cannot be misread as "the marketplace is clean".

## What is still deferred

A marketplace-wide sweep: every catalog-present skill in every plugin whose
prose delegates to a `disable-model-invocation: true` sibling **in its own
plugin**. The guard already resolves gated status from the sibling's own
frontmatter, so widening is a scope change plus whatever it surfaces — but it
needs a measured pass, because:

- `resolve_sibling()` re-globs the plugin's skills per reference, so a
  ~410-skill scan needs a cost check before it lands in pre-commit.
- Each finding is a judgement call (imperative delegation vs. navigational
  pointer), so the section-exemption list may need widening for section
  headings other plugins use.

## Acceptance

- [ ] Measure a full-corpus run (wall clock, findings count) and record it here
- [ ] Either widen `DELEGATION_SCOPE_DEFAULT` to the full corpus, or state why
      it stays scoped
- [ ] Fix whatever the widened scan surfaces, in the same PR as the widening
- [ ] Keep `SCOPE=` / `AUDITED=` accurate (flip `SCOPE_IS_REPO_WIDE` when true)

Refs #2442

laurigates:
@laurigates

Acceptance criteria 1 and 4 are done and criterion 2 is answered in PR #2487 (scope **stays** git-plugin, with the measured reason written into the script header). Criterion 3 needs a call from you — it is the only thing left.

## The measured pass

419 skills, via the `CHECK_DELEGATION_SCOPE` seam:

| | ISSUE_COUNT | wall clock |
|---|---|---|
| before | 47 | 247.3 s |
| after #2487 | **13** | **15.3 s** |

Two things the issue predicted came out differently, both worth recording:

[...]

Issue #2462, opened 2026-08-20
feedback(hooks-plugin): git commit -F <file> is blocked while its gh --body-file analogue is allowed


laurigates:
## Skill

`hooks-plugin/hooks/bash-antipatterns.sh` (the "commit message written to temp file" check, ~L415–438)

## Category

Enhancement

## Description

The hook blocks `git commit -F <file>` and redirects to the `git commit -m "$(cat <<'EOF' … EOF)"` heredoc form. But `-F <path>` is `git`'s exact analogue of `gh`'s `--body-file`, which this same hook was *taught to allow* in #1584/#1587 — and which `~/.claude/rules/tool-use-patterns.md` actively recommends for multi-line bodies ("Write the body to a file and pass `--body-file <path>`. This also dodges the multi-line quoting mess").

So the file-based body pattern is the recommended form for one command and blocked for its direct counterpart in the other.

The check fires as designed — this is not a pattern-matching false positive like #1584. `cat > /tmp/commit-439.txt` matches `cat\s*>\s*[^|]*commit`, there is a real `git commit`, and the message has a conventional-commit prefix. All three conditions genuinely hold. The question is whether `-F <path>` should be an accepted destination rather than something to redirect away from.

## Evidence

Observed in `laurigates/mcu-tinkering-lab` while committing a long message. Probing the shipped hook directly (v2.9.0) with a control:

| Command shape | Hook result |
|---|---|
| `cat > /tmp/msg.txt <<'EOF' … EOF` + `git commit -F /tmp/msg.txt` | **rc=2, blocked** |
| `cat > /tmp/body.md <<'EOF' … EOF` + `gh pr create --body-file /tmp/body.md` | rc=0, allowed |
| `git commit -m "fix(x): y"` (control) | rc=0 |

The control matters: it shows the hook is not simply blocking every `git commit`, so the first row is specifically about the `-F`-plus-temp-file shape.

Reproduction script (needed because writing the probe inline is itself blocked by the same check — a small piece of evidence in its own right):

```bash
out=$(printf '%s' "$CMD" | python3 -c 'import json,sys; print(json.dumps({"tool_name":"Bash","tool_input":{"command":sys.stdin.read()}}))' | bash hooks-plugin/hooks/bash-antipatterns.sh 2>&1); echo "rc=$?"
```

## Suggested Action

[...]

Issue #2442, opened 2026-08-19
feedback(git-plugin): git-pr-watch delegates to git-pr-feedback, which the model cannot reach


laurigates:
## Skill

`git-plugin/skills/git-pr-watch/SKILL.md`

## Category

Bug

## Description

`git-pr-watch` is catalog-present (no `disable-model-invocation`), and its
reaction table instructs the agent to hand off to two sibling skills:

| Line | Instruction |
|---|---|
| `:74` | Review comment / `CHANGES_REQUESTED` → "Address it via **`/git:pr-feedback`** (the canonical react-to-threads engine — do not duplicate its logic here)" |
| `:75` | CI failure → "Diagnose and fix via **`/git:fix-pr`**; re-push" |

After PR #2434, `git-fix-pr` is reachable by the model. **`git-pr-feedback` is
not** — it still carries `disable-model-invocation: true`, so an agent following
`git-pr-watch`'s own instructions can act on the CI branch but not the review
branch.

## Why this is not simply "unflag it too"

`git-pr-feedback` was the closest call in the #2439 audit, and the argument for
keeping it gated is real: its default path posts replies and resolves threads on
**other people's** PRs, and `--all` fans out worktree subagents that push. That
is a reasonable thing to require deliberate invocation for.

So the defect is the **mismatch**, not the flag as such. Either:

- **(a)** unflag `git-pr-feedback` and rely on its own confirmation steps, or
- **(b)** keep it gated and change `git-pr-watch:74` to stop presenting it as an
  action the agent can take — reword to surface it as a recommendation for the
  user to invoke (`"report the thread and suggest /git:pr-feedback"`), matching
  what the agent can actually do.

(b) looks correct given the third-party-PR mutation, but it is a judgment call.

## How this differs from the precedent

Precedent: #1843, same flag, same orchestrator-conflict class in `feedback-plugin`
(landed in #1859). There, `session-end` invoked the skill through the **Skill
tool**, so the conflict surfaced as a hard runtime error:

```
Skill feedback-plugin:feedback-session cannot be used with Skill tool due to disable-model-invocation
```

Here it degrades **silently instead**. `git-pr-watch`'s `allowed-tools` is:

```
mcp__github__subscribe_pr_activity, mcp__github__unsubscribe_pr_activity, Bash(gh pr view *), Read, TodoWrite
```

— neither `Skill` nor `SlashCommand`. The delegation is prose, so there is no
[...]

PR #2487, opened 2026-08-23
fix(scripts): eliminate delegation-guard false positives and measure the full-corpus scan

laurigates:
## What

Runs the measured full-corpus pass #2483 asks for, fixes what it exposed in the guard itself, and answers the widening question on the resulting numbers.

`Refs #2483` rather than `Fixes` — one acceptance criterion is deliberately left open, see **The widening decision** below.

## The measurement

Full corpus = all 419 `*-plugin/skills/*/SKILL.md`, via the existing `CHECK_DELEGATION_SCOPE` test seam:

| | FILES_SCANNED | ISSUE_COUNT | wall clock |
|---|---|---|---|
| before | 419 | 47 | **247.3 s** |
| after | 419 | **13** | **15.3 s** |

Both rows independently re-run and confirmed. `SCANNED_EMPTY=false`, `SELF_REFS_SKIPPED=258`, `GATED_STATEMENT_EXEMPTIONS=3`.

**34 of the original 47 were false positives**, in three structurally distinct classes.

### Class A — self-reference (30)

A skill referencing its own slash command. `git-api-pr/SKILL.md:201-206` is a Quick Reference usage table:

```
| Single file fix | `/git:api-pr file.ts --title "fix: typo"` |
```

`git-api-pr` *is* `/git:api-pr`. Documenting your own invocation is not delegating to anything. Same shape in `hooks-*`, `project-init`, `langchain-init`, `blog-post`, `git-issue`, `git-pr-feedback`, `git-upstream-pr`, `blueprint-prp-execute`.

### Class B — the preamble before the first `##` heading (1)

The scanner sets `section` only on `## `, so everything above the first H2 was scanned as section `""`, which `is_navigational_section` does not exempt. That region holds the frontmatter and the H1 title — and by house convention the H1 *is* the skill's own command (`# /project:init`).

### Class C — teaching the invariant (3)

The one that matters most. `blueprint-autopilot/SKILL.md:82`:

> Work-order **creation stays human-only** (`/blueprint:work-order` keeps `disable-model-invocation: true` — never invoke it from autopilot).

The line states this guard's own rule, forbids the invocation, and was flagged for saying so. Widening the scope with this unfixed would have turned the guard red on correct content — how a guard gets disabled rather than obeyed. The repo already makes this instruction-vs-explanation discrimination in `check-agent-tool-selection.sh` and `check-branch-containment-guidance.sh`; this follows that shape.

## Performance: the memoization wasn't the bottleneck

#2483 predicted `resolve_sibling()`'s per-reference re-globbing as the cost. Sibling resolution *is* now indexed once per plugin (with frontmatter parsed in pure bash) — but profiling put ~94% of the 247 s in a `printf | grep -oE` running **per scanned line**, across ~125,000 lines. That is now a `case` pre-filter plus a bash `[[ =~ ]]` extraction using the same ERE, leftmost-first, same token order.

Both remaining `grep` matches also moved from pipes to here-strings. That is a pre-existing latent bug, not cosmetics: under `pipefail` a `grep -q` that matches and closes the pipe early can SIGPIPE the writer and flip the `if` — the same race as #1744 and #2462.

## The widening decision: scope stays as-is

Performance no longer blocks it (15.3 s is inside a pre-commit budget), but `ISSUE_COUNT != 0`, so widening would put the guard red on `main` over content nobody has adjudicated. All 13 residuals are in `blueprint-plugin`, and each is a judgement call — reword into the recommendation form, or un-gate the sibling — which is the maintainer's call, not this PR's.

`DELEGATION_SCOPE_DEFAULT`, `SCOPE_IS_REPO_WIDE=false`, `.pre-commit-config.yaml` and `plugin-pr-checks.yml` are therefore **unchanged**. The measured before/after table and the reason are written into the script header, replacing the "tracked separately as issue #2483" sentence. `SCOPE=` / `AUDITED=` / `SCOPE_IS_REPO_WIDE=` stay accurate, so `STATUS=OK` still cannot be misread as "the marketplace is clean".

The 13 residuals are listed on #2483 with a read on each, so the decision can be made from the thread.

## Regression coverage

`scripts/tests/test-check-delegation-reachability.sh`, 71 → **105 assertions**. Per `.claude/rules/regression-testing.md`, every class is pinned in **both** polarities — without the negative half, a guard that silently stopped detecting anything would pass:

| Class | Positive (must pass) | Negative (must still ERROR) |
|---|---|---|
| A | self-reference | a *different* gated sibling on a comparable line |
| B | reference in the preamble | identical text under `## Execution` |
| C | the gated-statement form | a bare `Address it via /ns:gated-thing` |

Guard integrity: the fixture run must be non-vacuous (`FILES_SCANNED > 0`, `SCANNED_EMPTY=false`) — the #2219/#2290 zero-scan lesson — and the Class A fixtures drop their H1 so A's guard-integrity assertions are not silently satisfied by the Class B fix.

**Mutation-verified** (originals kept, each fix reverted individually):

| Mutation | FAILED |
|---|---|
| Class A reverted | 3 |
| Class B reverted | 2 |
| Class C reverted | 11 |
| verbatim pre-fix script | 18 |
| fixed script | 0 (105 passed) |

Guard-integrity assertions stayed green under every individual mutation, so each failure is attributable to its own class rather than to general breakage.

## Test plan

- [x] `scripts/tests/test-check-delegation-reachability.sh` — `PASSED=105 FAILED=0`
- [x] `scripts/check-delegation-reachability.sh` — `ISSUE_COUNT=0 STATUS=OK`, exit 0
- [x] `scripts/lint-shell-scripts.sh` — 0 errors (9 warnings, all pre-existing in `macos-plugin` bench scripts)
- [x] `scripts/run-skill-script-tests.sh` — `PASSED=138 SKIPPED=5 FAILED=0`
- [x] `pre-commit run --files <3 touched files>` — exit 0
- [x] Full-corpus before/after re-measured independently

`run-skill-script-tests.sh` reports `STATUS=ERROR` from 5 pre-existing `required_test_skipped` entries for tooling absent from this sandbox (`ast-grep` ×2, `cargo-generate`, `task` CLI ×2) — environmental, none related to these files.

## Note on shellcheck

`shellcheck` was **already failing on `main`** for the test file — SC2016 on ~20 fixture strings carrying backticked `/demo:feedback` inside single quotes, where the quoting is deliberate because the text must reach the checker unexpanded. Confirmed by running the hook against the unmodified file from `origin/main`. Per `.claude/rules/shell-scripting.md` § "Suppressing shellcheck findings" this adds a justified file-level `# shellcheck disable=SC2016` before the first command rather than churning 20 assertions.

## Deliberately unchanged

No `SKILL.md` is touched. Whether a genuine residual delegation should be reworded or its sibling un-gated is reserved for the maintainer — the same choice #2442 put to you before settling on the recommendation form for git-plugin.

Opened ready for review rather than draft, per the scheduled-maintenance routine that produced it.

https://claude.ai/code/session_018bBDx1mZeLgujessnPyfFC

---
_Generated by [Claude Code](https://claude.ai/code/session_018bBDx1mZeLgujessnPyfFC)_


laurigates:

Issue #2286, opened 2026-08-07
skills: strip unrendered template directives from deps-install, project-init, bun-add


laurigates:
## Evidence

Surfaced while landing #2284 (the `code-lint` half of #2265) during the 2026-W32 friction-learner run. Findings file: `~/.claude/friction-reports/2026-W32-frictions.md`.

`#2265` reported unrendered `{{ if PROJECT_TYPE }}` conditionals in **one** skill. The defect class is wider. Verified independently by direct grep against `main`:

```
grep -rn --include=SKILL.md -E '\{\{ *(if|endif|else|elif|for|endfor) ' .
```

| Skill | Plugin (release-please package) | Directive lines |
|---|---|---|
| `deps-install` | `tools-plugin` | **14** — 7 `{{ if PACKAGE_MANAGER == "…" }}` / `{{ endif }}` pairs (uv, bun, npm, yarn, pnpm, cargo, go) |
| `project-init` | `project-plugin` | **2** — `{{ if $3 == "--github" }}` / `{{ endif }}` |
| `bun-add` | `typescript-plugin` | **inline** — `bun add {{ if DEV }}--dev {{ endif }}{{ if EXACT }}--exact {{ endif }}$PACKAGE` |
| `code-lint` | `code-quality-plugin` | 8 — **being fixed in #2284**, not part of this issue |

**Not a defect, deliberately excluded:** `hooks-plugin/skills/hooks-permission-request-hook/SKILL.md:83` mentions `{{ if ... }}` in prose ("remove `{{ if ... }}` markers") — that is a legitimate citation of the syntax, not an unrendered directive. Any sweep or guard must not flag it.

Nothing in this repo renders these. An agent reading the skill sees raw template syntax presented as instructions.

## Proposal

Strip the directive lines from the three remaining skills, the same way #2284 does for `code-lint` — but **one PR per plugin**, because each is a separate release-please package and folding them together would mix scopes in a single squash-merge commit (`.claude/rules/conventional-commits.md`: the scope selects which package releases).

Two of the three are **not** as mechanical as `code-lint` was, which is why this is RESEARCH rather than three ready-to-go PRs:

- **`deps-install`** — plausibly the same behaviour-preserving deletion as `code-lint` if the surrounding headings already select the package manager. Needs checking; if they don't, the skill needs real detection prose, not just deletion.
[...]

PR #2502, opened 2026-08-25
fix(session-plugin): report how far HEAD trails upstream via a BEHIND key

### LINKED ISSUE #2500: feedback(session-plugin): session-survey reports ahead/dirty but never behind, so a stale checkout reads as clean
## What

`scripts/session-survey.sh`'s `GIT` section reports `DIRTY` and `UNPUSHED`, but there is no key for how far the checkout is **behind** its upstream. `grep -n "UNPUSHED\|BEHIND\|rev-list"` over the script returns only the two `UNPUSHED` echoes.

## What it looked like

At the end of a session in `ForumViriumHelsinki/silverbucket-helper` where four PRs had just been merged, the survey reported:

```
=== GIT ===
IN_GIT=true
BRANCH=main
DIRTY=false
UNPUSHED=0
STATUS=OK
=== END GIT ===
```

That reads as "on main, clean, nothing to push" — a settled checkout. It was three commits behind `origin/main` at the time, and by the end of the session, four.

The `COMMITS` section compounds it: it runs `git log --oneline` with no ref, so it lists HEAD's history. In that same digest it reported the pre-session commits and none of the four merges that were the session's entire output. `git log` on HEAD is the right default for feature-branch work, so I do not think that line is wrong on its own — it is only misleading because nothing nearby says HEAD is stale.

## Why it is worth a key

Staleness is what makes a checkout produce confidently wrong answers, and the wrongness looks like a finding rather than an error.

Concretely, in that session I ran `gitleaks detect` from the working tree and hit a config-load failure:

```
Failed to load config error="'AllowList.Paths[0]' expected type 'string', got unconvertible type 'map[string]interface {}'"
```

I reported that as a defect worth its own issue — the repo's secret gate appearing to be a no-op. It was not: a merged PR three commits ahead had already fixed exactly that malformation, and its config comment quotes the same error verbatim. The stale checkout produced a plausible, specific, entirely obsolete finding, and nothing in the survey suggested the tree was behind.

A `BEHIND=` key costs one `rev-list` and would have flagged it at the top of the wind-down.

## Suggested shape

```sh
behind=$("$git_bin" -C "$project_dir" rev-list --count 'HEAD..@{upstream}' 2>/dev/null || echo 0)
echo "BEHIND=${behind}"
```

`@{upstream}` rather than a hardcoded `origin/main`, so it degrades to 0 on a branch with no upstream — the same failure posture `UNPUSHED` already takes (see the comment at line 221 about branches with no upstream).

Consumers could then treat `BEHIND>0` the way `PROJECT_CONFIDENCE=low` is already treated: not an error, but a "this number was measured against a stale basis" caveat. `session-end`'s digest is the natural place to surface it, since that is the moment findings get written into issues and rules.

laurigates:
`session-survey.sh`'s `GIT` section reported `DIRTY` and `UNPUSHED` but had no key for how far the checkout trails its upstream, so a stale tree read as settled — and, as the issue documents, produced a confident, specific, entirely obsolete finding that a merged commit three ahead had already fixed.

Fixes #2500

## The change

`BEHIND=<n>` is emitted in the `GIT` section directly after `UNPUSHED`, and in the `--summary` block. It is a **caveat, not an error**: it never touches `STATUS` (stays `OK`), is not added to `THREADS`, and is emitted unconditionally so the section shape does not vary with repo state.

Consumers follow the key where they already handle `UNPUSHED`: `session-spinup-nudge.sh` mirrors the existing thread line, and `session-spinup`'s SURFACE filter gains `BEHIND ≥ 1`.

## One deliberate deviation from the issue's sketch

The issue prescribed:

```sh
behind=$(… rev-list --count 'HEAD..@{upstream}' 2>/dev/null || echo 0)
```

This PR uses the awk form instead, matching `UNPUSHED`'s existing implementation one line above:

```sh
behind=$("$git_bin" -C "$project_dir" rev-list --count 'HEAD..@{upstream}' 2>/dev/null \
  | awk '{n=$1} END{print n+0}')
```

`|| echo 0` after a command that can both print **and** fail yields a two-line value, which breaks the `KEY=VALUE` contract. That is #2286 — the bug already fixed in `UNPUSHED`, whose in-file comment describes the same trap ("…making the `-gt` test below an unconditional false"). Taking the sketch literally would have reintroduced the identical defect in its own sibling, so the acceptance criterion "mirror `UNPUSHED`'s implementation" won over the literal snippet. Test AO4 pins the bare-integer-line guard against exactly that.

`@{upstream}` rather than a hardcoded `origin/main` is kept as specified: it resolves per branch and fails cleanly on a branch with no upstream or a detached HEAD, degrading to `BEHIND=0` — the same posture `UNPUSHED` takes.

Worth noting for readers of the report: `BEHIND=0` is genuinely ambiguous — it means in-sync **or** no upstream **or** detached HEAD. That is inherited from `UNPUSHED`'s existing contract rather than introduced here, and it is now written down in `session-spinup/REFERENCE.md` instead of being folklore.

## Tests

`test-session-survey.sh` TEST AO (26 assertions) and the hook suite's W3 (3 assertions), per `.claude/rules/regression-testing.md`, plus a Known-Regressions row.

Guard integrity was weighted equally with coverage, since every `BEHIND=0` assertion would also pass against a hardcoded zero: **AO1** requires an in-sync branch *with a real upstream* to report `0`, and AO2/AO4/AO5 each carry a fixture-validity check that git itself agrees the fixture is 3-behind / upstream-less / detached. W3 asserts an in-sync clone stays **silent** before asserting the stale one fires.

Mutation-verified twice — 12 AO assertions fail against the pre-fix collector; deleting only the hook's thread line fails exactly W3's two while its guard-integrity assertion stays green.

Before: `PASS=522 FAIL=12`. After: `PASS=534 FAIL=0`, hook suite `21 passed, 0 failed`.

## Verification

- `bash scripts/run-skill-script-tests.sh` → `TOTAL=143 PASSED=138 FAILED=0`. The 5 `required_test_skipped` entries are pre-existing container gaps (`ast-grep` ×2, `cargo-generate`, `task` ×2 not installed), unrelated to this diff.
- `just lint-shell` → 0 errors (9 pre-existing warnings, all in the untouched `macos-plugin`).
- `just lint-compliance` → exit 0, zero ❌. The first `SKILL.md` wording pushed `session-spinup` from 9,898 to 10,090 chars, crossing the 10,000-char ⚠️ band; the SURFACE entry was trimmed back to 9,989 with the detail moved to `REFERENCE.md`, which has no ceiling.

Docs land in the same commit per `.claude/rules/docs-currency.md` — `session-spinup` SKILL.md/REFERENCE.md and the `session-plugin/README.md` collector table.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EZbdu63QLEE7ksTo6Yd3Ky)_


laurigates:

PR #2505, opened 2026-08-26
feat(agent-patterns-plugin): name the stderr-suppression trap in tool-result-traps

### LINKED ISSUE #2504: feedback(agent-patterns-plugin): tool-result-traps should cover suppressed stderr hiding a loud error as a clean negative
## What happened

While reviewing a PR I ran:

```
gh pr diff 36 -- .env.example 2>/dev/null | grep -E "^[+-]" | grep -i -B2 -A6 "partition"
```

Empty output. I read that as "this PR does not touch that file" and was one step from reporting
that a PR claiming `Closes #29` had not actually made the change that closes #29.

It had. The file was `+34/-5`.

## The mechanism — the rejection was loud, I muted it

`gh pr diff` takes no pathspec, and it says so, on stderr, with a nonzero exit:

```
$ gh pr diff 36 -- somefile
accepts at most 1 arg(s), received 2
$ echo $?
1
```

My own `2>/dev/null` discarded that message, leaving empty stdout — and empty stdout piped into
`grep` is indistinguishable from a successful command that found nothing. The nonzero exit
belonged to `gh`, but the pipeline reported the exit status of the last command, so that signal
was lost too.

So this is not a `gh` defect. It is `2>/dev/null` converting a hard, well-reported error into a
clean-looking negative result.

## Why it is worth a rule

`agent-patterns-plugin:tool-result-traps` already covers "a rejected flag looking exactly like
*no results*". This is the same family with a difference worth naming: **the tool did its job.**
No tool-side improvement helps, because the diagnostic was written, printed, and thrown away by
the caller. The habit that catches it is not "check whether the flag is supported" but "do not
suppress stderr on a command whose failure mode you have not already seen".

It also defeats the skill's own prescribed remedy in a specific way. The control test says to
re-run the same command shape against something known to be present. Run against a file the PR
definitely touches, this *still* returns empty — the shape is broken for every input, so the
control agrees with the false negative and confirms it.

What breaks it open is dropping the redirect:

```
gh pr diff 36 -- somefile        # accepts at most 1 arg(s), received 2
```

## Suggested addition to `tool-result-traps`

- **Suppressing stderr on an unproven command turns a loud error into a clean negative.**
  Redirect stderr away only on commands whose failure mode you have already seen. When a negative
  result gates a conclusion, re-run **without** the redirect before believing it — and note that
  a same-shape control test does not catch this class, because a shape that is broken for all
  inputs fails the control identically.
- Correct forms for this specific case: `gh pr diff <n>` and filter the unified diff, or
  `gh pr view <n> --json files` for per-file additions and deletions.

## Environment

`gh` 2.x, Claude Code, 2026-08-25. Reproduced deterministically; the transcript above is the
actual reproduction, not a reconstruction.

laurigates:
## What

Adds a section to `agent-patterns-plugin:tool-result-traps` covering the case where the caller's own `2>/dev/null` turns a loud, correctly-reported tool error into a clean-looking negative result.

It sits immediately after `## A rejected flag looks exactly like "no results"`, because it is that section's mirror — and the difference is the point: **the tool did its job.** `gh pr diff` takes no pathspec, so given one it exits non-zero and writes `accepts at most 1 arg(s), received 2` to stderr. The redirect discarded the diagnostic; the pipeline reported `grep`'s exit status rather than `gh`'s, so that signal was lost too. Empty stdout then reads as "this PR does not touch that file".

No tool-side improvement reaches this class — the message was written and printed before being thrown away downstream.

## Why it needed its own section rather than a line in the existing one

The skill's prescribed remedy **confirms** this failure instead of exposing it. Re-running the same shape against a file the PR definitely touches also comes back empty, because the shape is broken for every input — so the control agrees with the false negative. What breaks it open is dropping the redirect, not changing the input.

That is the load-bearing distinction, and it is why the guard below pins it separately.

## Changes

- `agent-patterns-plugin/skills/tool-result-traps/SKILL.md` — the new section (verbatim reproduction, the control-test-defeat paragraph, generalising habits, and the two correct forms: plain `gh pr diff` and filter the unified diff, or `gh pr view` with `--json files`); one new "When to Use" row; `modified:` bumped. The intro law was qualified in the same commit — it previously asserted "nothing on stderr" for *every* trap, which this one falsifies.
- `scripts/plugin-compliance-check.sh` — a `tool-result-traps`-scoped block in `check_skill_body()` pinning one literal token per idea: `failure mode you have already seen` (the habit) and `the control agrees with the false negative` (the control-test defeat — the subtlest half, and the first thing a prose "tightening" pass would drop).
- `.claude/rules/regression-testing.md` — the Known Regressions row.

## Note on the extra commits

Two commits landed on this branch after the original push, neither authored here:

- `fd28488 refactor(split): extract reference content from tool-result-traps` — the repo's own skill-splitter workflow, triggered because this addition pushed the body past the 10,000-char WARN band. It moved four later sections into a new `REFERENCE.md`.
- `486ede1 Merge remote-tracking branch 'origin/main'` — automated conflict resolution against #2508, which had independently added two sections to this same skill.

I verified the result rather than assuming it: every heading present on `main` survives across `SKILL.md` + `REFERENCE.md` (none lost), both pinned tokens remain in `SKILL.md` where the guard reads them, the new section is intact, and `SKILL.md` still links to `REFERENCE.md` so nothing is unreachable.

## Verification

- Both pinned tokens confirmed to sit on a **single line** — `grep -qF` is line-scoped, the trap #2174 records for `WIP at checkpoints`.
- Mutation-verified in both directions: replacing either token individually produces a ❌ naming it; the restored file reports none.
- `scripts/plugin-compliance-check.sh` exit 0, zero ❌.
- `pre-commit run` exit 0 on all changed files.

Fixes #2504

https://claude.ai/code/session_01Q3XDfZnYXT3ZX271cSaMCd


laurigates:

laurigates:
Checked the one recommendation above that changed on this PR, since it would be easy to misread as mine.

**`tool-result-traps: description is 296 chars`** — not from this branch. Measured across refs:

| Ref | Description |
|---|---|
| `5de0662` (base at branch point) | 281 chars |
| `9d983c9` (my commit) | 281 chars |
| `cca7f59` (**main**, via #2508) | **296 chars** |
| this PR's head | 296 chars |

It arrived through the automated main merge, so there is nothing to fix here and I have not touched it — trimming a description #2508 just wrote would be widening this PR into someone else's change.

Worth flagging separately though, because the number is closer to a cliff than the ⚠️ suggests: `.claude/rules/skill-quality.md` makes **>300 chars an ERROR**, and `audit-skill-descriptions.py --strict-length` enforces that in both pre-commit and the `Plugin: Lint skills` workflow. At 296 there are **4 characters of headroom** — the next person who adds a word to that description gets a hard block, on a line they may only be passing through. A ~40-char trim on `main` would put it back in the target band.

Everything else in the table is pre-existing and unchanged by this PR.

---

laurigates:

laurigates:

PR #2510, opened 2026-08-26
fix(configure-plugin): treat Renovate as satisfying the dependency layer

### LINKED ISSUE #2495: feedback(configure-plugin): configure-security recommends Dependabot with no Renovate detection
Ran `/configure:all` on `ForumViriumHelsinki/silverbucket-helper` (2026-08-24). `configure-security` reported:

```
DEPENDABOT=false
STATUS=WARN
  - SEVERITY=WARN TYPE=missing_dependabot MSG=no Dependabot config (.github/dependabot.yml)
```

The repo has a `renovate.json` in its root and has used Renovate for a long time. Dependency automation is fully configured; the check simply does not look for the tool actually in use.

`configure-security.sh:145`:

```sh
[ "$dependabot" = "false" ] && add_issue "WARN" "missing_dependabot" "no Dependabot config (.github/dependabot.yml)"
```

There is no corresponding `renovate` probe anywhere in the script.

## Why this is worse than a cosmetic false positive

Acting on the recommendation is actively harmful. Renovate and Dependabot both open dependency PRs, so configuring Dependabot alongside an existing Renovate setup produces **two bots raising duplicate PRs for the same upgrades**, which then race on the lockfile.

The failure mode is that a `--fix` run, or an agent working from the report without checking, adds `.github/dependabot.yml` to a repo that deliberately does not want one. The report gives no signal that anything needs verifying first — it reads as a straightforward gap.

The FVH org standardises on Renovate specifically, and runs it centrally from the infrastructure repo rather than per-repo (per-repo workflows are a documented billing antipattern there). So for that org the recommendation is wrong in every repo it fires on.

## Suggested fix

Detect the incumbent before recommending a replacement — same instinct as `tool-migration-cutover`:

1. Probe for `renovate.json`, `.renovaterc`, `.renovaterc.json`, `.github/renovate.json`, or a `renovate` key in `package.json`.
2. Emit `RENOVATE=true|false` alongside `DEPENDABOT=`.
3. When either is present, the dependency-automation layer is **satisfied** — no `missing_dependabot` warning.
4. Warn only when *neither* is configured, and phrase it as "no dependency automation (Renovate or Dependabot)" rather than naming one tool.

Worth applying the same shape wherever else a check names one specific tool as the standard while a sibling tool satisfies the same requirement.

laurigates:
## What

`configure-security` warned `missing_dependabot` on any repo without `.github/dependabot.yml`, with no Renovate probe anywhere in the script. A repo that has run Renovate for years reported a straightforward-looking gap.

Acting on that recommendation is actively harmful rather than merely noisy: Renovate and Dependabot are **alternatives, not complements**. Both open dependency-update PRs and both rewrite lockfiles, so installing the second bot makes them race each other on the same upgrades.

## Changes

- **`configure-security.sh`** — probes eight Renovate config paths plus a top-level `renovate` key in `package.json`; emits `RENOVATE=` and a `DEPENDENCY_AUTOMATION=` layer verdict (`RENOVATE || DEPENDABOT`) beside the unchanged `DEPENDABOT=`; counts the layer off the verdict; warns only when **neither** is configured, phrased as `no dependency automation (Renovate or Dependabot)`.
- **`SKILL.md` / `REFERENCE.md`** — document the new keys, tell the agent to read `DEPENDENCY_AUTOMATION` rather than `DEPENDABOT` alone, add a Renovate `## Context` probe, and gate Step 4 on checking the incumbent first. Detection alone would have fixed the *report* while leaving the *action* ("Create Dependabot config") unconditional.
- **`.claude/rules/regression-testing.md`** — the Known Regressions row.

## Two decisions worth flagging

**`TYPE=` renamed to `missing_dependency_automation`.** `grep -rn "missing_dependabot" .` returned exactly one hit repo-wide — the script's own `add_issue` call. Nothing consumed the literal, and it now names a different condition. The old token survives only as a negative assertion in the test.

**`DEPENDENCY_AUTOMATION` is a new key the issue did not ask for.** Without it every consumer re-derives the OR, and reading `DEPENDABOT` alone is precisely the misreading that caused this bug.

The script remains a pure filesystem probe — no network call was added.

## Verification

- New test group pins **both polarities**: Renovate-only and Dependabot-only each satisfy the layer and raise no warning; a project with neither still warns (guard integrity — a script that never warns would otherwise pass).
- The `package.json` discriminator is planted deliberately: a manifest whose only mention of Renovate is a **devDependency** must report `RENOVATE=false` — what a naive `grep renovate package.json` gets wrong. Detection uses `jq -e 'type == "object" and has("renovate")'`, with an anchored key match as the no-jq fallback.
- Mutation-verified: with the pre-fix script and the new test, the run aborts at `FAIL: expected RENOVATE=false`. Run directly against a Renovate-only fixture, the pre-fix script reproduces the reported output verbatim (`SECURITY_LAYERS_PRESENT=0`, `TYPE=missing_dependabot`); post-fix it reports `RENOVATE=true DEPENDENCY_AUTOMATION=true SECURITY_LAYERS_PRESENT=1` with no dependency warning.
- `plugin-compliance-check.sh` exit 0, zero ❌; `lint-shell-scripts.sh` 0 errors; `pre-commit` exit 0 on all changed files (including the sandbox Context-command execution check, which matters because a new `## Context` probe was added).

Fixes #2495

https://claude.ai/code/session_01Q3XDfZnYXT3ZX271cSaMCd

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q3XDfZnYXT3ZX271cSaMCd)_


laurigates:

laurigates:

PR #2520, opened 2026-08-27
feat(prose): add prose-check, a self-evaluation gate against the house rubric

laurigates:
## What

Adds `prose-plugin:prose-check`, a runnable gate that evaluates a draft against
`~/.claude/rules/communication.md` before it ships into a PR body, issue, doc,
or rule. Ships as a vale style package (`prose-plugin/styles/House/`), a skill
with two scripts, an opt-in PostToolUse nudge, and a regression guard with 12
paired test cases.

## Why

Agents write most outward-bound prose here. `communication.md` already names the
three tics to cut on sight, mandates ending on the fact, and specifies when a
TL;DR footer is required — but nothing checked a draft against it. The rule was
prose an agent might recall.

The goal is not a script that judges prose. It is to narrow the candidate set
deterministically so model judgment lands on a handful of flagged sentences
instead of a whole re-read. That is the split in
`.claude/rules/offload-to-deterministic-substrate.md`, applied to writing.

## How

Three layers, each justified by a measured limit in the one above it:

| Layer | Covers | Why not the layer above it |
|---|---|---|
| vale | token shapes, doc metrics, markdown scoping | code/table/heading skipping is free from its markdown parser; it has **no ordinal scope**, so it cannot express "paragraph-final" |
| harper | grammar, readability, sentence length | independent segmenter, second opinion on length |
| `positional-tics.py` | paragraph-final position + tic shape, TL;DR footer | the position half of "position plus shape" |
| model | *is* this a chiasmus? | irreducibly judgment |

`vale`, `harper-cli`, and `uv` are each optional. A missing one reports
`AVAILABLE=false` and the other layers still run.

Segmenter choice is measured, not assumed: pysbd scores 47/48 on the English
Golden Rules Set, against 43/48 for sentencex and 27/48 for punkt-grade
segmenters.

### Candidates, not verdicts

Every line the check emits is a sentence to judge. `communication.md` is
explicit that hedges carrying real uncertainty stay, so a `Hedge` hit is often a
true negative. `prose-distill/SKILL.md` returns `ISSUE_COUNT=0` from the script
layer and scores weak-word hits from the vale layer, because listing filler
words is its subject matter.

### Two mechanics the build turned up

**Joining hard line wraps before segmenting is load-bearing.** Every segmenter
benchmarked treats a newline as a sentence boundary and the rules tree wraps at
~76 columns. Without the join, `communication.md` measures 78 sentences / 8.2
mean words instead of 46 / 15.2.

**vale concatenates `raw:` list entries into a single pattern** rather than
OR-ing them the way it does `tokens:`. `TicketPlaceholder`'s five entries became
one impossible regex that matched nothing — and zero alerts from a broken rule
are indistinguishable from zero alerts from clean prose. Fixed with explicit
alternation, and `fixtures/house-rule-control.md` now trips every House rule
while the guard asserts each one fires (`RULES_EXERCISED=7`).

### Curated allowlist, not a wholesale load

`write-good.E-Prime` alone produced 37 of 47 alerts on `communication.md`. The
third-party rules are cherry-picked rule by rule in `styles/.vale.ini`; E-Prime
is explicitly disabled and the guard pins it off. Across the 52-file rules tree
the configured set now reports 112 alerts and zero E-Prime.

### Single-source guard

`communication.md` stays canonical. Every House rule carries a `link:` back to
the rule that owns its criterion instead of restating rationale.

`scripts/check-prose-house-style.sh` asserts the derived copy has not drifted.
It checks that the vale rule and the script layer agree on the significance
markers, and that the Filler/Hedge/ThroatClearing lists still match
`prose-distill/SKILL.md`.

## Verification

| Check | Result |
|---|---|
| Known-positive fixture | harper reports 45 words, the script reports 44, at `communication.md:107` — found independently by two layers |
| Planted control (real rules-tree prose) | 4 chiasmus candidates, 2 of them textbook mirrored clauses |
| Negative control | short plain answer with `--kind answer` → `STATUS=OK`, no TL;DR complaint |
| False-positive floor | `prose-distill/SKILL.md` → `ISSUE_COUNT=0` from the script layer |
| Absent-binary path | vale off `PATH` → `VALE_AVAILABLE=false`, other layers still run |
| Hook | unset → 0 bytes, exit 0; set → valid `updatedToolOutput` **object**, write unaffected |
| Guard regression suite | 12/12, every planted break caught, pristine copies pass |
| `plugin-compliance-check.sh` | prose-plugin all green |

`pre-commit run --all-files` has two failures that also fail on a pristine
`origin/main` worktree: the `shellcheck` hook (165 findings repo-wide) and
`test-check-plugin-readme-currency.sh` case H. Neither involves files changed
here — see the note below on case H.

## Follow-up

Filed as separate issues rather than a checklist here, so they outlive the merge:

- laurigates/claude-plugins#2521 — `check-plugin-readme-currency.sh:48` gates its
  test seam on `-n "$VAR"`, so the seam cannot express an empty staged set. TEST H
  therefore asserts the developer's index: it passes 25/25 in a clean worktree and
  fails for anyone running the suite with work staged.
- laurigates/claude-plugins#2522 — `check-docs-index.sh` has no local pre-commit
  trigger for a plugin-only change, so its `doc_count_drift` finding first appeared
  in CI on this PR while the local run was green.

## Companion PR

`laurigates/dotfiles#390` declares `vale` and `harper-ls` in mise and `harper`
in the Homebrew dev profile. This PR works without it — the tool layers are
optional.

https://claude.ai/code/session_01Bw7GGc869UR8aHTA41wRzU


laurigates:
## 🔧 Plugin Config Auto-Fix Applied

The plugin configuration files were out of sync. I've automatically fixed and pushed the changes.

<details>
<summary>Issues that were fixed</summary>

```
Checking plugin configuration sync...
Repository root: /home/runner/work/claude-plugins/claude-plugins

Found 2 issue(s):

  - Version mismatch for 'hooks-plugin': manifest=2.10.2, marketplace=2.11.0
  - Version mismatch for 'prose-plugin': manifest=1.3.4, marketplace=1.4.0

Run with --fix to automatically resolve these issues.
```
</details>

The workflow will re-run automatically. If it passes, you're good to merge!

Issue #2431, opened 2026-08-19
feedback(hooks-plugin): bash-antipatterns ^-anchored checks match heredoc body lines, blocking commits whose message mentions a watched command


laurigates:
## What

`bash-antipatterns.sh` greps the raw command string with `^`-anchored patterns. `^` in `grep` matches the start of **every line**, not the start of the command — so in a multi-line command, a **heredoc body line** that happens to begin with a watched word is treated as if it were the command itself.

Hit while committing the fix for `foundryvtt-mcp#206`: a `git commit -F -` whose message described a *connection timeout* was blocked with "REMINDER: The 'timeout' command is usually unnecessary — remove the timeout wrapper". There was no wrapper; the word was prose in the commit message.

## Repro

```
printf '%s' "$CMD" | jq -Rs '{tool_name:"Bash", tool_input:{command:.}}' | bash hooks-plugin/hooks/bash-antipatterns.sh
```

| `$CMD` | exit | correct? |
|---|---|---|
| `git commit -F - <<MSG`⏎`timeout waiting for the ack was the symptom`⏎`MSG` | **2 (blocked)** | ✗ false positive |
| `git commit -F - <<MSG`⏎`the symptom was a 10s timeout before the ack`⏎`MSG` | 0 | ✓ (word mid-line) |
| `timeout 30 node x.js` | 2 (blocked) | ✓ genuine wrapper |
| `git status --short` | 0 | ✓ control |

The mid-line and control rows are what make the first row a defect rather than a design choice: the check is sensitive purely to the word's **column position inside a heredoc body**, which carries no meaning.

## Not limited to the timeout check

Four checks use `^`-anchored greps against `$COMMAND` (v2.8.0: lines 402, 542, 583, 616). Probing them the same way:

| Check | Heredoc body line starts with the phrase | Result |
|---|---|---|
| `timeout` wrapper (402) | `timeout waiting for the ack…` | **blocked** ✗ |
| `git add -A` / `git add .` (542) | `git add . was the mistake that lost the edit` | **blocked** ✗ |
| `git reset --hard` (583) | `git reset --hard would have discarded it` | passes ✓ |

So a commit message that *describes* a `git add .` mistake cannot be committed — which is exactly the kind of message someone writes right after making that mistake.

## Why it matters

[...]

PR #2525, opened 2026-08-28
fix(hooks-plugin): exclude heredoc bodies from the teach hook's pipeline count

### LINKED ISSUE #2518: hook: bash-antipatterns-teach counts heredoc-body pipes — long-pipeline nudge fires on markdown tables
## Evidence

Found live during the 2026-W35 friction-learner orchestration run — the nudge fired on the orchestrator's own `gh issue comment` command, which contains **zero shell pipes**. The `|` characters it counted were markdown table cells inside a heredoc body.

`hooks-plugin/hooks/bash-antipatterns-teach.sh` scans raw `$COMMAND`. It has **0** references to `COMMAND_SHELL_ONLY`; the sibling `bash-antipatterns.sh` has **24**. So heredoc bodies and trailing `#` comments are counted as shell text.

The segmentation step strips quoted strings and splits on `;` / `&&` / `||` / newlines. A heredoc body is neither quoted nor a statement separator, so each markdown table row survives as its own "pipeline". A row such as

    | `echo/printf > file` -> Write | 18 | 16 | 15.2% | 12.5% |

carries 6 pipe characters and satisfies the discouraged-head test, because the cell text literally contains a `printf` token followed by a space. The detector matched a table cell **describing another detector**.

## Reproduction

Run against a **copy** of the shipped hook (deployed md5 `247c7b3093e179ac136734cccc8c6849`, verified unchanged before and after):

| Case | Command shape | Result | Wanted |
|---|---|---|---|
| A | heredoc write whose body is a markdown table | **FIRED** | silent |
| B | same table text as a single-quoted `--body` argument | silent | silent |
| C | genuine 5-pipe `cat`-headed pipeline | FIRED | FIRED (control) |
| D | `echo hello` | silent | silent (control) |

**B is the load-bearing row**: the identical table text is already handled correctly when it arrives as a quoted argument, so the hook has the "this is data, not shell" concept — it just does not extend it to heredocs. C and D are the must-still-fire / must-stay-silent controls, so a permissive patch cannot pass silently.

## Proposal

Feed the teach hook the same `COMMAND_SHELL_ONLY` projection the safety hook already computes (heredoc bodies and trailing `#` comments stripped), rather than raw `$COMMAND` — for the `long-pipeline` segmentation at minimum, and audit the `cat`/`head`/`tail` and `grep`/`rg` nudges in the same file for the same exposure.

The two hooks are separate scripts, so this is either a small shared helper or a lift of the existing projection. That choice is the only design question; the behaviour is not in doubt.

## Why this is not already a PR

It was found by the orchestrator during Phase 2, not by the diagnosis phase, so it carries no corpus baseline — `long-pipeline` read **0 corpus events** in W35 (third consecutive zero week). The one instance here is live, reproduced, and single-cause, but the week's data does not tell you how often it bites.

Note the asymmetry that makes it worth fixing anyway: the safety detectors are pure-regex **by design** and fire inside heredoc bodies deliberately, because a false positive there is the correct trade. `bash-antipatterns-teach.sh` is a **non-blocking style nudge** with no such justification — it is spending the agent's attention on markdown.

## Acceptance criteria

- Case A silent, cases C and D unchanged, on the shipped hook.
- A checked-in regression case pinning the heredoc-body shape (per `.claude/rules/regression-testing.md`).
- Future weeks: `long-pipeline` stays at 0 events, and no friction-learner run is nudged by its own report prose.

## Related: this is the third data-as-pattern instance in three runs

The W35 watch list tracks "pattern-literal-as-data on safety detectors" at a gate of >=10 events with >=30% data-only. Live instances so far: W34 (permission-bits literal in a JSON test payload), W35 diagnosis (network-pipe-to-shell pattern quoted in report prose), and W35 Phase 2 — **filing this very issue was blocked**, because an earlier draft of this body quoted a permission-bits literal while describing the W34 instance. The body was reworded rather than retried, per `handling-blocked-hooks.md`.

Those three are the safety hooks behaving as designed and are **not** part of this issue's scope. They are recorded here only because they establish the pattern this teach-hook defect belongs to: detectors reading documentation about detectors as if it were shell.

laurigates:
## What

`hooks-plugin/hooks/bash-antipatterns-teach.sh` scanned raw `$COMMAND` with **zero** references to `COMMAND_SHELL_ONLY`, while the sibling safety hook has 24 — so heredoc bodies and trailing `#` comments were counted as shell text. All five detectors now scan a lifted `COMMAND_SHELL_ONLY` projection instead.

## The A/B/C/D matrix

| Case | Shape | Pre-fix | Post-fix | Wanted |
|---|---|---|---|---|
| A | heredoc write whose body is a markdown table | **FIRED** | silent | silent |
| B | same table text as a single-quoted argument | silent | silent | silent |
| C | genuine 5-pipe `cat`-headed pipeline | FIRED | **FIRED** | FIRED (control) |
| D | `echo hello` | silent | silent | silent (control) |

C and D are what stop a permissive patch passing silently — a change that simply disabled the detector would turn C red.

## The audit found four more, by a different mechanism

The issue asked me to audit the `cat`/`head`/`tail` and `grep`/`rg` nudges. All four siblings had the same raw-`$COMMAND` exposure, but via a *second* bug: **`^` in grep anchors to the start of every line**, so a heredoc-body line merely *beginning* with the watched word was read as the executed command. Confirmed empirically against the pre-fix hook:

- `head`/`tail` — a `git commit -F -` whose message body contains `tail -20 build.log` emitted the Read offset/limit hint
- `grep`/`rg` — a body line `grep -rn foo src/` emitted the Grep hint
- `find` — a `gh pr create --body "$(cat <<'EOF' … )"` repro section containing `find . -name '*.ts'` emitted the Glob hint
- `ls` — a body line `ls *.md` emitted the Glob hint
- `cat` — exposed only in principle (its own `! grep -Eq '<<|cat\s*>'` exemption already suppressed it for any command containing a heredoc), now fixed structurally anyway

All fixed by the same projection.

## Design choice: lift, don't share

`bash-antipatterns.sh` is **untouched** (`git diff --name-only` confirms). Its detectors are pure-regex *by design* and fire inside heredoc bodies on purpose — a false positive there is the correct trade for a safety block. Refactoring it into a shared helper to serve a style nudge would put that behaviour at risk for no benefit (`.claude/rules/hook-block-vs-nudge.md`). The projection is the same heredoc-stripping awk and quote-state-aware `strip_trailing_comments()` used since #2431/#2106.

Switching `cat` to the projection is safe because the awk keeps the heredoc-**opening** line, so both `<<` and `cat >` still register.

## Regression test

`hooks-plugin/hooks/test-bash-antipatterns-teach.sh`, 44 → 50 assertions, each piping JSON into the **shipped** script rather than retyping its logic.

Two **over-strip** controls guard the projection itself, so it cannot swallow real commands: a `grep -rn` on the line *after* a heredoc terminator must still fire, and `cat README.md  # skim the intro` must still fire — a trailing comment annotates a command, it does not exempt it.

Mutation-verified against `git show HEAD:…` (the shipped pre-fix hook): **44 passed / 6 failed** pre-fix — case A plus the four sibling-nudge heredoc cases and the trailing-comment case — with controls B, C and D all passing pre-fix. **50 / 0** post-fix. The suite discriminates in exactly one direction.

## Test plan

- `bash hooks-plugin/hooks/test-bash-antipatterns-teach.sh` → 50 / 0
- `bash hooks-plugin/hooks/test-bash-antipatterns.sh` → exit 0, 0 FAILs (the ast-grep structural block SKIPs — no `ast-grep` on PATH in this container; pre-existing and equally true before this change)
- `shellcheck -S warning` clean on both files
- `bash scripts/lint-shell-scripts.sh` → 0 errors (9 pre-existing `macos-plugin` warnings)
- `check-context-engineering.py --strict` → exit 0

## Notes

No doc change was owed: `.claude/rules/bash-tool-replacements.md` documents `long-pipeline` as counting pipes *per pipeline, not per invocation*, which stays exactly true — this fixes what counts as shell text, not the documented contract. The third file is the Known Regressions row `.claude/rules/regression-testing.md` requires.

The issue's third acceptance bullet ("future weeks: `long-pipeline` stays at 0 events, and no friction-learner run is nudged by its own report prose") is a forward-looking observation this PR cannot itself discharge — worth confirming on the W36 reading.

Fixes #2518

---
_Generated by [Claude Code](https://claude.ai/code/session_01TSPUbAp8KVZVE4VmNdCVnq)_


laurigates:

laurigates:
**Superseded by #2533 — recommend closing this unmerged.** Flagging rather than resolving the conflict, because resolving it would mean re-landing a second implementation of a fix `main` already has.

#2533 merged as `61e2816b` and fixes #2518 in the same file, by the **shared-helper** route this PR deliberately declined: `bash-antipatterns-teach.sh` now sources `hooks/lib/command-views.sh` and carries 16 `COMMAND_SHELL_ONLY` references. That is the design question the issue itself flagged as "the only design question", and #2533 answered it the other way. The conflict `mergeable_state=dirty` reports is the two implementations colliding in `bash-antipatterns-teach.sh` — not a stale base.

**Verified `main` already satisfies the issue's acceptance criteria.** Extracted `main`'s hook plus its `lib/command-views.sh` into a scratch dir and piped the matrix at the shipped script:

| Case | Shape | `main` result | Wanted |
|---|---|---|---|
| A | heredoc write whose body is a markdown table | **silent** | silent |
| C | genuine 5-pipe `cat`-headed pipeline | **FIRED** | FIRED (control) |
| D | `echo hello` | silent | silent (control) |

**And `main`'s regression coverage is a superset of this PR's.** Its `test-bash-antipatterns-teach.sh` already pins case A, a heredoc-body case for each of the five detectors, the trailing-comment case, and the over-strip control this PR added (a real `grep` after the heredoc terminator) — plus one this PR does **not** have: a real `grep` sharing the heredoc-*opening* line. So there is no test delta worth porting either.

Nothing here is left to salvage: the fix and its tests both exist on `main` in a better-factored form. My recommendation is to close this PR unmerged rather than resolve the conflict.

Two notes:

- Auto-merge (squash) is currently enabled on this PR. It cannot fire while the PR is conflicted, but it is worth disabling if you close rather than merge.
- **#2518 is still open.** #2533's merge appears not to have closed it, and this PR's `Fixes #2518` is what would have. If you close this unmerged, #2518 needs closing by hand — the defect it describes is genuinely fixed on `main`.

I have not resolved the conflict, changed auto-merge, or closed anything.

---

laurigates:
**Correction — disregard my previous comment.** It was wrong and was posted after this PR had already merged.

This merged at 07:06:56 as `d9323a5d`, resolved down to **+73 lines across 2 files** — `test-bash-antipatterns-teach.sh` and the `regression-testing.md` row — with the hook change dropped because #2533 (`61e2816b`) already supplied the projection. That is the right outcome and better than what I recommended: the shared-helper implementation stands, and this PR's regression cases land on top of it.

What went wrong on my side: I read the PR at head `9440c8f8` and got `state=open, mergeable_state=dirty`, then ran `git merge-tree` against that head. The conflict was real *for that snapshot*, but the branch had already been resolved onto `61e2816b` (head `07f8fa95`), so I was reasoning about a superseded state — and I posted the recommendation without re-reading immediately beforehand. My "main's test coverage is a superset, no delta worth porting" claim was also wrong; 72 lines of these tests were additive and did land.

No action needed from anyone. My earlier note that **#2518 needs closing by hand** is also void — this PR carried `Fixes #2518`, so the merge closes it normally.

---

Issue #2527, opened 2026-08-28
Extract the config-drift finding contract into a reusable probe module


laurigates:
## What

Extract the finding / waiver / delta machinery currently inline in
`health-plugin/scripts/config-drift.py` into a stdlib-only sibling module, so
that a new corpus probe plugs into an existing contract rather than
re-implementing it.

Four pieces, all presently private to `config-drift.py`:

| Piece | Today | Wanted |
|---|---|---|
| Finding shape | dict literals with `severity` / `kind` / `paths` / `summary` scattered across seven `check_*` functions | one constructor that validates the shape |
| Waivers | `~/.claude/config-drift-waivers.json`, keyed by both sides' content hashes so an edit to either side revives the finding | reusable, path-parameterised |
| Delta fingerprint | absent — #2319 specifies `kind` + sorted paths but there is nothing to implement it in | `fingerprint(finding)` plus a previous-run set comparison |
| Exit codes / report render | `0 clean, 1 warn, 2 error, 3 internal`, plus the `=== SECTION ===` / `KEY=VALUE` / `STATUS=` render from `.claude/rules/structured-script-output.md` | shared renderer |

## Why

`#2319` is blocked on the third row. Its acceptance criteria are entirely
delta-reporting behaviour ("a second run with an unchanged corpus emits
nothing"), and there is no fingerprint implementation for that logic to live
in.

Three further probes are queued behind the same contract — agent-definition
overlap, hook double-gating (`offload-to-deterministic-substrate.md` warns
against it with no mechanical check behind the warning), and coverage-hole
detection. Each needs self-expiring waivers and delta reporting; none of them
needs `config-drift.py`'s similarity code. Splitting the contract from the
similarity computation is what lets those ship as small files.

The waiver design is the part worth preserving deliberately: keying on both
sides' content hashes is what stops a recurring report from decaying into a
backlog you learn to skip.

## How

- `health-plugin/scripts/lib/probe.py`, stdlib only. The `scripts/lib`
  **directory** convention exists, but every instance of it is shell
  (`macos-plugin/.../lib/common.sh`, `git-plugin/.../lib/load-config.sh`,
  `hooks-plugin/hooks/lib/drift-protocol.sh`); `scripts/lib/extract-md-elements.py`
  is hyphenated and executed as a subprocess, not imported. There is **no**
  in-repo precedent for a Python sibling import, so this one needs its own test.
- The import form is `from lib.probe import ...`, **not** `import probe`.
[...]

Issue #2319, opened 2026-08-08
Schedule the config-drift full pass with delta-only reporting


laurigates:
## What

Mount the expensive tier of `config-drift.py` (added in #2318) on a scheduled trigger, reporting **only the delta** since the previous run.

## Why it is not in #2318

Three reasons, all of which argue for landing it separately:

1. **Different repo.** The `.routines/` LaunchAgent pattern (`<name>.sh` + `com.lgates.<name>.plist` + `routines-status.sh`) lives in the portfolio repo, not in `claude-plugins`.
2. **Depends on a stable path.** The wrapper invokes the analyzer by path; that path should be settled by a merged PR before something schedules it. This is `tool-migration-cutover`'s rule applied to a new tool rather than a replacement — do not wire the trigger until the thing it triggers is actually in place.
3. **Different cost tier.** The SessionStart probe deliberately runs `--fast --no-embed` (0.05s). The scheduled run is the only place the embedding model is paid for.

## Shape

- Notify-only by default: run the full pass, emit nothing when there is no *new* finding.
- Delta, not backlog. A weekly report that re-lists the same 117 known findings is one you stop reading by week three. Compare against the previous run's finding set on a stable fingerprint (`kind` + sorted paths), so the delta survives reordering and rescoring.
- One rolling issue, commented on — not a new issue per week.
- Read-only. It surfaces; it never edits a rule, moves a file, or opens a PR (`drift-detection-triggering.md`: a probe surfaces, mutation needs an explicit invocation).
- Fail loudly on analyzer failure rather than reporting a clean sweep — a false all-clear is worse than no report.

## Acceptance

- [ ] A first run with no prior state records the baseline and stays silent.
- [ ] A second run with an unchanged corpus emits nothing (`STATUS=CLEAN`).
- [ ] Introducing one new duplicate rule produces exactly one new finding — the control that separates "nothing changed" from "the delta logic is broken and always reports zero".
- [ ] Analyzer failure (`rc=3`, or empty output) reports an error rather than a clean sweep.

## Also worth deciding when this lands

The first real run will surface a **large one-time backlog** — 74 `review_staleness` findings on the current corpus, most of which are a convention that was never enforced rather than genuine rot. Either bulk-waive the existing set as a baseline, or start `STALE_DAYS` high (365 → ~21 findings) and ratchet down. Turning it on at the current default without deciding this makes the first report unreadable, which is how a periodic audit gets ignored permanently.

Refs #2318

[...]

Issue #2141, opened 2026-07-27
refactor(skills): apply the demoted body-section mandates — ~103K tokens of required boilerplate


laurigates:
Follow-up to #2139 (context-engineering benchmark), finding **F2**. This partially **reverses** #1156 — see "Relationship to #1156" below before starting.

## The measurement

Two `skill-quality.md` mandates account for **13.5% of all skill body mass**:

| Mandate | Skills | Chars | ~Tokens | Share of body mass |
|---|---|---|---|---|
| `## When to Use This Skill` | **407 / 407** | 280,610 | ~70,000 | 9.2% |
| `## Agentic Optimizations` | 257 | 131,943 | ~33,000 | 4.3% |

Anthropic's [Claude 5 context-engineering post](https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models) locates when-to-use in the artifact's `description`, not in a body section, and has no requirement resembling `Agentic Optimizations`. The 2026-07 marketplace benchmark had already flagged both as house-only conventions unsupported by published guidance.

`.claude/rules/context-engineering.md` (landed in #2139) demotes them:

- `## When to Use This Skill` → **recommended where it disambiguates sibling skills**
- `## Agentic Optimizations` → **optional**; keep where the commands are the payload (CLI wrappers)

This issue is applying that decision to the corpus.

## This is not a blanket delete

The judges repeatedly **credited** the When-to-Use table where two or more sibling skills could plausibly match the same intent — that disambiguation is the load-bearing part. From the `claude-md` pair judgment, the table earned a win specifically because it *"disambiguates explicitly against near-miss sibling skills (meta-promote, session-distill, health-skill-audit) so an agent can route between overlapping artifacts."*

The test per skill: **does any sibling in this plugin compete for the same intent?**

- **Yes** → keep the table, and make sure the contrast column names real siblings
- **No** → drop it; the `description` already carries the trigger

Same shape for `Agentic Optimizations`: keep it on CLI-wrapper skills where the compact command table *is* the deliverable; drop it where it restates what the body already says.

## Relationship to #1156

#1156 standardised these tables across all plugins (162 of 298 skills at the time). That work was correct **for the model era it was written in**. This issue is not a retraction of it — it is the same question re-measured against Claude 5 generation guidance, where the cost is now legible and the benefit is narrower than "every skill".

Please do not close #1156 as wrong; reference it as the prior decision this supersedes.

## Sequencing

[...]

PR #2532, opened 2026-08-29
refactor(health-plugin): extract the config-drift finding contract into lib/probe.py

laurigates:
## What

**PR1 of the two-PR split you set out on #2527**: the finding / waiver / delta / render machinery moves out of `config-drift.py` into a stdlib-only sibling `health-plugin/scripts/lib/probe.py`, with **zero output change**. Per your 2026-08-28 comment — "PR1 extracts with zero output change; PR2 normalizes and inverts the assertions that pin the current bugs" — nothing here is normalized and no bug is fixed. Quirks are reproduced exactly, and listed below so PR2 has the inventory.

5 files, +668/−190 against the current `main`. `lib/probe.py` is 289 lines.

## Byte-identity

`--root . --fast --no-embed --format=json` at this repo's root: **1021 bytes, byte-identical** to `origin/main`, exit 0 and empty stderr on both. Re-verified after #2531 merged and `main` was merged into this branch.

That corpus reaches only 2 of the 9 construction sites, so 1021 bytes is a thin oracle on its own. A **15-case matrix** carries the rest: a planted fixture corpus in all four output formats, plus `waived`, `waiver-expired`, `--gate`, `--since`, the coverage-metric control gate, semantic-pass-unavailable, missing-root, and the repo corpus in all four formats. Timestamps normalized; everything else compared raw, including exit code and stderr. **All 15 hash identically** (`COMBINED=0bccda53…`), reaching **8 of 9** sites.

The 9th (`semantic_overlap_*`) needs `fastembed`, which is not installed in this environment, so its key shape is pinned by a unit assertion instead of the oracle. **If you can run the matrix on a machine with the embed deps, that closes the last gap** — it is the one claim here not backed by execution end to end.

## Verification

| Check | Result |
|---|---|
| `test-config-drift.sh` | **PASSED=93 FAILED=0 STATUS=OK** (this branch's 50 + #2531's 43) |
| `run-skill-script-tests.sh --only '…test-config-drift.sh'` | `STATUS=OK ISSUE_COUNT=0` |
| `ruff check` / `ruff format --check` (0.15.8) | clean |
| `lint-shell-scripts.sh` | 0 errors (9 pre-existing `macos-plugin` warnings) |
| `check-git-sandbox-guards.sh --strict` | `STATUS=OK ISSUE_COUNT=0` |
| `check-context-engineering.py --strict` | exit 0 |
| `pre-commit run --files <changed>` | exit 0 |
| `just lint-health` | all plugins OK |

Timing, median of 5 at repo root: **0.410 s → 0.417 s (+7 ms, +1.7%)** — the extra module import. Both sit above the 0.31 s in the issue, but that figure is from your machine; the before/after delta on identical hardware is the meaningful number. The BEFORE copy was checked to actually emit its 1021 bytes, so the comparison is not measuring a crash.

### The import mechanism, verified by execution

From cwd `/var`, under bare `python3` **and** `uv run --script`: `from lib.probe import …` resolves, `import probe` raises `ModuleNotFoundError`. No `__init__.py`. TEST 9 pins all three; TEST 10 pins stdlib-only-ness by importing with `site-packages` stripped from `sys.path`.

This matters because the failure mode is silent — `config-drift-probe.sh` treats empty output as "no findings", so a broken import is indistinguishable from a clean corpus.

### Mutation results — 17 mutations, all killed

Flat import; added `__init__.py`; added PEP-723 block; third-party import in the module; unvalidating `finding()`; reordered keys; `fingerprint` including `score` / not sorting / ignoring `kind` / ignoring `paths` / ignoring singular `path`; always-empty and always-full `delta`; hash-ignoring, a-side-only and never-suppressing `waived()`; non-escalating `--gate`; always-clean `exit_code`; `--no-embed` no longer skipping the semantic pass.

**Two defects in this change were found by that pass, before commit.** A `fingerprint`-ignoring-`kind` mutant survived 50/50, because the kind-discrimination assertion varied the path set too and was really re-testing the path half; it now compares two kinds over a *fixed* path set (`duplicate_rule_lexical` vs `semantic_overlap_rule_rule` on the same pair — a collision that genuinely occurs). Separately, a `$(...)` backup was stripping a trailing newline, which is fatal in a test whose subject is byte-level hashes.

The `--no-embed` poisoning test carries its control both ways: marker **absent** under `--no-embed`, **present** without it.

## Deliberately left for PR2

Reproduced exactly rather than fixed, so byte-identity stays the oracle:

- **The `path`/`paths` split.** `--since` filters on `paths` only, so `broken_pointer_stub` and `review_staleness` findings (which carry singular `path`) **survive a `--since` filter they should not**. Visible in the oracle — `fixture:since` keeps both. `finding_paths()` reads both spellings so `fingerprint()` is correct in the meantime.
- **`emit_status` omits `ISSUE_COUNT=` and `=== END CONFIG DRIFT ===`**, both required by `.claude/rules/structured-script-output.md`. Adding them is an output change.
- **`fingerprint()` is not wired into the analyzer's output path** — that would break byte-identity. Module-only, pinned by tests, ready for #2319 to consume.
- The `emit_probe` remediation map and `config-drift-probe.sh`'s jq map have **drifted**: the hook knows `coverage_metric_broken` and `semantic_overlap_rule_skill`, the Python map does not. Left as-is, but now single-sourced in `probe.REMEDIATION` so PR2 can reconcile both in one place.

## #2531 — merged, and its interaction resolved

**Resolved; no action needed.** #2531 merged, and the automated conflict resolution merged `main` into this branch (`3876715`). The merge preserved both sides exactly — `frontmatter()` here is byte-identical to `main`'s, `lib/probe.py` is intact, no conflict markers.

It did leave the suite red (67/26), for a reason neither change owns alone: #2531's TEST 10–12 load `config-drift.py` via `importlib.util.spec_from_file_location`, which does not put the script's own directory on `sys.path` — and that directory is exactly what PEP 420 uses to resolve this branch's new `from lib.probe import …`. Fixed in `be12858` by establishing it in the three harnesses (not in `config-drift.py`, which would be production code accommodating a test, and would mask the broken-import case TEST 9 exists to catch). Suite is back to 93/0 and CI is green. Full detail in the comment below.

## Caveats

- **`--help` output changed.** A paragraph was appended to the module docstring, which argparse uses as its `description`, explaining where the contract now lives. It is not analyzer output and `docs-currency.md` wanted it, but it is not byte-identical — flagging rather than burying it.
- One **Known Regressions** row was appended per `.claude/rules/regression-testing.md`. This PR fixes no bug, but the extraction *introduces* a new silent-failure class (the import), and that table is where this repo records what pins such a thing.

Refs #2527


laurigates:

laurigates:
#2531 merged, and the automated conflict resolution merged `main` into this branch as `3876715`. The auto-resolution was correct at the file level but left the suite **red — 67 passed / 26 failed**. Pushed `be12858` to fix it; CI is green again on that head (all 6 checks, `test` included).

## What broke

Not a bad merge. The merge preserved both sides exactly — `frontmatter()` here is byte-identical to `main`'s, `lib/probe.py` is intact, no conflict markers. Neither change was wrong alone; the **combination** is:

- #2531's TEST 10–12 load `config-drift.py` via `importlib.util.spec_from_file_location`.
- This branch added `from lib.probe import …` to `config-drift.py`, which PEP 420 resolves off the **script's own directory**.
- `spec_from_file_location` does not put that directory on `sys.path` the way running the script does. So every direct-call assertion died at `ModuleNotFoundError`, taking out all of TEST 10 (frontmatter parsing), 11 (the inline-`---` case) and 12 (the embedding cache) at once.

All 26 failures were in #2531's tests; TESTs 1–9 and 13–18 stayed green. Suite is now **93/0** (this branch's 50 + #2531's 43).

## The fix, and where it belongs

In the harness, not in `config-drift.py`. The harness is simulating "run the script", so it has to establish what running the script provides:

```python
sys.path.insert(0, os.path.dirname(os.path.abspath(analyzer)))
```

Teaching the analyzer to repair its own `sys.path` would be production code accommodating a test, and it would mask exactly the broken-import failure TEST 9 exists to catch.

## Re-verified against the merged `main`

| Check | Result |
|---|---|
| `test-config-drift.sh` | **93 passed / 0 failed** |
| `--fast --no-embed --format=json` vs `origin/main` | **byte-identical**, 1021 bytes, exit 0, empty stderr both sides |
| `ruff check` / `format --check` | clean |
| `lint-shell-scripts.sh` | 0 errors (9 pre-existing `macos-plugin` warnings) |
| `check-git-sandbox-guards.sh --strict` | `ISSUE_COUNT=0` |
| `check-context-engineering.py --strict` | exit 0 |
| `pre-commit` on the changed file | exit 0 |

The commit is typed `test(health-plugin)` — only the test file changed, so it publishes nothing.

One note on the `⚠️` in the compliance bot's three comments above: `health-plugin/skills/health-check/SKILL.md` is 16132 bytes on both `main` and this branch and is untouched by this diff. It's a recommendation on a check that passes, well under the 26000 ceiling, and already tracked by #2141 — no action taken.

---

laurigates:
**Superseded by #2536 — closing this.** Same issue (#2527 PR1), same new file, same design. #2536 is a strict superset, so reviewing both would be wasted effort.

I opened this at 06:23 when no open PR referenced #2527; #2536 arrived at 13:07 from a separate run. The duplication is mine to withdraw, not yours to adjudicate.

## Why #2536 rather than this one

Overlap is 4 of my 5 files, including `lib/probe.py` and `config-drift.py`. Everything this PR establishes, #2536 also establishes — including, independently, the **identical** fix for the `importlib` interaction: the same two `sys.path.insert` sites in `fm_get` and the TEST 12 driver, for the same stated reason. Two implementations converging on that separately is good evidence it is the right fix rather than a workaround.

On top of that, #2536 carries work this PR does not:

| | #2532 (this) | #2536 |
|---|---|---|
| Hook-level failure detection | — | `analyzer_failed` / `analyzer_bad_root` / `corpus_unreadable` in `config-drift-probe.sh` |
| Module test suite | via `test-config-drift.sh` | dedicated `test-probe-lib.sh`, 133 assertions |
| Byte-identity harness | 15-case matrix, run by hand | reusable `compare-ref-output.sh`, 68 root × format pairs |
| Second consumer | the test suite | `probe-delta.py`, a real one |
| CI wiring | — | `required-to-run-tests.txt` entries |

The hook-level row is the one that decides it. This PR treated the new silent-import-failure class as a **test** concern; #2536 also closes it at the hook, where `drift_emit` writing `"findings": []` is otherwise indistinguishable from a clean corpus. That is the more important half and this PR does not have it.

## One thing worth porting

#2536's file list does not include `.claude/rules/regression-testing.md`. That file's own pre-merge checklist wants a **Known Regressions** row for the silent-import-failure class the extraction introduces. #2536 more than satisfies the *substance* (that is what `test-probe-lib.sh` and the `required-to-run-tests.txt` entries are), so this is a bookkeeping row, not a gap in coverage — but it is a documented convention and cheap to add. The row from this branch is at `be12858` if it is useful as a starting point.

The branch `refactor/probe-module-2527` is left in place; nothing here is deleted, and this is reopenable if #2536 turns out not to land.

---

PR #2533, opened 2026-08-29
fix(hooks-plugin): scan the shell-only projection in bash-antipatterns-teach

laurigates:
## What

`bash-antipatterns-teach.sh` scanned the raw `$COMMAND`. It now scans the same `COMMAND_SHELL_ONLY` projection the sibling safety hook already computes — heredoc bodies stripped, then trailing `#` comments — in all 13 detector reads.

## Why

Every detector is `^`-anchored, and `^` matches the start of every **line**, so a heredoc-BODY line that merely *looked* like a watched command was nudged as if it were one.

The live report: the nudge fired on a `gh issue comment` containing **zero** shell pipes. The `|` characters it counted were markdown table cells inside the heredoc body, and a cell containing the word `printf` satisfied the discouraged-head test — the detector matched a table cell *describing another detector*.

### The audit found it in 5 of the 6 detectors

Probed against the shipped hook before the fix:

| Detector | Heredoc-body payload | Pre-fix |
|---|---|---|
| `long-pipeline` | markdown table | **FIRED** |
| `head` / `tail` | `head -50 notes.md` | **FIRED** |
| `find` | `find . -name "*.ts"` | **FIRED** |
| `grep`/`rg` | `grep -rn foo src/` | **FIRED** |
| `ls` | `ls *.md` | **FIRED** |
| `cat` | `cat README.md` | silent — **by accident**, exempted by its own unrelated `<<` clause |

**The exposure ran the other way too**, which the issue did not name. The read/search detectors disqualify themselves when the command contains a `|`, so a `|` anywhere in a heredoc body *silently suppressed a genuine hint* on the real command beside it. Two of the nine pre-fix test failures are these false negatives, so this is not merely a suppression.

## The design question, and how it was settled

The issue named one open choice — "a small shared helper or a lift of the existing projection".

The two awk programs are lifted into `hooks-plugin/hooks/lib/command-views.sh` (`strip_heredoc_bodies` / `strip_trailing_comments` / `command_shell_only`), sourced by the **teach hook only** and **fail-open**: a missing library exits 0 rather than erroring, because a teach nudge is never worth a broken hook.

`bash-antipatterns.sh` is left **byte-identical** and does not source it. Its own header establishes that the safety blocks must fire in *every* context and must not depend on something that can be missing — and a sourced file is such a dependency. Having both source the library was tested as viable (the full 231-case suite runs green here with `ast-grep@0.43.0`, the CI pin, installed), so this is a deliberate call rather than a limitation; the two copies are pinned behaviourally by their respective suites rather than textually. **Easy to redirect if you'd rather have the single definition** — say so and I'll make the safety hook source it too.

Ordering is preserved from the original: comment stripping runs *after* heredoc stripping, so it can only ever touch executable text, never a body line or its closing delimiter.

## Verification

The issue's A/B/C/D matrix, replayed against the shipped hook:

| Case | Want | Pre-fix | Post-fix |
|---|---|---|---|
| A — markdown table in a heredoc body | silent | **FIRED** | silent |
| B — the same table as a quoted `--body` arg | silent | silent | silent |
| C — genuine 6-pipe `cat`-headed scrape | FIRED | FIRED | FIRED |
| D — `echo hello` | silent | silent | silent |

**B is load-bearing**: the identical table text arriving as a quoted argument was already handled correctly, so the hook had the "this is data" concept and simply did not extend it to heredocs.

### Regression cases (15 new assertions)

`hooks-plugin/hooks/test-bash-antipatterns-teach.sh` gains a "heredoc bodies and trailing comments are DATA, not shell (#2518)" block: the A–D matrix, one heredoc-body case per detector (with `cat` pinned rather than fixed), and the trailing-comment half in both polarities — comment pipes must not inflate the pipe count, while a genuine 6-pipe scrape carrying a trailing comment must still fire.

**Guard integrity is weighted equally.** A projection that simply blanked the whole command would satisfy every `assert_silent` above, so three GUARD cases require a real `grep` to still fire: one sharing the heredoc-opening line, one after the terminator, and one on line 2 of a multi-line command — the last pins that per-line anchoring survives, which is the invariant #2431 established.

### Results

| Suite | Pre-fix | Post-fix |
|---|---|---|
| `test-bash-antipatterns-teach.sh` | **45 passed / 9 failed** | **54 / 0** |
| `test-bash-antipatterns.sh` (safety, untouched) | 231 / 0 | 231 / 0 |

Mutation-verified with the library and tests held constant — only the hook reverted — so the 9 failures are attributable to this change alone. `lint-shell-scripts.sh` reports 0 errors (the 9 warnings are pre-existing `macos-plugin` ones); `shellcheck` clean via pre-commit.

`.claude/rules/regression-testing.md` gains its Known Regressions row per the repo checklist.

## Note

Opened as a **draft** because this environment's standing instruction is to do so; the routine that produced it asks for ready-for-review. Flip it whenever you like.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GbT3UNDXV2J56SnXYBJmFN)_


laurigates:

PR #2550, opened 2026-08-31
feat(evaluate-plugin): sum the always-loaded surface across repos (--also)

laurigates:
## The gap

A per-repo C5 budget cannot see the cost a **session** actually pays. Measured 2026-08-31 across the three Claude-config repos:

| Repo | Always-loaded | Own gate |
|---|---|---|
| claude-plugins | ~23.0k tok | ✅ at **23%** of its 100,000-char budget |
| repos-claude-config | ~12.8k tok | ❌ none |
| dotfiles | ~9.2k tok | ❌ none |
| **stacked** | **~45.0k tok** | **nothing could fire** |

The every-session surface this dimension exists to bound was growing in the one place nothing measured. `--also` closes that.

## What it does

`--also PATH` (repeatable) sums the C5 surface — `CLAUDE.md` + rules with no `paths:` glob — across roots and gates the total with `--portfolio-budget`. Emits `C5_PORTFOLIO_*` plus a per-repo `C5_PORTFOLIO_BREAKDOWN`.

Only C5 aggregates. C1–C4 and C6 measure authoring quality inside one marketplace and stay scoped to `--project-dir`; summing them across unrelated repos would be meaningless.

Without `--also`, output is byte-identical to before — the portfolio keys are absent entirely, which is asserted.

## Three properties that keep the number trustworthy

Each is pinned by an executable test, and each test was **control-tested** — I broke the behaviour and confirmed the assertion fails, so none of them is a test that passes for the wrong reason.

1. **The breakdown sorts on every emitted field, not on the repo basename.** Two checkouts can share a basename (`~/a/config`, `~/b/config`), and a stable sort keyed on the name alone falls back to argument order — non-determinism in the one script whose headline contract is that two runs over an unchanged tree are byte-identical. I introduced this defect in the first pass and caught it re-reading the diff; it is fixed and tested.
2. **The portfolio ERROR sorts ahead of every WARN.** Otherwise `--max-issues 40` truncates away the only line explaining why `STATUS=ERROR` — a red run with no visible cause.
3. **`--also` and `--target` are rejected together.** `--target` drops rules and `CLAUDE.md`, which is the *entire* C5 surface, so the total would be silently wrong rather than absent.

`always_loaded_surface()` deliberately does not call `discover()`: that walks the tree for `SKILL.md`, and an `--also` root may contain large unrelated checkouts (the portfolio root has `external/`).

## The rule change

`context-engineering.md` gains the distinction that decided the scoping work in the sibling repos:

> **`paths:` fires only when a matching file is EDITED — it cannot scope a rule whose hazard fires on a command.**

Getting that backwards silently disarms a guard while the budget number improves. The rule now sorts candidates by *what triggers the failure* — edit / command / intent / every turn — and says to split a rule that mixes them, which is what the companion dotfiles PR does to a 20 KB `chezmoi-conventions.md` whose apply-time hazards can fire without touching any source file.

## Verification

- `tests/test-check-context-engineering.sh`: **42 pass, 0 fail** (was 24 — 18 new assertions).
- `run-skill-script-tests.sh --only '*evaluate-context-engineering*'`: `STATUS=OK`, discovered and passed.
- Full suite exits 0 (5 pre-existing `required_test_skipped` ERRORs from tools absent in this sandbox: ast-grep, cargo-generate, task CLI — unrelated to this change).
- `ruff check` + `ruff format --check` clean; this repo's own C5 ratchet still exits 0.

Companion PRs: laurigates/dotfiles#392 (scoping + the chezmoi split), laurigates/repos-claude-config#35 (the `just context-budget` wiring, which consumes these flags).

https://claude.ai/code/session_016rTMFif8HeRJqG8yX85Su6

---
_Generated by [Claude Code](https://claude.ai/code/session_016rTMFif8HeRJqG8yX85Su6)_

### DIFF
--- a/.claude/rules/context-engineering.md
+++ b/.claude/rules/context-engineering.md
@@ -1,7 +1,7 @@
 ---
 created: 2026-07-26
-modified: 2026-07-26
-reviewed: 2026-07-26
+modified: 2026-08-31
+reviewed: 2026-08-31
 paths:
   - "**/SKILL.md"
   - "**/skills/**"
@@ -63,6 +63,24 @@ unscoped rule is the exception and should say why. If it reads as a procedure
 with a trigger ("when releasing", "before a destructive op"), it is a skill —
 promote it via `agent-patterns-plugin:meta-context-diet`.
 
+**`paths:` fires only when a matching file is EDITED — it cannot scope a rule
+whose hazard fires on a command.** This is the trap that decides most scoping
+calls, and getting it wrong silently disarms a guard while the budget number
+improves. Sort the rule by what triggers its failure, not by what it talks
+about:
+
+| The failure happens when you… | Lever | Example |
+|---|---|---|
+| edit a file of a known shape | `paths:` glob | zsh keybindings, `.gitattributes`, chezmoi naming prefixes |
+| run a command | **stays unscoped**, kept lean, detail in a linked doc | `chezmoi apply` deleting unmanaged files; `re-add` skipping templates |
+| act on intent ("diagnose CI", "release") | promote to a **skill** | `ai-review-max-turns`, `multirepo-ci-cd` |
+| any turn, unprompted | stays unscoped — it is a real invariant | never merge a stranger's PR |
+
+A rule that talks about chezmoi is not automatically edit-triggered: an apply
+that destroys data can touch no source file at all. When a single rule mixes
+both (dotfiles' 20 KB `chezmoi-conventions.md` did), **split it** along that
+line rather than scoping the whole thing.
+
 **Guidance lives in exactly one place, and the others name it.** Cross-plugin
 means a `plugin:skill` name reference, never a shared `REFERENCE.md`
 (`skill-consolidation.md`). If you catch yourself restating a table that another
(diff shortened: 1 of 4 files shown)


laurigates:
The compliance ⚠️ on `evaluate-plugin/evaluate-improve` is **pre-existing and not from this PR** — flagging it so a reviewer doesn't chase it.

Verified rather than assumed:

- `evaluate-improve/SKILL.md` is byte-identical on `origin/main` and on this branch (11,272 chars each way). This PR's diff is four files, all under `evaluate-context-engineering/` plus `.claude/rules/context-engineering.md`.
- Last commit touching it is `f5e000c` (2026-08-21), ten days before this branch.
- It surfaces here only because the compliance check reports per *plugin*, and this PR touches `evaluate-plugin`.

Not fixing it in this PR, deliberately. It is a ⚠️ at 11.2k against a 26,000-char ceiling, and restructuring an unrelated 11k skill body would widen a PR that is otherwise a scanner change plus its tests.

Worth noting the finding is also explicitly deprioritized by the rule this PR strengthens — `context-engineering.md`'s cost model ranks by `always-loaded cost × frequency`, and a `SKILL.md` body is paid only when the skill fires:

> The post's ">80% removed, no measurable loss" figure is about an always-loaded system prompt. A 20,000-char skill body is a smaller problem than a 5,000-char unscoped rule.

So this advisory is real but low-priority by our own measure. If it's worth doing, it's a separate PR against `evaluate-improve` — happy to open one.

---

Issue #1973, opened 2026-07-05
Consider extracting vault-agent into standalone repository


laurigates:
## Summary

Extract `vault-agent` into its own repository (`laurigates/vault-agent`), the same way `git-repo-agent` was extracted in #1017. `vault-agent` is a standalone Claude Agent SDK / Typer CLI that lives in this monorepo but is not a Claude Code plugin (absent from `marketplace.json` and release-please). Extracting it enables clean `uvx vault-agent` / `uv tool install vault-agent`, independent versioning, and PyPI publishing.

This is the direct sibling of #1017 — reuse the proven template there ([`laurigates/git-repo-agent`](https://github.com/laurigates/git-repo-agent), PR #1, and this repo's sync workflow PR #1972).

## Motivation

Same rationale as #1017:

- `uvx vault-agent` / `uv tool install vault-agent` without a `claude-plugins` checkout.
- Clearer separation of concerns and independent release cadence for a self-contained CLI.
- Consistency with `git-repo-agent`, now extracted.

## Current coupling

`vault-agent/src/vault_agent/prompts/compiler.py` reads Obsidian-plugin skills from the monorepo at runtime:

```python
_REPO_ROOT = _MODULE_DIR.parent.parent.parent  # vault-agent/
_PLUGINS_ROOT = _REPO_ROOT.parent              # claude-plugins/
# sources: obsidian-plugin/skills/vault-{frontmatter,tags,templates,wikilinks,orphans,stubs,mocs}/SKILL.md
```

This is ADR-0005 ("skill-source-obsidian-plugin") — analogous to `git-repo-agent`'s compiler reading its sibling plugins.

**Key difference from #1017:** `git-repo-agent` had *already* solved the standalone-install problem (ADR-009: a compiler fallback that ships pre-compiled `generated/` artifacts). **`vault-agent` has not** — its `compiler.py` has no standalone mode and no committed `src/vault_agent/prompts/generated/`. So the extraction has an extra **Phase 0** prerequisite that `git-repo-agent` did not.

## Proposed approach

Mirror #1017's phased, verify-before-remove cutover (per this repo's `tool-migration-cutover` rule), with the added Phase 0:

### Phase 0 — standalone-install prompt fallback (prerequisite; the ADR-009 analog)
[...]

Issue #1017, opened 2026-04-09
Consider extracting git-repo-agent into standalone repository


laurigates:
## Summary

git-repo-agent is a standalone Claude Agent SDK CLI tool that lives in this monorepo but is not a Claude Code plugin (not in marketplace.json or release-please). Extracting it into its own repo would:

- Enable clean `uvx git-repo-agent` and `uv tool install git-repo-agent` workflows
- Provide clearer separation of concerns
- Simplify independent versioning and release management

## Current Coupling

The only coupling to this monorepo is `compiler.py`, which reads 35 SKILL.md files from 8 sibling plugins at runtime to build subagent prompts. The compiled output is 7 files totaling ~328KB.

## Speculative Approach: Pre-compiled prompts with optional live compilation

Ship pre-compiled prompt files in the package. When running inside the monorepo (dev mode), compile fresh from SKILL.md sources. When installed standalone, read from the `generated/` directory. The generated files are derived artifacts, not source duplication.

### Why not alternatives

- **Git submodule**: uv/uvx does clone submodules (with absolute URLs), but it would pull in the entire claude-plugins monorepo — heavyweight for 328KB of derived content
- **Separate PyPI package for prompts**: over-engineered for 328KB of static markdown
- **Runtime download**: adds network dependency and failure modes to CLI startup

### Key changes

1. **Modify `compiler.py`** — add fallback to read from `generated/` when plugin sources aren't available:

   ```python
   @lru_cache(maxsize=None)
   def get_compiled_prompt(subagent_name: str) -> str:
       skill_paths = SUBAGENT_SKILLS.get(subagent_name)
       if not skill_paths:
           return ""

       # Live compilation when running inside claude-plugins monorepo
       if _PLUGINS_ROOT.is_dir() and any(
           (_PLUGINS_ROOT / p).exists() for p in skill_paths
       ):
           return compile_subagent(subagent_name, skill_paths)

       # Standalone mode: read pre-compiled generated files
       generated = _MODULE_DIR / "generated" / f"{subagent_name}_skills.md"
       if generated.exists():
           return generated.read_text(encoding="utf-8")

       return ""
   ```

2. **Create `laurigates/git-repo-agent` repo** with contents of `claude-plugins/git-repo-agent/` as root

3. **Update `pyproject.toml`** with license, readme, repository URL for PyPI publishing

[...]

PR #2566, opened 2026-09-02
refactor: promote the two orphaned agent-CLI rules to plugin skills

laurigates:
Stacked on #2562 (base: `chore/remove-extracted-agents`) — retarget to `main` after that merges.

#2562 leaves two path-scoped rules whose globs match nothing in this repo, because the CLIs they were written beside are gone. Retargeting them to `**/*_agent/**`, as #2562 does, keeps the content somewhere it can never fire. This moves it to where the audience actually is.

## The audience is not this repo

Measured across 53 `pyproject.toml` under `~/repos/laurigates/`:

| Repo | Typer | agent-sdk | worktrees | `typer-cli-completion` | `agent-cli-worktree-safety` |
|---|---|---|---|---|---|
| `silverbucket-cli` | ✓ | — | — | **✓** | — |
| `silverbucket-mcp` | ✓ | — | — | **✓** | — |
| `vault-agent` | ✓ | ✓ | ✓ | **✓** | **✓** |
| `git-repo-agent` | ✓ | ✓ | ✓ | ✓ | ✓ |
| `hw-diag-agent` | — | ✓ | none found | — | — |
| `claude-plugins` | — | — | — | — | — |

`typer-cli-completion` has three live consumers today and none here. `silverbucket-cli` has a `[project.scripts]` Typer entry point, so it hits the exact `Shell None is not supported` failure the rule documents, and the rule has never fired for it — it was only ever co-located with `git-repo-agent`.

`hw-diag-agent` uses the agent SDK but I found no worktree code in it, so it is not claimed as a worktree-safety consumer.

## What moved

| From | To |
|---|---|
| `.claude/rules/typer-cli-completion.md` | `python-plugin/skills/typer-cli-completion/` |
| `.claude/rules/agent-cli-worktree-safety.md` | `agent-patterns-plugin/skills/agent-cli-worktree-safety/` |

Genericised in the move, since a marketplace skill reaches people with no `laurigates` context:

- The shellingham skill loses the portfolio framing and keeps `git-repo-agent`'s implementation as a cited example rather than as the definition of scope. The failure table gains the general shape — *anything that generates completions non-interactively*, which is exactly when you want them.
- The worktree skill describes "an agent SDK" and "an interactive tool call" instead of naming `claude-agent-sdk` and `AskUserQuestion` as the only instances, and surfaces the diagnostic tell (a run that wrote files and reported "No changes were made") up front.

Every concrete invariant, code block, and required-test row survives. Two additions, both from re-reading the originals: `--porcelain` rather than `git diff` is called out as load-bearing (a newly created file is untracked, so `git diff` cannot see it), and the safety-net commit is required to print a warning, because a silent safety net hides the bug it compensates for.

## Two deliberate omissions

**No "When to Use" section on either skill.** `context-engineering.md` demoted it to *recommended where sibling skills compete for the same intent*; neither of these competes with anything, so the description carries the trigger. `plugin-compliance-check.sh` emits a ⚠️ and exits 0 for exactly this case — that is the two remaining warnings on this branch, and there are zero ❌.

**Only the Typer skill gets a pointer stub.** A path-scoped stub lands separately in [`laurigates/repos-claude-config`](https://github.com/laurigates/repos-claude-config) at `laurigates/.claude/rules/typer-cli-completion.md`, scoped to the four consumer repos, so the skill surfaces when you are in one of them. `agent-cli-worktree-safety` is catalog-discovery only, by decision — nothing auto-fires for it in `vault-agent`. Say the word and it gets the same three-line stub.

## Verification

| Gate | Result |
|---|---|
| `plugin-compliance-check.sh` | 0 ❌, 2 ⚠️ (both the "When to Use" recommendation above) |
| Skill body size | 6.0K / 8.0K chars — both `OK`, well under the 10K band |
| Skill descriptions | 197 / 200 chars — both within the 200-char target band |
| `check-docs-index.sh --strict` | `STATUS=OK`, `DIAGRAM_SVG_NODES == DIAGRAM_NODES` |
| `lint-mcp-tool-references.sh` | clean |
| `check-delegation-reachability.sh` | `STATUS=OK` |
| pre-commit (full run, via `git commit`) | passed |

Catalog counts refreshed with `/docs-refresh`: agent-patterns 21→22, python 16→17, across README, PLUGIN-MAP, and the d2 diagram with its re-rendered SVG committed alongside.

Refs #1017
Refs #1973

https://claude.ai/code/session_018p79wKVj4hCd19kTWmzmQx


laurigates:
Automatic conflict resolution could NOT complete — the merge has been aborted and the branch is unchanged. Manual resolution is needed.

Conflicts and the correct resolution:

- .claude/rules/agent-cli-worktree-safety.md — DU (deleted on this branch, edited on the base branch): accept the deletion.
- .claude/rules/typer-cli-completion.md — DU (same): accept the deletion.
- CLAUDE.md — the rules-table row for agent-cli-worktree-safety.md was added on the base branch and removed here: drop the row, since the rule file is gone.

Both sides agree in substance. The base branch (#2562) retargeted the two rules' paths globs and rewrote their intra-repo links as GitHub URLs now that git-repo-agent/vault-agent live in their own repos; this PR promotes the same content to agent-patterns-plugin/skills/agent-cli-worktree-safety/SKILL.md and python-plugin/skills/typer-cli-completion/SKILL.md. I verified the two new SKILL.md files already carry the base branch's URL rewrites (git-repo-agent ADR-003, test_worktree_changes.py, test_worktree.py, main.py, test_completion.py), so accepting the deletions loses nothing from the base branch.

Why it was aborted: staging the deletions requires removing files under .claude/rules/, and every route to that (git rm, git update-index --force-remove, rm) is blocked by this environment's sensitive-path protection. CLAUDE.md resolved cleanly, but a partial resolution is not committable, so the merge was aborted rather than left half-done.

To finish manually: run the merge, then 'git rm' the two rule files, delete the whole conflict block in CLAUDE.md (both markers and the agent-cli-worktree-safety row), 'git add CLAUDE.md', and 'git commit --no-edit'.

No untrusted instructions were found in the conflicted content.

PR #2575, opened 2026-09-02
ci(audits): fix five false-positive generators behind the monthly scheduled reports

laurigates:
Triage of the six open issues filed by scheduled workflows (#2555, #2556, #2557, #2573, #2507, #2494). Every claim was verified against the repo before acting; the result is that most of the reported findings are generator defects, so this PR fixes the generators rather than remediating their output.

## What was wrong

| Report | Reported | Actual |
|---|---|---|
| #2555 Infra Compliance | 🔴 token leak in `test-skill-scripts.yml` | The regex alternative `sk-` matches the literal `task-`. All three hits are YAML comments (`task-add`, `test-task-id-stability.sh`, `task-3.4.2.tar.gz`); the file references no secret at all. A second instance in `workflow-model-audit.yml:142` went unreported. |
| #2555 Infra Compliance | `stranded-work-audit.yml` missing path filters | Cron-only workflow — it structurally cannot have them. Flagged for a prose comment. Meanwhile `plugin-pr-checks.yml` **falsely passed** on comments explaining why it deliberately has none. |
| #2556 Blueprint Health | 171 stale skills, "update the `modified` date" on each | 110 of 167 share one `modified:` date. Following the recommendation literally writes a fresh date into 167 files — manufacturing next month's report. |
| #2557 Agentic Audit | 173 missing Agentic Optimizations, 2 `model: haiku` | Rec 1a declined (the rule prescribes a filter, not abolition); Rec 1b is void — there is no `model:` scraper, model appraisal is a prompt judgement step. |
| #2507 Research Radar | Amend a technique catalog in `prompt-engineering-plugin` | No such catalog. The prompt forbids reading plugin source while asking the model to map papers onto plugin *names* — the hallucination is structural. |

## Changes

**`scripts/infra-compliance-check.sh`** — length-anchor `sk-` and `Bearer` in the token detector; make the path-filter check trigger-aware (yq node query with a comment-stripping fallback, and a fallback that survives a present-but-broken `yq`); scope the `paths:` lookup to the `pull_request` trigger so a `paths:` on another trigger or inside a `run: |` block no longer produces a false pass; add `--project-dir`. Security posture 14/15 → 15/15, overall 92 → 93.

**`scripts/blueprint-health-check.sh`** — label the metric with the field it measures, roll the stale table up by `modified:` date, itemise only the >180d tail (11 rows), and rewrite the recommendation so it names the tail count and the dominant cohort instead of ordering 167 date bumps. Report length 237 → 109 lines.

**`scripts/agentic-audit-precompute.sh`** (new) — the precompute extracted out of inline YAML in `scheduled-audits.yml` so it is testable; all four `find` walks prune `.claude/worktrees`; adds the stale-review date-cohort section the report itself asked for (256 stale across 46 distinct dates, largest cohort 109). Output verified byte-identical to the old inline block apart from the new section. Flagging predicates deliberately unchanged.

**`.github/workflows/research-radar.yml`** — replace *"Do not read plugin source files"* with a bounded token-level grounding requirement: one `Grep` per surfaced paper against the concrete surface the suggestion asserts exists, drop or re-target on no hit, state the pattern in the issue body. A path-existence probe would not have caught the W35 hallucination — the plugin directory does exist.

**`scripts/check-workflow-model.sh`** — an unknown flag was swallowed into the explicit-files list, so `--strict` scanned zero workflows and printed a passing report. Now rejected with exit 2 (the #2057 class). No caller passes flags.

**`rust-plugin/skills/mockito-http-mocking/SKILL.md`** — the one genuinely actionable finding in #2556: add `reviewed: 2026-07-03`, back-dated to the authoring date rather than today. `modified:` deliberately not bumped.

## Testing

Five new suites plus additions to two existing ones. Every assertion was mutation-verified — the guarded change was reverted and the suite confirmed to go red.

```
test-agentic-audit-precompute       22 passed
test-blueprint-health-report        39 passed   (fixture-driven, 3 mktemp trees)
test-infra-compliance-check         33 passed   (7 filter fixtures × 3 yq states, paired token polarities)
test-check-research-radar-grounding 15 passed
test-audit-scripts-exit             17 passed
test-check-workflow-model           43 passed
pre-commit run --files <18 changed> 34 hooks passed, 0 failed
```

All six new suites are wired into `.pre-commit-config.yaml` and `plugin-pr-checks.yml`, and the three defects are recorded in `.claude/rules/regression-testing.md`.

## Deliberately not done

- **Bulk `reviewed:` / `modified:` date bumps** — the noise source, not the fix.
- **Dropping the Agentic Optimizations check** — `context-engineering.md:112` says "Optional; keep it where the commands are the payload", which prescribes a filter. Deleting it would also strand two referents in the same workflow.
- **Adding path filters** — a path-filtered required check never reports and wedges the merge (#2258). The two affected workflows get an explicit `paths-exempt` marker instead.
- **The `actions/checkout` v4 gate** (`infra-compliance-check.sh:104`) — warns on all 27 refs precisely for being at `@v6`. Real, but out of this PR's verified set; raised on #2555.
- **Anything on #2494** — the 13 ComfyUI pack repos are outside this session's scope, so all 77 findings are unverifiable here.

Comments recording the corrections have been posted on each issue so the next scheduled run does not re-surface them.

https://claude.ai/code/session_01N3EW9FWZuTCf31YvyghqLC

---
_Generated by [Claude Code](https://claude.ai/code/session_01N3EW9FWZuTCf31YvyghqLC)_


laurigates:
CI green on `a377d04` — 8 checks pass, `doc-audit` skipped, no conflict.

**The `compliance` failure and its fix.** `test-agentic-audit-precompute.sh` failed on the runner while passing locally, on the one assertion proving the stale list is populated. `section()` passes boundary headings to awk as *dynamic* regexes, and the stale-list boundary was written `'^## Skills with stale reviews \(>90 days\):$'`. A backslash-escaped paren is undefined in POSIX awk, so builds disagree: the runner's awk warned ``escape sequence `\(' treated as plain `('`` and then read the parens as a **group**, so the heading never matched, the section sliced empty, and the assertion failed. Reproduced before fixing, with a shim stripping the escapes before handing off to the real awk — pre-fix `Passed: 21, Failed: 1` with the identical message, post-fix 22/0.

Also closed the direction that hid it: an empty slice **silently satisfies** every "must NOT contain" assertion over that section, so a mis-slice weakens the suite in a way no failure reports. All four sections are now asserted non-empty before any assertion runs, failing with the boundary that did not match.

One correction to the auto-fix commit I merged (`08050e0`), since its comment is now in the tree: it attributes the split to GNU-vs-BSD awk and says the bug is "invisible on macOS". That does not match the measurement — this box and the runner are both Linux, and mawk 1.3.4 20240123 tolerates the escape while the runner's older mawk does not. It is a build difference, not a platform one; the merged comment says so and drops the unverified platform claim. The fix itself was identical on both sides.

**The three `rust-plugin` ⚠️ from Plugin Compliance Review: not fixing here, deliberately.** All three are pre-existing on `main` and untouched by this PR — verified against `origin/main`: `cargo-worktree-builds` and `mockito-http-mocking` descriptions are already over the band, and `cargo-llvm-cov/SKILL.md` is byte-identical at 10,337 chars. This PR's only change under `rust-plugin/` is a single `reviewed:` frontmatter line, which is what scoped the compliance run to that plugin and surfaced them. They are advisory (⚠️, not ❌) and the `compliance` check passed. Rewriting two descriptions and extracting a 10K skill body is real content work in a plugin this PR touches with one metadata line, so it belongs in its own change rather than widening this one.

---

Issue #2574, opened 2026-09-02
feedback(hooks-plugin): no-calendar-estimates fires on measured hardware throughput, and offers no unit that fits


laurigates:
## What happened

The `no-calendar-estimates.sh` Stop hook blocked a response for saying a GPU render pass would take **"roughly 70–80 minutes"**.

That figure was not an estimate of agent work. It was extrapolated from a directly measured rate: one segment rendered 90 frames in 1m31s (≈ 1.0 s/frame), times the piece's 3870 frames.

## Why the guidance did not fit

The hook's message offers these units:

> tokens consumed, context-window share remaining, effort tier (low / medium / high / max), tool-call count, or files / lines to touch

None of them can express how long a diffusion model takes to render a fixed number of frames on an RTX 4090. That quantity is wall-clock by nature — it is a property of the hardware, not of the agent. The restatement had to become "3780 frames at a measured 1.0 s/frame" and leave the reader to do the multiplication, which is strictly less useful than the number they wanted.

## Why the rule is still right

The underlying rule is good and I do not want it relaxed. Agent-effort estimates in calendar time are dishonest — the model has no clock and no idea how long a human will wait. The problem is narrower: the matcher does not distinguish

| | example | honest unit |
|---|---|---|
| **(a) agent effort** — what the rule targets | "this refactor will take me about an hour" | effort tier, tool calls, files touched |
| **(b) external machine work** — measured or extrapolated | "the 3870-frame render takes ~75 min at 1.0 s/frame" | wall-clock, because that is what it is |

(b) also covers CI runs, large builds, model downloads, long test suites — anything where the agent is *reporting a measurement* rather than *predicting its own pace*.

## Suggested fix, in preference order

1. **Extend the message** to name the external-job case and say how to phrase it (rate × quantity, or "measured at N s/unit"). Cheapest, and it removes the dead end where the offered units cannot express the fact.
2. **Narrow the match** — e.g. don't fire when the duration sits near a measured rate, a frame/row/file count, or words like `render`, `CI`, `build`, `download`.

Option 1 alone would have been enough here: the block was recoverable, it just cost a round-trip and produced a worse sentence.

## Environment

- Session 2026-09-02, repo `laurigates/comfyui-nodes`, during a film delivery.
- Blocked text: `roughly 70–80 minutes for 3870 frames`.
- Preceding measurement in the same session: one segment, 90 frames in 1m31s.

Issue #2306, opened 2026-08-07
feedback(hooks-plugin): git-stash-reminder reports pre-existing stashes as session-created when cwd changes repo


laurigates:
## Skill

`hooks-plugin/hooks/git-stash-reminder.sh` (observed on 2.8.1; same logic present in 2.5.13–2.8.1)

## Category

Bug

## Description

The Stop hook reported a stash created **2025-03-31** as "created during this session", and did so as a *blocking* Stop-hook error — twice in one session, in a session that performed zero git writes.

Two linked defects produce it:

**1. The baseline is repo-scoped to SessionStart cwd, but compared against Stop-time cwd.**

`git-stash-session-init.sh` writes the baseline once at SessionStart for the then-current directory. `git-stash-reminder.sh` later runs `git -C "$CWD" stash list` at Stop time. If the cwd has moved into a *different* git repository in between — a nested checkout, a sibling repo, a vendored tree — the stashes found at Stop belong to a repo the baseline never described, so every pre-existing stash there is classified as new.

This is what happened here. SessionStart cwd was a workspace root whose repo has no stashes; the baseline was therefore written empty. During the session the cwd moved into a nested, separately-cloned repo one level down, which has a 16-month-old stash. That stash was reported as session-created.

**2. An empty-but-present baseline file passes the missing-baseline guard.**

```sh
# lines 46-50
if [ -z "$SESSION_ID" ] || [ ! -f "$BASELINE_FILE" ]; then
    exit 0
fi
```

The guard exits silently only when the file is **absent**. A file that exists but is empty proceeds, and the skip test then never matches:

```sh
# line 62
if [ -n "$BASELINE_HASHES" ] && echo "$BASELINE_HASHES" | grep -qF "$hash" 2>/dev/null; then
```

With `BASELINE_HASHES` empty, `[ -n ... ]` is false, so no stash is ever skipped and all are counted new. An empty baseline is indistinguishable from "no stashes existed at session start", but it is also exactly what a repo-with-no-stashes produces — so defect 1 reliably feeds defect 2.

## Evidence

Session `2e424766-…`, 2026-08-07. Hook fired twice, both blocking:

```
Found 1 git stash(es) created during this session in /…/ComfyUI. Review before exiting:
  stash@{0} (11848h 57m ago): WIP on master: 8edc1f44 Support more float8 types. → git stash pop
```

[...]

PR #2584, opened 2026-09-03
ci(hooks-plugin): wire the ten unwired hook regression suites into pre-commit

laurigates:
## What

Adds a `.pre-commit-config.yaml` entry for each `hooks-plugin/hooks/test-*.sh` suite that had none — ten of the twelve. Each is modelled on the existing `test-bash-antipatterns` entry (`language: system`, `pass_filenames: false`, a `files:` regex naming the hook under test and its suite).

## Premise correction

The task described these suites as *"never executed by CI or pre-commit."* **The CI half is false.** `scripts/run-skill-script-tests.sh` already discovers all 13 through its `*/hooks/test-*.sh` glob, and `.github/workflows/test-skill-scripts.yml` fires on `*-plugin/hooks/**`:

```
$ bash scripts/run-skill-script-tests.sh --only 'hooks-plugin/hooks/test-*.sh'
PASS=./hooks-plugin/hooks/test-no-calendar-estimates.sh
...
TOTAL=13  PASSED=13  FAILED=0  STATUS=OK
```

The premise is an easy one to reach and worth recording: a repo-wide grep for a suite's **filename** finds only its own usage comment, because discovery is glob-based. Every glob-discovered test reads as entirely unwired.

The real gap was **local-only** — the same shape as #2141 (developer friction invisible to every green PR), not the #2164 / #2219 / #2221 class where a guard genuinely had no CI signal. Editing a hook gave no feedback until after a push.

## Non-obvious pairings

Three suites do not map to their hook by name, and were resolved by reading the suite rather than by a mechanical `s/test-//`:

| Suite | Also triggers on | Why |
|---|---|---|
| `test-bash-antipatterns-teach.sh` | `lib/command-views.sh` | Sourced by the hook under test (#2518) |
| `test-git-stash-reminder.sh` | `git-stash-session-init.sh` | Drives both halves of the pair (#2306) |
| `test-test-verification.sh` | `test-verification.sh` | The **hook** is named `test-verification.sh`; neither is the other's typo |

## Verification

- **Per entry, not in aggregate:** running pre-commit against each individual hook file fires exactly its own suite, for all eleven — including `git-stash-session-init.sh`, whose trigger belongs to a differently-named suite.
- `pre-commit validate-config` passes; `pre-commit run --files .pre-commit-config.yaml` is clean.
- All 13 suites pass on the current head. Slowest is `test-external-pr-merge-guard.sh` at 5.5s; every regex is narrow, so the local cost lands only on a commit touching that hook.

## Relationship to #2574

When this branch was cut, #2574's fix was unmerged and the calendar suite was 26 assertions. Composition was checked rather than assumed by running that branch's hook and suite together: **33 passed, 0 failed**.

That fix has since landed on `main` as #2579 and is now merged into this branch, so the point is no longer hypothetical — the merged tree runs **33 passed, 0 failed** through the new entry. The 33rd assertion is the control pinning `PATTERN_FUTURE` / `PATTERN_MARKER` against silent narrowing; it is now wired rather than merely anticipated.

## Acceptance

```
$ pre-commit run --files hooks-plugin/hooks/no-calendar-estimates.sh
no-calendar-estimates hook detection regression (#2574)......................Passed
```

## Merge history note

`main` was merged in twice — once by me and once by the repo's conflict automation, independently. Because `.gitattributes` marks `.claude/rules/regression-testing.md` `merge=union`, both copies of the row this PR adds survived the second merge (the stale pre-#2579 wording alongside the corrected one). The stale copy is removed in `41c6f6d`; the file is now `main` plus exactly one row, verified by row count and a duplicate-line scan.

Refs #2574

https://claude.ai/code/session_01VQGrBHKZXhTNdWetxrnF5v


laurigates:

PR #2561, opened 2026-09-02
fix(plugins): adapt rules, skills, agents, and guards for Claude Fable 5.1

laurigates:
## Summary

Sweep of this repo against Claude Fable 5.1 (released 2026-09-01). Sources: the Fable 5.1 / Mythos 5.1 system card, the platform release notes, the Claude Code changelog 2.1.232–2.1.257 and the model-config / sub-agents / skills / tools-reference docs. Method and decisions are recorded in `docs/audits/fable-5.1-adaptation-2026-09-02.md` (15 scoped finders, two adversarial verifier lenses per finding, completeness critic; 132 confirmed findings for this repo).

Merged with `main` after the git-repo-agent and vault-agent CLIs were removed there (87f2b3e): the eight modify/delete conflicts were resolved by accepting the removal, so the `fix(git-repo-agent)` commit in this branch's history now contributes nothing to the diff.

## What changed

- **Model policy stated for the current generation.** `model: opus` stays the committed floor for plugin agents; `fable` is sanctioned for the hardest delegated reasoning and accepted by `check-agent-model.sh`; `effort:` (now a frontmatter field on agents and skills) is the cost lever and the guard validates it. The Opus 4.8 / Sonnet 4.6 figures are labelled as the generation they were measured on, with the Tier 2 sweep named as the re-verification.
- **Harness currency.** `fable`/`best` aliases, `xhigh`, `PreModelSwitch`/`PostModelSwitch`, background-by-default spawns and the `fork` subagent type, auto mode as the built-in starting mode, subagent model resolution order and `CLAUDE_CODE_SUBAGENT_MODEL_FORCE`, `experimental.cacheTtl`.
- **Task-tool availability (2.1.233+).** One canonical note in `agentic-permissions.md`; every other mention points to it. The `--allowedTools … TodoWrite …` grants in workflows are kept: on current models that grant is the opt-in. The first verified draft called them inert; the tools reference corrected that, and the audit record says so.
- **Fable behaviour encoded as rules.** A hook block is on the hazard, not the spelling; never restate authorization the user did not give (system card §6.2.1); subagent briefs quote the user verbatim; reports ground status claims in tool output; goal + constraints + verification over step ladders for judgment work.
- **`[1m]`-conditioned guards** reworded for 1M-default sessions (they read as "does not apply" on Fable 5.1).
[...]

Issue #2333, opened 2026-08-08
run-skill-script-tests.sh does not discover repo-root scripts/tests/, so guards there need hand-wired CI


laurigates:
## What

`scripts/run-skill-script-tests.sh` discovers tests with three globs:

```
*/skills/*/scripts/tests/test-*.sh
*-plugin/scripts/tests/test-*.sh
*/hooks/test-*.sh
```

None matches **repo-root** `scripts/tests/test-*.sh`. So a guard living at `scripts/check-*.sh` — where most whole-repo guards live — has a self-test that the runner never picks up, and therefore **no CI signal unless someone hand-wires a step** into `plugin-pr-checks.yml`.

## Why this is worth a structural fix rather than another wiring commit

It is the same class three times now:

- **#2219** — six `check-*.sh` guards had no CI trigger; a `paths:` filter never matched `scripts/**`
- **#2221** — a SKIP reported as PASS, so `test-template-parity.sh` had *never once* executed in CI
- **this session** — I added `scripts/check-dead-api-endpoints.sh` with a 17-assertion self-test, shipped it, and only noticed afterwards that the runner had not picked it up

The recurring shape is a guard that looks protected and is not. `required-to-run-tests.txt`'s own header already records the gap in prose ("repo-root `scripts/tests/` is not covered by `run-skill-script-tests.sh`'s discovery globs"), which means it is known and is being worked around one commit at a time.

The hand-wiring is not free, and it is not automatic: I wired mine only because I went looking. The next author will not necessarily.

## Suggested fix

Add a fourth glob:

```
-o -path './scripts/tests/test-*.sh'
```

Two things to check when doing it, both of which this repo has been bitten by:

1. **Discovery must stay relative.** Per #2290 the walk runs from inside the root against `.`-relative paths, so a scan root that is itself an agent worktree cannot match its own prune. A new glob written absolute would reintroduce that.
2. **Watch the denominator.** Adding a glob raises `TOTAL`, so any guard-integrity assertion pinning an exact count (`scripts/tests/test-run-skill-script-tests.sh`) needs updating in the same change.

Once discovered, the existing `required-to-run-tests.txt` mechanism covers the rest — a listed test that skips is already an ERROR.

## Counterpoint worth weighing

Some repo-root self-tests may be deliberately CI-step-wired because they need a dependency the runner does not install, or need to run at a specific point in the job. Adding the glob would run them twice. Cheap either way, but worth a look before landing.

## What worked well, for balance

[...]

PR #2587, opened 2026-09-04
test(scripts): declare the calendar-estimates suite in the required-to-run manifest

### LINKED ISSUE #2580: test(hooks-plugin): test-no-calendar-estimates.sh is never run by CI or pre-commit
## What

`hooks-plugin/hooks/test-no-calendar-estimates.sh` exists, passes (33 assertions), and **nothing runs it**. It is a regression suite that only fires when a human invokes it by hand.

## Evidence

Verified by grep across the repo. The file appears in neither wiring point:

| Wiring point | `test-bash-antipatterns.sh` | `test-no-calendar-estimates.sh` |
|---|---|---|
| `.pre-commit-config.yaml` | ✅ `:573-577` (`id: test-bash-antipatterns`, with a `files:` pattern gating on the hook and its test) | ❌ absent |
| `scripts/required-to-run-tests.txt` | ✅ `:35` | ❌ absent |

The only reference anywhere is the `# Run: bash …` line inside the file's own header at `:12`.

`test-bash-antipatterns-git-env-isolation.sh` is likewise wired (`.pre-commit-config.yaml:579-583`), so the calendar-estimates suite looks like the odd one out rather than a deliberate exclusion — but that is worth confirming before wiring it, which is why this is an issue and not a PR.

## Why it matters now

Surfaced while landing #2579 (fixes #2574). That PR adds a **control assertion** — the reporter's exact blocked text must still block — whose entire purpose is to make a future silent narrowing of `PATTERN_FUTURE` / `PATTERN_MARKER` go red. As things stand it cannot do that job: nothing executes it.

This is the failure mode `.claude/rules/regression-testing.md` is about, and specifically the lesson recorded in the #1417 follow-up — a regression gate that does not actually execute lets the very bug it names ship again. A suite that is green because nobody ran it is indistinguishable from a suite that is green because the code is correct.

## Suggested fix

One entry in `.pre-commit-config.yaml` mirroring the `test-bash-antipatterns` block:

```yaml
      - id: test-no-calendar-estimates
        name: Test no-calendar-estimates hook
        entry: bash ./hooks-plugin/hooks/test-no-calendar-estimates.sh
        language: system
        pass_filenames: false
        files: '^hooks-plugin/hooks/(no-calendar-estimates\.sh|test-no-calendar-estimates\.sh)$'
```

and/or a line in `scripts/required-to-run-tests.txt`. Copy whichever fields the neighbouring entry actually uses — the block above is shaped from it but should be diffed against the real one before committing.

## Decisions this needs

1. **Both wiring points or one?** They serve different purposes (local pre-commit vs. the CI required-set) and the sibling suite is in both — worth confirming that is the intended pattern rather than history.
2. **Is any hook test deliberately unwired?** If some are excluded on purpose (runtime cost, flakiness), this one may belong in that set, and the right fix is to record the exclusion rather than wire it.

Deliberately not bundled into #2579: it starts running a suite on every contributor's commit, which is a wider change than that PR's scope, and the two questions above are not derivable from the code alone.

Refs #2574, #2579

https://claude.ai/code/session_01M3qqVxAZrwT2faQo6KJTPe

laurigates:
## What

Declares `hooks-plugin/hooks/test-no-calendar-estimates.sh` in
`scripts/required-to-run-tests.txt`, and records the finding as a Known
Regressions row in `.claude/rules/regression-testing.md`.

Two files, 13 added lines, no deletions. No hook, test, workflow or
pre-commit config was touched.

## Why

**#2580's premise was stale in both halves, and this PR now says so.**

The issue reported the suite as absent from both wiring points. Both
claims had expired by the time it was worked:

| Wiring point | State at HEAD |
|---|---|
| `.pre-commit-config.yaml` | Already present at `:678-683`, added by `cecc6e51` (PR #2584) |
| Runner / CI execution | Already executing — the runner's `*/hooks/test-*.sh` glob has discovered it all along |
| `scripts/required-to-run-tests.txt` | **Genuinely missing** — the one real gap, and what this PR closes |

The CI half is easy to get wrong, so it is measured rather than
asserted. Running **origin/main's own runner against origin/main's own
tree**, with no manifest entry present:

```
PASS=./hooks-plugin/hooks/test-no-calendar-estimates.sh
REQUIRED_DECLARED=14
REQUIRED_IN_SCOPE=0
STATUS=OK
```

Discovered and executed, with the manifest silent about it.
`.pre-commit-config.yaml:638-645` states the same thing in prose ("All
13 `hooks-plugin/hooks/test-*.sh` files are ALREADY run in CI").

**So what does the entry buy?** The manifest's other half — the
denominator guard its own header names. A declared path that matches no
discovered test raises `required_test_missing` and exits 1, which turns
a rename, a move, or a discovery-glob collapse (the #2219 / #2333 class)
from silence into a red run.

## Decisions the issue asked for

**Decision 2 — accidental, not deliberate.** Checked before wiring, on
four independent signals: `git log` on the hook and its test (#1379 →
#1570 → #2561 → #2579) records no exclusion; a grep across `scripts/`,
`.github/` and `hooks-plugin/README.md` finds the suite named nowhere
but its own header; no comment anywhere says any hook suite is excluded
on purpose; and the only README rows about this hook document its
`CLAUDE_HOOKS_ENABLE_CALENDAR_ESTIMATES` opt-in, which gates the **hook**
and never its test — the suite sets that variable itself.

**Decision 1 — both wiring points, for different reasons.** The manifest
is not the sibling pattern for pre-commit: of the 13 hook suites only
`test-bash-antipatterns.sh` is listed there, and it is listed because it
SKIPs without `ast-grep`. The manifest has since grown a second class —
`test-check-audit-liveness.sh`, `test-check-golden-set-coverage.sh`,
`test-fetch-arxiv-window.sh`, `test-check-dead-tool-grants.sh` — declared
because no environment makes a skip legitimate. This suite is squarely
that class: it synthesizes its own JSONL transcripts, touches no network
and no `gh`/`task` binary, and carries no skip branch at all.

## Verification

Both directions, on the real corpus rather than a fixture.

| Check | Result |
|---|---|
| `bash hooks-plugin/hooks/test-no-calendar-estimates.sh` | 33 passed, 0 failed, exit 0 |
| Runner scoped to the suite, entry present | `PASS=`, `REQUIRED_DECLARED=15`, `REQUIRED_IN_SCOPE=1`, `REQUIRED_VIOLATIONS=0` |
| **Mutation** — test file moved aside | `SEVERITY=ERROR TYPE=required_test_missing TEST=hooks-plugin/hooks/test-no-calendar-estimates.sh`, `REQUIRED_VIOLATIONS=1`, `STATUS=ERROR`, **exit 1** |
| All 13 hooks-plugin suites | `TOTAL=13 PASSED=13 SKIPPED=0 FAILED=0 REQUIRED_VIOLATIONS=0 STATUS=OK` |
| Baseline on origin/main (no entry) | `PASS=`, `REQUIRED_DECLARED=14`, `REQUIRED_IN_SCOPE=0` — the suite already ran |
| pre-commit at each commit | exit 0 |

No new script check. This is a **wiring** defect, the same class as
#2164 / #2333 / #2574, so the fix is the declaration and the manifest
itself is the guard. `scripts/tests/test-run-skill-script-tests.sh`
already pins `required_test_missing` behaviour, including a
guard-integrity assertion that every manifest entry resolves on disk.
Adding a second guard would double-gate a working mechanism.

## Corrections made after review

An independent verifier put this on hold over three factual defects,
each of which had put a wrong statement into a durable file. All three
are fixed in `951581c2` and `e0e51ab8`.

1. **Framing.** The manifest comment closed on "a control that does not
   provably execute is not a control", which reads as though the suite
   had not been executing. It had been. Replaced with what the entry
   actually buys — the `required_test_missing` denominator guard — and
   proven with the origin/main baseline run above.

2. **Line reference.** "Its **last** assertion block is the #2574
   control" is positionally false. That control sits at `:129-138`
   (mid-file); the file's last two assertion blocks are the reason-token
   loop at `:190-198` and the JSON-validity check at `:202-208`, of 218
   lines. A reader auditing the control by reading the tail lands on a
   different assertion. Now names the range.

3. **Closed list that was not closed.** The "second class" enumeration
   in `.claude/rules/regression-testing.md` omitted
   `scripts/tests/test-check-dead-tool-grants.sh`, which sits at
   `required-to-run-tests.txt:84` under the same no-legitimate-skip
   justification.

## External adversarial review

Re-reviewed after the corrections (`opencode`, `qwen3.8-flash`, plan
agent) against the full `origin/main...HEAD` diff, briefed to hunt
non-obvious defects. It returned one substantive finding, which was
**accepted after verifying it against the shipped patterns**:

> The comment claims the `:129-138` block pins `PATTERN_FUTURE` /
> `PATTERN_MARKER`, but its input matches only `PATTERN_MARKER`.

Correct, and it was an over-claim inherited from the sentence being
repaired. Measured rather than read — the control's input carries no
modal from `PATTERN_FUTURE`'s
`(will|would|should|could|may|might|'ll|going to|gonna)` class and no
take/require/need verb:

```
CONTROL:     PATTERN_FUTURE=no  PATTERN_MARKER=yes
FUTURE_CASE: PATTERN_FUTURE=yes PATTERN_MARKER=no
```

So narrowing `PATTERN_FUTURE` leaves that block green; it is pinned by
the future-tense cases at `:113-118` instead. Fixed in `e0e51ab8` —
the comment now names each block against the pattern it actually pins.

Nothing was rejected.

## Scope note

The PR title moved from `test(hooks-plugin)` to `test(scripts)`. Per
`.claude/rules/conventional-commits.md` the scope is the changelog
label and the **files touched** decide what releases; this diff touches
only `scripts/` and `.claude/rules/`, so `hooks-plugin` mislabelled it.
Nothing releases either way — `test` triggers no bump and no plugin
directory is touched.

Closes #2580

https://claude.ai/code/session_01GwfCsqB6VrNvf8GpC4rxp8


laurigates:

Issue #2506, opened 2026-08-26
feedback(terraform-plugin): validate-terraform-apply.sh matches the phrase anywhere, blocking PR bodies and its own prescribed remedy


laurigates:
## Skill

`terraform-plugin/hooks/validate-terraform-apply.sh` (observed at version `1.7.6`)

## Category

Bug

## Description

The hook matches the phrase `terraform<space>apply` **anywhere in the command string**, with no notion of command position. Two consequences, the second worse than the first:

1. **It fires on documentation prose.** A `gh pr edit --body` whose PR description merely *quotes* the command is denied, even though nothing executes.
2. **The remedy it prescribes is itself blocked.** The error message says to rerun "referencing the saved plan file" — but `terraform apply tfplan` matches the same pattern and is refused identically. There is no input that satisfies the hook. It is not a gate; it is an unconditional deny with instructions attached.

The matcher is the whole story (line 30):

```sh
if ! echo "$COMMAND" | grep -qE '(^|\s)terraform\s+apply(\s|$)'; then
    exit 0
fi
```

`(^|\s)` accepts a match after *any* whitespace in the string — inside a single-quoted argument, inside a heredoc body, inside a `--body` payload. After that check the script falls straight through to an unconditional `block`, with no test for `tfplan`, `-out`, or a prior plan.

## Evidence

Probed by piping JSON into the **shipped** script (not a retyped copy), with a passing control so the harness is not vacuously blocking:

```
gh pr edit --body containing the phrase        exit=2  TERRAFORM SAFETY: Run 'terraform plan' before 'terraform apply'.
the hook's own prescribed remedy               exit=2  TERRAFORM SAFETY: Run 'terraform plan' before 'terraform apply'.
bare apply (true positive)                     exit=2  TERRAFORM SAFETY: Run 'terraform plan' before 'terraform apply'.
control: plan -out (should pass)               exit=0
```

Row 2 is `terraform apply tfplan` — the exact command the hook's own message tells the user to run.

Encountered three times in one session in `ForumViriumHelsinki/infrastructure`, none of them an apply:

| # | Command | What it was |
|---|---|---|
| 1 | `gh pr edit --body "…"` | PR description documenting a targeted apply for a human reader |
| 2 | `python3 - <<'PY'` heredoc rewriting that PR body | same text, one layer further from execution |
| 3 | `gh issue list --search "validate-terraform-apply…"` | a **read-only search** — while researching this very report |

[...]

PR #2590, opened 2026-09-04
fix(terraform-plugin): decide apply command position with ast-grep

laurigates:
## What

`terraform-plugin/hooks/validate-terraform-apply.sh` matched `terraform apply`
anywhere in the command string and then blocked unconditionally, so it refused
text that executes nothing — a PR body quoting the phrase, a heredoc body, a
read-only search query — and refused its own prescribed remedy, since
`terraform apply tfplan` matched the same pattern. No input satisfied the gate.

Command position is now decided by **`ast-grep --lang bash`** (tree-sitter-bash),
the classifier `hooks-plugin/hooks/bash-antipatterns.sh` adopted in #2008. Each
`command` node of the Bash call is examined on its own. An apply naming a saved
plan file is allowed; `-auto-approve` is blocked in every form.

## Why the second implementation

The first draft on this branch hand-rolled the projection in awk: heredoc bodies
dropped, quoted spans collapsed to a `__Q__` placeholder, split into statements
at unquoted separators. It shipped 45 green assertions. An adversarial
re-review reproduced **six bypasses**, each of which the shell genuinely
executes and each of which the pre-fix regex blocked.

Lauri's decision, given in-session: switch the command-position decision to
`ast-grep --lang bash`, following the #2008 precedent, and fail open where
ast-grep is absent. The reasoning: a real parse settles quoting, redirection,
loops and wrappers as a class, whereas one regex patch per bypass leaves the
seventh open.

## The six bypasses, closed

Every input below is the reviewer's verbatim failing case. All exited **0** on
the previous branch head and exit **2** now, each with its own `assert_blocked`
row in the suite.

| # | Failing input | Why it passed |
|---|---|---|
| 1 | `terraform apply "-auto-approve"`, `terraform apply '-auto-approve'` | `__Q__` was counted as a saved plan file, so quoting the flag disabled the `-auto-approve` branch |
| 2 | `terraform apply "-refresh=false"` | same rule made **any** quoted argument look like a reviewed plan |
| 3 | `terraform apply 2>&1 \| tee /tmp/tf.log`, `terraform apply 2> /dev/null` | `<`/`>` were plain separators, so the fd digit was left as a positional operand |
| 4 | `for d in a b; do terraform apply -auto-approve; done`, the `if`/`then` spelling | splitting at `;` left `do`/`then` as the first token and the walker broke on it |
| 5 | `timeout 300` / `nice` / `stdbuf -oL` / `xargs` / `uv run` / `eval` prefixes | the wrapper allowlist was a closed seven-name set |
| 6 | `env -i terraform apply`, `time -p terraform apply` | a bare short flag after a wrapper consumed the next token — the program name |

The parse removes these as a class rather than one at a time: a redirection's fd
digit belongs to a `file_redirect` node, a loop body's command carries no `do`
keyword, and `timeout 300 terraform apply` is one node whose word list contains
`terraform`, so no wrapper allowlist and no flag-arity table exist to be
incomplete. Quotes are now **removed rather than masked**, because the shell
removes them: `terraform apply "-auto-approve"` really delivers `-auto-approve`
to the process.

The one thing tree-sitter cannot see into is a shell invoker's script argument,
so `bash -c "…"`, `sh -c '…'` and `eval "…"` have their quoted arguments
re-parsed as shell. That also closes the review's nit — `sudo -u tf bash -c
"terraform apply"`, a hole on `main` too.

## Corrected claims

Three statements in the previous PR body and README did not reproduce.

| Claim | Correction |
|---|---|
| README: `terraform apply -auto-approve` (any form, plan file or not) — Blocked | Did not hold: the quoted spellings exited 0. It holds now, and the README table lists the quoted spelling, the redirection, the wrapper and the loop explicitly. |
| "An operand the hook cannot resolve is never accepted as a reviewed plan file" | Held only for `$`/backtick operands. A quoted span and a redirection fd digit were both accepted. Now true as written for `$VAR`, `$(…)` and backticks, and the two other cases no longer arise. |
| "Guard integrity is weighted equally with the false-positive half" | Not a property of that suite: 45 assertions contained no case for a quoted flag operand, a redirection, a `do`/`then`-prefixed statement, or an unlisted wrapper. Each of the six now has its own row, with paired allow-controls. |

Two further corrections this change makes to its own documentation:

- **The gate is not unconditional.** Where `ast-grep` is not installed, the hook
  does not fire at all. That is the same fail-open trade `bash-antipatterns.sh`
  makes, and it is now stated in the hook header and the README rather than
  implied away — a PreToolUse safety block that hard-fails without its parser
  breaks every Bash call.
- **A program name assembled from an expansion is not caught.** `A='terraform
  apply -auto-approve'; $A` carries no `terraform` token. Verified: the
  pre-#2506 whole-string regex, the awk draft and this parse all exit 0 on it,
  so it is a standing limitation rather than a regression. Closing it would mean
  re-entering every assignment's value as shell, which blocks a bare `NOTE='we
  ran terraform apply'` — the #2506 false positive again.

## The false-positive half still holds

The three cases from #2506 are allowed, and the block messages prescribe only
commands the hook accepts. `check_prescriptions` does not grep for a literal: it
extracts every `terraform …` line from each message, feeds it back through the
hook, requires exit 0, and asserts at least 3 commands were extracted, so a
message drifting back into prescribing something it refuses fails even if
reworded.

## Suite and mutation

76 assertions, up from 45. Every case pipes JSON into the shipped hook.

| Hook under test | Result |
|---|---|
| `git show origin/main:…` (pre-#2506 regex) | 51 passed / 25 failed |
| `f5c5940d` (the awk draft this replaces) | 57 passed / 19 failed — all six bypass rows red |
| Shipped parse | **76 passed / 0 failed** |

New controls, so the added mechanisms cannot be satisfied by blocking
everything: `timeout 300 terraform apply tfplan`, `for d in a b; do terraform
plan -out=tfplan; done`, a loop quoted inside a `gh pr comment --body`, and a
redirection alongside a real plan file must all stay allowed.

**The fail-open path is pinned in both directions**, which is what keeps the
rest non-vacuous: with `CLAUDE_HOOKS_TERRAFORM_APPLY_NO_ASTGREP=1` an
`-auto-approve` apply must exit 0, and without it the same command must exit 2.
Absent the second assertion, a run on a box lacking ast-grep would report every
row green over a hook that no-ops. The suite therefore moved to the ast-grep
group in `scripts/required-to-run-tests.txt`, beside `test-bash-antipatterns.sh`
(CI installs ast-grep already).

## External re-review

Ran against the corrected diff. Three findings; one accepted, two rejected
after checking them against the code.

**Accepted — MEDIUM, real:** `QUEUE=("${NEXT[@]}")` on an empty array is a fatal
unbound-variable error under `set -u` before bash 4.4, and `#!/usr/bin/env bash`
finds bash **3.2** on a stock macOS. Reproduced: `echo hello` and `terraform
apply tfplan` both exited **1** under `/bin/bash 3.2` — the hook errored on
every command it meant to allow, which no blocked-case assertion can see.
Guarded the copy. `BASH_COMPAT` does not restore the old behaviour, so the pin
runs three allow-cases through `/bin/bash` itself and prints its version;
reintroducing the unguarded copy turns exactly those three rows red and nothing
else.

**Rejected — MEDIUM, premise false:** *"`A='terraform apply -auto-approve'; $A`
is a coverage regression the old whole-string regex did not have."* The old
regex is `(^|\s)terraform\s+apply(\s|$)`; inside the assignment the phrase is
preceded by a quote, not whitespace, so it never matched. Measured all three
versions on that exact input: `main` 0, awk draft 0, shipped 0. It is a standing
limitation, not a regression — now documented in the hook header and the README
rather than left implicit.

**Rejected — LOW, not present:** *"an unescaped `|` inside a GFM table cell
splits the row."* The cell reads `` `terraform apply 2>&1 \| tee log` `` — the
pipe is already escaped.

## Gates

- `bash terraform-plugin/hooks/test-validate-terraform-apply.sh` → `Passed: 76, Failed: 0`
- `bash scripts/run-skill-script-tests.sh --only 'terraform-plugin/hooks/test-*.sh'` → `TOTAL=1 PASSED=1 SKIPPED=0 FAILED=0 REQUIRED_VIOLATIONS=0 STATUS=OK`
- `shellcheck` on both files → clean, no disables added
- `bash scripts/lint-shell-scripts.sh` → 0 errors
- `pre-commit run --files …` over all six touched paths → every applicable hook Passed
- Every row of the README verdict table re-run against the shipped hook
- `bash scripts/tests/test-check-feature-flags-catalog.sh` → 8 passed, 0 failed

CI run 33899901232 went red on `check-feature-flags-catalog.sh --strict`, which
requires every `CLAUDE_HOOKS_*` flag read in a source to appear in
`hooks-plugin/docs/feature-flags.md`; the new fail-open override is catalogued
in a follow-up commit. The terraform suite itself passed on Linux with the
CI-pinned ast-grep 0.43.0 (`PASS=./terraform-plugin/hooks/test-validate-terraform-apply.sh`).

A third live instance of the reported bug landed while writing that commit: the
installed hook (cached v1.7.7) refused `git commit -m "…the terraform apply
gate's fail-open override…"`, because the phrase sat inside the quoted message.
The message had to be moved into a file and passed with `-F` — the same "route
the text around the guard" workaround the issue names as worse than the block.

Refs #2506

https://claude.ai/code/session_01GwfCsqB6VrNvf8GpC4rxp8


laurigates:

laurigates:

PR #2589, opened 2026-09-04
ci(scripts): give check-docs-index a local pre-commit trigger

laurigates:
## What

Adds a `check-docs-index` entry to `.pre-commit-config.yaml` so
`scripts/check-docs-index.sh` runs locally on the paths that commonly move a
stated per-plugin skill/agent count, and corrects the claims that measurement
falsified — in `docs/audits/README.md`, in `.claude/rules/regression-testing.md`,
and in the entry's own comment.

Closes #2522

## Why the issue's proposed wiring would not have worked

The issue's option (a) was a blocking `--strict` entry that "refuses the commit
until the README count is updated". It does not. `doc_count_drift` is **WARN**,
and `check-docs-index.sh` assigns `exit_severity=1` only inside its
`*SEVERITY=ERROR*` branch, so `--strict` exits **0** on a drifted fixture. A
pre-commit hook that exits 0 prints nothing, so a blocking-only entry would have
detected the drift locally and never shown it to the committer.

That exit-0-on-WARN behaviour is **not a new finding** — `.claude/rules/regression-testing.md`
already records it from #2333 ("`--strict` exits 0 on a WARN, so CI was green
while the self-test failed"). What is new here is its consequence for a
pre-commit entry: `verbose: true` is what carries the WARN to the committer, and
`--strict` is retained only so the ERROR tier keeps blocking.

## Blockers closed since the last review

**1. "The ERROR findings rooted in those inputs cannot fire here" was false.**
The entry is `pass_filenames: false`, so once the hook fires the checker walks
the **whole repo**: the `files:` trigger bounds *whether* it runs, never *what*
it can block on. Reproduced before correcting — a dangling row planted in
`feedback-plugin/README.md`, a path the trigger does not select:

```
pre-commit run check-docs-index --files git-plugin/skills/git-commit/SKILL.md
→ Failed, exit 1, duration 4.35s
  SEVERITY=ERROR TYPE=readme_row_dangling
  MSG=feedback-plugin/README.md lists /feedback:ghost but no matching skill directory exists
```

The commit staged neither that README nor that defect. Both copies of the claim
(`.pre-commit-config.yaml` and the always-loaded
`.claude/rules/regression-testing.md` row) now state the real mechanism and name
that consequence, so a committer blocked by an ERROR they did not touch can tell
why. The entry's behaviour is unchanged; widening the trigger would add the
whole-repo walk to more commits without changing any verdict.

**2. The `--strict` finding was presented as new.** Reworded to credit the
[...]

PR #2601, opened 2026-09-05
docs(multi-model-delegation): add the direct OpenCode Go gateway mechanics

laurigates:
## What

Adds a `## Calling the OpenCode Go Gateway Directly` section to
`agent-patterns-plugin/skills/multi-model-delegation`, covering three mechanics that bite when PAL's
MCP server is not connected to the session and you reach the same models at the gateway PAL's own
`opencode_go` provider uses.

## Why

The skill's existing "PAL Mechanics That Bite" table assumes the MCP route. That route is not always
available — a session with no `pal-mcp-server` connected still has `OPENCODE_API_KEY` and the same
catalogue behind it — and the direct path fails in ways the MCP path does not. Each of the three
below cost real time before it was understood, and none is inferable from the endpoint being
OpenAI-compatible.

## The three, and how each was established

**The catalogue you enumerate from is not the catalogue the gateway serves.** A user asked for
`qwen3.8-flash`. It was absent from models.dev's `opencode` provider *and* from PAL's pinned
`conf/opencode_go_models.json`, whose nearest entries are `qwen3.7-max` / `qwen3.6-plus`. The
gateway's own `/v1/models` served it. The reverse also holds: models.dev listed `gemini-3.8-flash`,
which the gateway rejects with `Model gemini-3.8-flash is not supported`. Both snapshots were wrong
in opposite directions, and substituting a plausible near-miss id would have reviewed with a
different model than the one requested, silently.

**A non-streaming request hangs on a long generation.** `http=000` after the full `--max-time`, zero
bytes, no error body. The obvious suspect is payload size, and it is wrong: a 24 KB real payload hung
for the full 900 s while a 33 KB payload of trivial padding returned 200 in 2.4 s. The same request
streamed 3.8 MB with `"stream": true`.

**A reasoning model can spend its entire budget reasoning and emit nothing.** On a 15 KB diff,
`qwen3.8-flash` produced **179,283 characters of `reasoning_content` and zero characters of
`content`**, never reaching `[DONE]`. Raising `max_tokens` does not help, because the budget is going
to reasoning. Cutting the input to ~4.5 KB per call did; the chunks are independent, so they run
concurrently.

## Two consequences recorded for the protocol

Both follow from chunking and neither is obvious from it:

- **A chunked review is partial by construction.** Chunks fail individually (10 of 33 did, on the
  session this came from). The failed-chunk count has to travel with the findings, or a region that
  never returned reads as a region with nothing wrong in it.
- **An empty findings array from a broken reviewer is indistinguishable from a clean review.** The
  harness needs a planted-defect control before any negative is trusted — the existing
  `agent-patterns-plugin:tool-result-traps` law, which this path violates in a new way.

## Verification

- `just lint-compliance` — exit 0. `multi-model-delegation` is not flagged; the warnings in the
  output are pre-existing on `typescript-plugin` and `workflow-orchestration-plugin`.
- `just lint-context-commands` — `All context commands OK`.
- Pre-commit ran the full guard suite on commit: docs-index audit `STATUS=OK` (skill and agent counts
  match disk), skill-description audits, MCP tool-reference denylist, dead-endpoint and
  delegation-reachability guards all passed.
- SKILL.md is 16,703 chars, under the 26,000 ceiling.

The `nudge-plugin-readme-currency` hook fired advisory-only. Not acted on deliberately: this adds a
section to an existing skill without changing its description, its frontmatter tools, or the plugin's
skill count, so `agent-patterns-plugin/README.md` has nothing to restate. The docs-index audit agrees.

## What this does not cover

Everything above was measured against one model (`qwen3.8-flash`) on one workload (GitHub Actions
diffs), in September 2026. The streaming and unbounded-reasoning behaviours are plausibly general to
reasoning models on that gateway but were not re-measured on `glm-*`, `kimi-*`, or `minimax-*`, and
the ~4.5 KB chunk size is what worked rather than a derived threshold. The catalogue-drift finding is
a property of the snapshots as of that date and will not stay true.

### DIFF
--- a/agent-patterns-plugin/skills/multi-model-delegation/REFERENCE.md
+++ b/agent-patterns-plugin/skills/multi-model-delegation/REFERENCE.md
@@ -1,0 +1,48 @@
+# Multi-Model Delegation - Reference
+
+Edge-case mechanics for consulting foreign models: diagnosing an empty
+deferred-tool lookup, and driving the OpenCode Go gateway directly when PAL's
+MCP server is unreachable.
+
+## `No matching deferred tools found` has two causes
+
+Do not read that message as proof the prefix is wrong — under the **correct**
+prefix it means something else entirely, and the two want different responses:
+
+| What you observe | Cause | Do this |
+|---|---|---|
+| The prefix you tried is not the key `claude mcp list` reports | Wrong prefix | Retry with the reported key |
+| The **correct** prefix also finds nothing, PAL's tools never appear in any deferred-tool reminder, yet `claude mcp list` says Connected | The server's tools were never registered *in this session* — likely it connected after session start | Restart the session, or drive the server directly over stdio JSON-RPC (issue #2437) |
+
+One trap in that direct-stdio workaround, worth stating because its symptom
+misleads: **keep stdin open until the response arrives.**
+`subprocess.run(..., input=...)` closes stdin after writing, so the server shuts
+down mid-call and returns an empty result that looks exactly like a hung or
+non-responding model rather than a transport error.
+
+## Calling the OpenCode Go Gateway Directly
+
+PAL is the normal route. When its MCP server is not connected to the session,
+the same models are reachable at `https://opencode.ai/zen/go/v1/chat/completions`
+with `OPENCODE_API_KEY` — the endpoint PAL's own `opencode_go` provider uses.
+Three mechanics bite there that do not bite through PAL, all measured on
+`qwen3.8-flash` reviewing GitHub Actions diffs (2026-09):
+
+| Mechanic | Symptom | Fix |
+|---|---|---|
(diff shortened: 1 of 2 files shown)


laurigates:
Adjacent observation from a concurrent session, plus a correction to my own diagnosis that this PR's measurement settles.

**The observation, on the `opencode run` CLI rather than the raw gateway.** Across three `Workflow` waves I used `opencode run --agent plan -m opencode-go/qwen3.8-flash -f <diff> -- "<review prompt>"` as an external review step, ~15 invocations. Two of them returned nothing: the diffs attached were 1,148 and 458 lines, and the runs stalled mid-read until the `timeout 900` fired. A third attempt on the same content, scoped to a 71-line region, returned findings normally.

The agent that hit it control-tested the tool rather than assuming it was broken — `opencode run --agent plan -m opencode-go/qwen3.8-flash -- "Reply with exactly: CONTROL_OK"` returned `CONTROL_OK` in seconds — and then attributed the stall to **payload size**.

**That attribution does not survive this PR.** The row here measures a 24 KB payload hanging for the full 900 s while a 33 KB payload of trivial content returned 200 in 2.4 s, so size is the wrong discriminator and generation length is the live suspect. My session's data is consistent with that reading too: the 71-line retry differs from the 1,148-line attempt in how much there is to *say* about it, not only in bytes.

So the CLI-side observation is corroborating evidence for this PR's mechanic, not a separate size threshold, and I would not want the "payload size" framing to propagate from my session's notes.

**What is genuinely CLI-specific and may be worth a line here**, if it holds up: `opencode run` presumably streams already, so if the same hang reaches it, `"stream": true` is not the lever a caller can pull. The mitigation available at the CLI is to scope the attachment — review a region rather than a whole diff — and to keep `timeout` on the invocation so a stall is a bounded failure rather than a hung agent. I have not established that the CLI hang and the raw-HTTP hang are the same fault; it is one uncontrolled observation against your controlled one.

Flagging rather than editing, since this is your PR and the measurement is yours.

---
Context: `laurigates/claude-plugins` batch of 2026-09-04 (#2587, #2588, #2589, #2590, #2591, #2593, #2594).

PR #2616, opened 2026-09-06
feat(tools-plugin): document what `just --list` renders, and a tool that reads it

laurigates:
## What

Corrects `justfile-expert` on what `just --list` actually renders, fixes a
factually wrong `set fallback` description, and adds
`tools-plugin/scripts/just-recipe-help.py` — a recipe's comment block plus the
argparse surface of the script it wraps.

## Why

`just --list` shows **one line** per recipe. With no `[doc()]` attribute that
line is the **last line** of the comment block above the recipe — not the
first, and not the block. Neither the skill nor its 1028-line `REFERENCE.md`
mentioned `[doc]` at all, while Best Practices said *"Document always: add
comment before each recipe."* Followed literally, that advice degrades a
listing silently: a block ending in an example lists as the example.

Observed in `laurigates/comfyui-nodes` `lab/justfile` — 47 of 59 recipes
listing a fragment, and one listing **nothing**, because its comment block had
been separated from it by a later insertion and was being attributed to the
recipe above:

```
cadence +ARGS       # one you have to think about every time.
test-sweep PYBIN=   # is belt-and-braces here and load-bearing everywhere else.
orphans *FLAGS      #                                   <- nothing at all
```

## Corrections

| | |
|---|---|
| `set fallback` | documented as *"Fail on first error in recipe"*. It is not error handling — it makes `just` search **parent directories** for a recipe this justfile does not define. Verified: a child justfile with it set resolves `just parent-recipe` from the parent |
| `[doc]` | absent from the attribute list. Added, with bare `[doc]` (suppresses) and the comment fallback |
| `[private]` | also hides from `--summary`, not just `--list` |
| `[group]` | both spellings work (`[group: "x"]` and `[group("x")]`), and `--groups` lists them |
| `mod` | does **not** inherit `set fallback` — `just sub::parent-recipe` fails |

Plus a new section on the `--list` contract: the fallback table, the
**adjacency** binding (a blank line ends a block; inserting a recipe between a
block and its recipe silently reassigns it), and the fact that a recipe with a
required positional has **no `--help` form** — `just` refuses the call before
the tool runs, so `just recipe --help` is not an escape hatch.

Every claim verified against `just` 1.58.0 rather than recalled.

## The tool

The flags of a wrapped script were reachable only by opening it, and both
escapes are closed: `--help` is refused for a recipe with a positional, and
supplying dummies hits an argparse that sits behind a module-scope
`import numpy` on a checkout without it.

So `just-recipe-help.py` reads the **shipped source** — the comment block from
the justfile, the flags from the script's AST. Stdlib only, so it answers the
same where the dependencies are absent, which is exactly where `--help` cannot.
It resolves `{{VAR}}` paths through `just --evaluate` when `just` is present and
falls back to a basename search when it is not.

## The audit criterion took three attempts

This is the part worth reviewing. `--audit` reports recipes whose rendered line
is unusable — and two more obvious rules were tried and rejected:

| Criterion | Result |
|---|---|
| "has no `[doc()]`" | flags **33 of 33** recipes in this repo's own justfile. A gate that fires on everything is one nobody reads |
| "has a multi-line comment block" | punishes the careful habit of ending a long block with a deliberate one-line summary — which `config-drift-calibrate` in this repo already does |
| **shape of the rendered line** | indented sub-item / command example / lowercase continuation / no block at all |

Measured with the shipped rule: **0 findings** on this repo's justfile, **47 of
59** on the known-bad one, each with a named reason.

## Verification

- `tools-plugin/scripts/tests/test-just-recipe-help.sh` — 22 assertions, wired
  into `.pre-commit-config.yaml`.
- **Teeth-checked**: four mutations — the params-with-`=` regex, the `_`-prefix
  privacy rule, the fragment criterion, the unreadable-flag refusal — each
  CAUGHT by the assertion it names, with a control correctly MISSED.
- `scripts/plugin-compliance-check.sh` exits 0.
- Run against three unrelated justfiles (this repo's, `comfyui-nodes` root, and
  `lab/`) rather than only the one it was built from.

Two notes for the reviewer:

- `ruff format` reformatted the script on first commit and **un-anchored one
  mutation** — it reported `BAD ANCHOR`, not a pass. Re-anchored and re-verified.
- `justfile-expert/SKILL.md` is now 17,865 chars against the 26,000 ceiling. The
  "consider extracting" advisory at 10,000 was already firing before this change
  (14,994); the compliance check exits 0. The `--list` contract is a correctness
  rule, so it belongs in the body rather than `REFERENCE.md`, but the MCP
  integration section below it is a fair candidate for extraction later.

A companion PR updates `configure-plugin:configure-justfile` to audit what
`--list` renders instead of whether a comment exists. Split because a
squash-merge lands the **PR title**, so one PR spanning both plugins would
apply a single release type to both packages.

https://claude.ai/code/session_0147YE9fuCKttGe9MxASqy4x

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -209,6 +209,12 @@ repos:
         language: system
         files: '^scripts/(check-subagent-types\.sh|tests/test-check-subagent-types\.sh)$'
         pass_filenames: false
+      - id: test-just-recipe-help
+        name: just recipe-help / --list description regression
+        entry: bash ./tools-plugin/scripts/tests/test-just-recipe-help.sh
+        language: system
+        files: '^tools-plugin/scripts/(just-recipe-help\.py|tests/test-just-recipe-help\.sh)$'
+        pass_filenames: false
       - id: check-unrendered-templates
         name: no unrendered {{ if }} template conditionals in skill bodies (#2265)
         entry: ./scripts/check-unrendered-templates.sh
--- a/tools-plugin/scripts/just-recipe-help.py
+++ b/tools-plugin/scripts/just-recipe-help.py
@@ -1,0 +1,647 @@
+#!/usr/bin/env python3
+"""What does this just recipe take? -- the flag list, without opening the script.
+
+`just --list` renders ONE line per recipe, and a justfile that keeps its
+reasoning in the comment block above each recipe cannot show it there. Worse,
+with no `[doc()]` attribute just falls back to that block's LAST line -- almost
+always the tail of an example -- so a well-commented justfile lists as a wall
+of fragments and the flags of the tool a recipe wraps are reachable only by
+opening it.
+
+Two mechanical reasons the obvious escapes do not work:
+
+  1. `just <recipe> --help` is REFUSED BY JUST for any recipe with a required
+     positional: "got 1 positional argument but takes at least 2". There is no
+     bare-help spelling for such a recipe.
+  2. Supplying dummy positionals does reach the tool -- and a script whose
+     argparse sits behind a module-scope import of a heavy dependency raises
+     ImportError before argparse ever runs, so --help is unreachable on any
+     checkout without that dependency installed.
+
+So this reads the SHIPPED SOURCE rather than running it: the recipe's own
+comment block out of the justfile, and the flags out of the script's AST. It
+imports nothing but the stdlib and needs no virtualenv, so it answers the same
+everywhere.
+
+It is NOT a substitute for argparse's own output and must not drift into one.
+Where the tool's dependencies are installed its own --help is authoritative,
+and this prints the exact command that reaches it.
+
+USAGE
+    just-recipe-help.py <recipe>        # comment block + flags for one recipe
+    just-recipe-help.py <script.py>     # one script by name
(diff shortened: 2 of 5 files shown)


laurigates:
The red `test` check is a pre-existing `main` failure, not this branch.

The run's only failure is `scripts/tests/test-lint-mcp-tool-references.sh`, on
two assertions about `agent-patterns-plugin:multi-model-delegation`'s
documentation content:

```
FAIL: multi-model-delegation splits the two causes of a ToolSearch miss
FAIL: multi-model-delegation records the reporter's stdio workaround caveat
PASSED=18 FAILED=2
```

I ran it on a tree that is `origin/main` plus a single unrelated one-file docs
commit and it fails identically, so nothing here reaches it — this branch
touches `tools-plugin/{skills,scripts}` and `.pre-commit-config.yaml` only.

This branch's own test passed in the same run:
`PASS=./tools-plugin/scripts/tests/test-just-recipe-help.sh`, within
`DISCOVERED=159 / PASSED=158 / FAILED=1`.

Worth a separate issue — that suite has been red on `main` independently of
this work.

PR #2622, opened 2026-09-07
fix(feedback-plugin): deny bare force pushes instead of prompting on every push

laurigates:
## Problem

`feedback-plugin/hooks/check-open-pr.sh` emitted `permissionDecision: "ask"` for **every** `git push` to a branch with an open PR, and never inspected `--force` at all — the token loop `continue`d past `--force` / `--force-with-lease` and nothing downstream read them.

Two consequences:

1. **The prompt fired on the normal workflow.** Adding commits to an open PR is what a PR is for, so the false-positive rate was near 100%. It never distinguished a history rewrite from a fast-forward, which is the litigation test in `.claude/rules/hook-block-vs-nudge.md` ("does the gate exempt the dangerous variant while blocking benign ones?").
2. **`ask` is the only PreToolUse decision the model cannot read.** The reason string was written as instructions to an agent (*"add `[force-push-ok]` to the commit message"*), but only the human ever sees it. A "No" reaches the agent as a bare permission denial that ends its turn — in a workflow subagent, that means a stalled run needing a manual re-prompt.

## Change

Narrow the trigger to an **unconditional** force push and switch the decision to `deny`, whose reason is returned to the model as tool feedback.

| Command shape | Before | After |
|---|---|---|
| `git push origin feature/x` (fast-forward) | `ask` | silent |
| `git push --force-with-lease …` / `--force-if-includes …` | `ask` | silent |
| `git push --force …`, no open PR | silent | silent |
| `git push --force …`, open PR, `[force-push-ok]` in last commit | silent | silent |
| `git push --force …` / `-f` / `+refspec`, open PR | `ask` | **`deny`** → model re-runs with `--force-with-lease` |

`--force-with-lease` is exempt because the lease *is* the check the prompt asked a human to perform: git refuses when the remote ref moved since the last fetch, so it cannot silently discard a reviewer suggestion, a CI auto-fix, or a coworker agent's commit. Gating it added an interruption and no safety.

The `[force-push-ok]` escape hatch stays keyed on the commit message rather than the command, so the model cannot self-serve it (`.claude/rules/handling-blocked-hooks.md`).

## Parser bugs fixed along the way

- **`-f` was consumed as the branch name.** The skipped-flag enumeration listed `--force` but not `-f`, so `git push -f origin feature/foo` resolved `target_branch="-f"`. Now any leading-dash token is skipped, and positional order (`[remote] [refspec]`) determines the branch — which also fixes a custom remote name being read as a branch.
- **A push inside a compound command was parsed with its neighbours' arguments.** The command is now split on statement separators and only the `git push` segment is parsed. Splitting uses bash parameter expansion rather than `sed`, since BSD `sed` emits a literal `n` for `\n` in an `s///` replacement (`.claude/rules/shell-scripting.md`).
- `refs/heads/` is stripped from the refspec destination before the `gh pr list --head` lookup.

## Second commit: a stale guard blocking `main` (`369617d`)

The `test` job was red on this PR when it opened, on `scripts/tests/test-lint-mcp-tool-references.sh` — and identically red on a clean `origin/main` worktree, so it was never this PR's failure. It blocked every PR in the repo from 2026-09-05.

The docs were not the defect. #2601 (`e6ec35f`) **relocated** two sections out of `agent-patterns-plugin/skills/multi-model-delegation/SKILL.md` into its `REFERENCE.md` sidecar and left a pointer behind; both asserted phrases are alive at `REFERENCE.md:7` and `:18`. Only the guard's search path was stale — the same file's CASE 8 ("REFERENCE.md sidecars are scanned") passed throughout, which is what narrows it to the search path.

The two assertions that pin relocated detail now search both files, resolved from the skill's own directory rather than a second hardcoded path, so a future relocation between `SKILL.md` and its sidecar cannot break them again. The three that pin the skill's own framing stay on `SKILL.md`. `grep -qH` per `.claude/rules/shell-scripting.md`: GNU `grep` drops the filename prefix on a single-file argument, which is exactly the shape here when a skill has no sidecar.

Applied on this branch rather than split out because the repo's branch policy keeps this session on one branch, and leaving it meant every PR stayed blocked.

## Testing

The force-push suite lived at `hooks/tests/test_check_open_pr.sh`, matching **neither** the directory nor the hyphen in `scripts/run-skill-script-tests.sh`'s `*/hooks/test-*.sh` discovery predicate — it was the only hook suite in the repo shaped that way, and had never run in CI. Moved to `hooks/test-check-open-pr.sh` and extended from 6 to 19 assertions:

```
19/19 test(s) passed
```

Covering each force spelling (`--force`, `-f`, `-fu`, `+refspec`), both lease exemptions, the plain and compound pass-through cases, the `refs/heads/` and current-branch-fallback paths, and both escape hatches. The `gh` shim answers only for `feature/foo`, so a mis-parsed branch surfaces as a silent pass rather than a false deny.

The guard fix was verified in **both** directions rather than only on the fixed tree: 20/20 against the current corpus, and re-arming the defect by renaming both phrases in `REFERENCE.md` reproduces `PASSED=18 FAILED=2` — so the widened search cannot degrade into matching nothing.

`scripts/lint-shell-scripts.sh` (0 errors), `check-docs-index.sh --strict`, `check-plugin-readme-currency.sh`, `check-prose-house-style.sh`, and `check-context-engineering.py --strict` are clean.

## Docs

- `feedback-plugin/README.md` — new "Open-PR force-push guard" section with the outcome table and the two rationales; both hook rows updated.
- `.claude/rules/hooks-reference.md` — new "Who reads `permissionDecisionReason`" subsection. The audience distinction (`deny` → the model, `ask` → the human, with no agent-facing variant of `ask`) was undocumented anywhere in the repo, which is what made the original miscategorisation easy to reach for.
- `.claude/rules/regression-testing.md` — a row for the stale-guard class: a guard pinning a literal to one file goes stale the moment that content moves under the `references/` convention, the same shape as the `parallel-agent-dispatch` split in #2143.

https://claude.ai/code/session_015xWTGVZyVPrGa1J7F47kDC

---
_Generated by [Claude Code](https://claude.ai/code/session_015xWTGVZyVPrGa1J7F47kDC)_


laurigates:
## `test` is red on a pre-existing `main` failure, not this PR

`test` fails on `scripts/tests/test-lint-mcp-tool-references.sh` (2 of 20 assertions). It is not this PR's:

- This diff touches `feedback-plugin/hooks/`, `feedback-plugin/README.md`, and `.claude/rules/hooks-reference.md` only — nothing under `agent-patterns-plugin/`.
- The same two assertions fail on a clean `origin/main` worktree (`PASSED=18 FAILED=2`, identical).
- Every other suite passed: `DISCOVERED=160 PASSED=159 FAILED=1`, the one failure being this.

No fix PR for it exists (searched open PRs for both `multi-model-delegation` and `lint-mcp-tool-references`), so there is nothing to port. Not re-running: the failure is deterministic and reproduces locally on `main`'s tree, which rules out a flake more firmly than a re-run would.

### Root cause

The guard is stale, not the docs. #2601 (`e6ec35f`, merged 2026-09-05) **relocated** two sections out of `multi-model-delegation/SKILL.md` into its `REFERENCE.md` sidecar, leaving a pointer behind:

```
-### `No matching deferred tools found` has two causes
-...
-misleads: **keep stdin open until the response arrives.**
+If a lookup under the correct prefix still finds nothing, see
+[REFERENCE.md](REFERENCE.md) — that message has a second cause.
```

Both phrases are present today at `REFERENCE.md:7` and `REFERENCE.md:18`. The knowledge the guard protects is intact and reachable; only its search path is wrong — it greps `SKILL.md` alone (`test-lint-mcp-tool-references.sh:313`), while the file's own CASE 8 already establishes that sidecars are in scope ("REFERENCE.md sidecars are scanned", which passes).

### Proposed patch

Two assertions at `scripts/tests/test-lint-mcp-tool-references.sh:335` and `:341` should search the sidecar as well:

```diff
+  skill_docs=("$skill")
+  [ -f "${skill%/SKILL.md}/REFERENCE.md" ] && skill_docs+=("${skill%/SKILL.md}/REFERENCE.md")
+
   # Problem 4: the issue's own evidence is that the CORRECT prefix also missed.
-  if grep -q 'has two causes' "$skill"; then
+  if grep -qH 'has two causes' "${skill_docs[@]}"; then
     ok "multi-model-delegation splits the two causes of a ToolSearch miss"
   else
     bad "multi-model-delegation splits the two causes of a ToolSearch miss" \
       "expected the correct-prefix-also-misses case to be documented"
   fi
-  if grep -q 'stdin open until the response arrives' "$skill"; then
+  if grep -qH 'stdin open until the response arrives' "${skill_docs[@]}"; then
     ok "multi-model-delegation records the reporter's stdio workaround caveat"
```

(`-H` per `.claude/rules/shell-scripting.md` — GNU `grep` drops the filename prefix on a single-file argument, which is exactly the shape here when the sidecar is absent.)

Not applying it in this PR: the failure is in unrelated code and fixing it here would widen a `feedback-plugin` hook change into `agent-patterns-plugin`'s test guard. Happy to apply it on this branch if you would rather unblock in one go — it blocks every PR in the repo until it lands somewhere.

---

laurigates:
Applied the guard patch in `369617d` at the author's request, so this PR unblocks `main` as well as itself.

The fix is slightly different from the patch I proposed above: rather than building the path array inline at the two assertions, the sidecar is resolved once from the skill's own directory, so a future relocation between `SKILL.md` and its sidecar cannot break the assertions again. Assertions about the skill's own framing (`registered under`, `claude mcp list`, `claude mcp add -s user`) stay on `SKILL.md` — only the two that pin relocated detail search both files.

Verified in both directions rather than only on the fixed tree:

- `20/20` against the current corpus (was `PASSED=18 FAILED=2`).
- Re-arming the defect — renaming both phrases in `REFERENCE.md` so they exist in neither file — reproduces `PASSED=18 FAILED=2`. Without that half, a patch that widened the search into matching nothing would have looked identical to a working one.

Also clean: `scripts/lint-shell-scripts.sh` (0 errors), `check-docs-index.sh --strict`, `check-context-engineering.py --strict` (exit 0), and `test-check-docs-index.sh`.

---

laurigates:

PR #2668, opened 2026-09-16
feat(hooks-plugin): ask before a workflow spawns more agents than the limit

laurigates:
## What

A PreToolUse hook on the **Workflow** tool that emits `ask` when a workflow script's estimated agent count exceeds a configurable limit (default 10, matching the `medium` `workflowSizeGuideline`).

- `hooks-plugin/hooks/workflow-scale-guard.sh` — the hook
- `hooks-plugin/hooks/workflow-scale-estimate.py` — static agent-count estimator, emits a `KEY=VALUE` rollup
- `hooks-plugin/hooks/test-workflow-scale-guard.sh` — 15 cases
- Registered under `PreToolUse` / matcher `Workflow`; flags added to `docs/feature-flags.md`; pre-commit entry for parity with sibling hook tests

## Why

A workflow's cost is set almost entirely by its agent count, and that number is a property of the script's shape, not of the prompt that asked for it. Two shipped mechanisms look like they cover this and do not:

- **`workflowSizeGuideline`** is advisory system-prompt text that ends *"This is a guideline, not a hard limit — follow it unless the user's prompt calls for a different scale."* A sweep-shaped prompt reads exactly like a call for a different scale, so it can be reasoned past without being wrong on its face.
- **`skipWorkflowUsageWarning`** is, per the CLI's own schema, *"whether the user has accepted the multi-agent workflow usage warning. Until set, auto permission mode prompts before running a workflow."* It is a **one-time** acceptance. Once set, auto mode stops prompting before every workflow, at every scale, permanently.

Observed 2026-09-15: five workflow runs in one session spawned **496 subagents** against a `medium` guideline of **10**, and cost **$528** in an afternoon — 67% of that day's total spend and roughly 8x the entire GitHub Actions bill across 37 repos. Nothing prompted, because the acceptance flag had been set long before.

The spend was dominated by prompt-cache **creation**: 55.2M cache-write tokens at 1.25x input price, against 252.5M cache reads at 0.1x. Every fresh agent builds its own cache and short-lived ones never read it back, so cost scales with agent *count*, not with how much context each one reuses.

| Token class | Volume | Share of that session's $528 |
|---|---|---|
| cache write | 55.2M | $345 (64%) |
| cache read | 252.5M | $126 (24%) |
| output | 2.45M | $61 (11%) |

## How it decides

| Script shape | Behavior |
|---|---|
| Bounded fan-out within the limit (literal array, `.slice(0, N)`, `Array.from({length: N})`) | Silent |
| One `agent()` per runtime-length list | Silent at defaults |
| Two or more `agent()` sites inside one runtime-length fan-out | `ask` |
| A runtime-length fan-out nested inside another | `ask` |
| Bounded fan-out whose product exceeds the limit | `ask` |
| Resume, saved workflow by `name`, unparsable script, no `agent()` | Silent (fails open) |

A fan-out whose length is only knowable at runtime is neither waved through nor hard-blocked — it is costed at an assumed width (default 8) so the limit governs every shape. That single mechanism replaced an earlier tier ladder and puts the split where the cost actually is: **agents per item and nesting depth**, not fan-out presence. One reviewer per changed file stays silent; four agents per item does not.

**`ask`, not a block, on purpose.** The failure was not that the workflow was forbidden, it was that nobody was asked. A block would make the agent judge whether the scale is justified — the judgment that already went wrong. `ask` puts the number in front of the person paying and lets them approve in one keystroke.

## Verification

Control-tested against the four workflow scripts the CLI persisted from that real session, via both the inline `script` and the `scriptPath` input:

| Script | Estimate | Verdict |
|---|---|---|
| `onboarding-decision-facts` | ~16 | ask |
| `onboarding-docs-patch-batch1` | ~32 | ask |
| `onboarding-docs-patch-batch2` | ~32 | ask |
| `onboarding-gap-audit` | ~520 | ask |

All four would have prompted. The negative controls carry the real contract — a guard that asks on ordinary workflows gets disabled within a day — so the suite pins that small bounded workflows, an explicit `.slice(0, 5)` cap, one agent per runtime item, and `agent(` appearing only inside strings or comments all stay silent.

```
bash hooks-plugin/hooks/test-workflow-scale-guard.sh   # PASS=15 FAIL=0
bash scripts/lint-shell-scripts.sh                     # 0 errors
bash scripts/plugin-compliance-check.sh hooks-plugin   # all green
bash scripts/check-feature-flags-catalog.sh            # STATUS=OK MISSING_COUNT=0
shellcheck hooks-plugin/hooks/workflow-scale-*.sh      # clean
```

## Known gaps, all fail-open

`agent()` inside a `${...}` template interpolation is invisible (prompts live in templates, calls do not); fan-out through a helper function is not followed; arrays built by `push` in a loop are not modeled. Each of these under-counts rather than over-counts, which keeps the false-positive rate at zero at the cost of some misses.

https://claude.ai/code/session_013BrE9johJKvwk9NqKGZQon


laurigates:
Context and follow-ups for this gate:

- #2670 — the cost work behind it: agent count, not context size, is what sets a Workflow run's bill (55.2M cache writes = 64% of the $528 session).
- #2669 — the CI half of the same spend question.

This PR closes the "nobody was asked" half only. It does not make a run cheaper; it puts the number in front of someone before the run starts.

PR #2701, opened 2026-09-20
fix(blueprint-plugin): narrow ADR required sections to Context/Decision/Consequences

### LINKED ISSUE #2446: docs(blueprint): 20 of 22 ADRs lack the template sections the schema requires
## What

20 of 22 ADRs are missing the `## Options Considered` and `## Related ADRs`
sections that `blueprint-plugin/schemas/adr.schema.json` declares required.
Four also lack `## Decision`, one lacks `## Consequences`.

Only `0014-reusable-workflows-in-repository.md` and
`0020-blueprint-autonomy-levels.md` carry all five.

| Section | Missing in |
|---|---|
| `Context` | 0 of 22 |
| `Decision` | 4 of 22 |
| `Consequences` | 1 of 22 |
| `Options Considered` | 20 of 22 |
| `Related ADRs` | 20 of 22 |

## Why this is filed now

The gap is pre-existing, but until #2440 nothing could see it: the ADR schema
was referenced by no code at all, and the hook that did the validating carried
its own separate field list. The reconciled validator reports all 53 instances
as WARN.

WARN is deliberate. Blocking would refuse an edit to a 2026-01 ADR until
someone back-wrote an "Options Considered" section it never had, which is the
style-block `.claude/rules/hook-block-vs-nudge.md` prohibits. But a warning
nobody is chasing is also how this stayed invisible for months, hence this
issue.

## Reproduce

```
uv run --quiet --script blueprint-plugin/scripts/check-schema.py --kind adr --file docs/adrs/0001-plugin-based-architecture.md
```

Or across the corpus, counting by section:

```
cd docs/adrs && for s in "Options Considered" "Related ADRs"; do n=0; for f in [0-9]*.md; do grep -q "^## ${s}$" "$f" || n=$((n+1)); done; echo "$s: $n"; done
```

## What this needs

A judgement call, not a mechanical sweep. Two honest options:

1. **Write the missing sections.** These are real decisions with real
   alternatives that were weighed; the rationale exists in the PRs and issues
   around each ADR. This is research, not formatting — inventing an "Options
   Considered" from the ADR's own text would fabricate deliberation that may
   not have happened.
2. **Narrow the schema's required set** to what the corpus actually sustains
   (`Context`, `Decision`, `Consequences`) and drop the other two to optional.
   Defensible if the template asks for more structure than these ADRs warrant.

Option 2 is a smaller diff and would take the warning count from 53 to 5. It
should be an explicit decision recorded somewhere, not a quiet edit to the
schema — the required set is currently the one the pre-reconciliation hook
enforced on `Write`.

Deliberately not attempted in #2440: writing 40 sections of decision rationale
for decisions made months ago is fabrication, and silently relaxing the schema
in the same PR that made it authoritative would have buried the choice.

laurigates:
Implements the #2446 decision taken during open-issues triage: narrow the schema rather than back-write 40 sections of decision rationale.

## What changed

`blueprint-plugin/schemas/adr.schema.json` required five sections. 20 of this repo's 22 ADRs carry neither `Options Considered` nor `Related ADRs`, so the validator reported 45 section warnings nobody could honestly clear — inventing an "Options Considered" from an ADR's own text would fabricate deliberation that may never have happened.

`Options Considered` and `Related ADRs` are now **optional**. The ADR template still asks for all five, and `/blueprint:adr-validate` remains the gate that can insist on them for new ADRs. Missing sections still WARN rather than block, per `.claude/rules/hook-block-vs-nudge.md`.

## Measurement

| | Before | After |
|---|---|---|
| Section WARNs across `docs/adrs/[0-9]*.md` | 45 | **5** |
| Total WARNs | 46 | 6 |
| ERRORs | 0 | 0 |

The extra non-section WARN on both sides is ADR-0014's `status: Superseded` with no `superseded-by` — unrelated to this change.

**The issue's stated baseline of 53 does not reproduce.** Measured today: 45 section WARNs. The issue's own per-section table (20 `Options Considered` + 20 `Related ADRs` + 4 `Decision` + 1 `Consequences`) sums to exactly 45, so the discrepancy is in its total, not its counts. The "land at 5" target held.

## Files

- `blueprint-plugin/schemas/adr.schema.json` — required set becomes `Context`/`Decision`/`Consequences`; the `sections` description rewritten so the file no longer states the old set.
- `docs/adrs/0023-adr-required-sections-narrowed.md` — **new.** The issue insisted this be an explicit recorded decision, not a quiet schema edit. It carries all five sections itself, practising what the narrowed schema still permits.
- `blueprint-plugin/scripts/tests/test-check-schema.sh` — its "a missing required section WARNs" assertion built its fixture by deleting `## Related ADRs`, which is now legitimately clean, so the suite went red on the schema edit. Fixture switched to `## Consequences` (still required), preserving the assertion's intent, plus a companion assertion that a three-section ADR validates `STATUS=OK` — both polarities now pinned.
- `blueprint-plugin/README.md`, `docs/adrs/README.md` — the schema table row, the WARN rationale, and the ADR index row. These land in the same commit per `.claude/rules/docs-currency.md`.

Neither `validate-adr-frontmatter.sh` nor `check-schema.py` carries its own copy of the section list (the hook is a two-line exec into `validate-frontmatter.sh`; the script reads `sections.required` from the schema), so neither needed editing. Confirmed by the passing `no field lists:` assertions.

## Verification

- `bash blueprint-plugin/scripts/tests/test-check-schema.sh` → `PASSED=24 FAILED=0` (was `PASSED=22 FAILED=1` after the bare schema edit, before the fixture fix)
- `bash scripts/check-docs-index.sh` → `STATUS=OK ISSUE_COUNT=0`
- `pre-commit run --files <the five paths>` → exit 0
- Single-file reproduce from the issue on `docs/adrs/0001-plugin-based-architecture.md`: `STATUS=WARN ISSUE_COUNT=2` → `STATUS=OK ISSUE_COUNT=0`
- New ADR-0023 validates `STATUS=OK ISSUE_COUNT=0`

**Not run:** `blueprint-plugin/hooks/spec/validate_adr_frontmatter_spec.sh` — shellspec is not installed in this environment.

## Incidental finding, not fixed here

`blueprint-plugin/hooks/validate-frontmatter.sh:17` cites `scripts/tests/test-schema-field-parity.sh` as the guard against a field list reappearing in a hook. **That file does not exist anywhere in the repo.** The protection itself is real and passing — it is the four `no field lists:` assertions in `test-check-schema.sh` — so only the comment's path is wrong. Left alone as out of scope for this diff.

Closes #2446

https://claude.ai/code/session_015DMUd2TJ3deUE6DsHsaTHA

---
_Generated by [Claude Code](https://claude.ai/code/session_015DMUd2TJ3deUE6DsHsaTHA)_


laurigates (commented):
Reviewed with five `opencode-go/deepseek-4.1-flash` subagents.

## Overall verdict

**Risk: MEDIUM** — the schema narrowing itself is correct, minimal, and SSOT-compliant, but the PR ships inaccurate claims about where the “five-section ceiling” still lives, and leaves two doc surfaces stale.

---

## Blocking issues (should be fixed before merge)

1. **The stated enforcement ceiling does not exist.**
   - `adr.schema.json`’s new `sections.description`, `blueprint-plugin/README.md`, and ADR-0023 all say the ADR template and `/blueprint:adr-validate` “still ask for all five” / “can insist on them for new ADRs.”
   - In reality, `/blueprint:adr-validate` validates relationships, reference integrity, numbers, and index drift — it never checks sections. There is also no in-repo ADR authoring template that lists five sections; `docs/adrs/README.md` §ADR Format documents the Nygard three-section shape, and `/blueprint:derive-plans` points at a MADR template whose headings don’t match the schema.
   - **Fix:** either add a real section check to `blueprint-adr-validate` (or create a five-section template), or rewrite the schema description and ADR-0023 to say `Options Considered`/`Related ADRs` are recommended but unenforced.

2. **`blueprint-plugin/docs/hook-design-decisions.md:117-124` still lists all five sections as “P0 - Blocking.”**
   - This is now wrong on both counts (two sections are optional, and sections WARN rather than block). Same docs-currency class the PR applies elsewhere.

3. **ADR-0023 describes `blueprint-plugin/README.md` as still stale.**
   - Its Consequences/Mitigations prose says the README “still enumerates five … needs the same narrowing — tracked on #2446,” but this PR already narrows those rows and closes #2446. Recast to past tense.

4. **`docs/adrs/README.md` is missing ADR-0023 under `## Categories`.**
   - Every prior ADR is categorized; `domain: architecture` fits “Architecture Evolution.”

5. **The user-facing warning message still cites non-existent enforcers.**
   - `adr.schema.json` `sections.title` reads: *“required `## ` section missing — the ADR/PRP template declares it; add it or run the matching validate skill.”* This is the string users actually see, and it points at a template/skill that don’t check sections.

6. **Dangling filename in SSOT docs.**
   - `adr.schema.json` and `validate-frontmatter.sh` cite `scripts/tests/test-schema-field-parity.sh`, which does not exist. The PR body flags it as out of scope, but it’s a one-line comment fix in a file whose sole-enforcement-point role this change depends on.

---

## Suggestions (non-blocking)

- **Test fixture fidelity:** the new “optional pair must NOT warn” assertion uses `sed` ranges that exit 0 even if the headings are absent/renamed. Add a guard that the fixture really has exactly three sections (e.g. `grep -c '^## ' == 3`) so a no-op sed can’t pass silently.
- **`+1d` sed assumption:** the fixture deletes “heading + exactly one following line.” Works today, but a blank line or multi-line section body would make the test mean something else. Use a range to the next `## ` or add a comment.
- **Assertion count drift:** `.claude/rules/regression-testing.md` still says `test-check-schema.sh` has 23 assertions; it now has 24.
- **`has_issue` uses regex-by-default** (`grep -q` without `-F`), though all callers currently pass literal strings. Use `grep -qF` to match its “substring” contract.
- **Dormant shellspec specs** (`hooks/spec/validate_adr_frontmatter_spec.sh` etc.) assert blocking `ERROR: Missing required section` — pre-existing, but they now contradict the WARN-only design. Fix or delete in a follow-up.

---

## Positive notes

- **SSOT genuinely holds:** no code path duplicates the section list; `validate-adr-frontmatter.sh` is a thin `exec` and `check-schema.py` reads `sections.required` from the schema.
- **Severity stays WARN-only:** `x-blueprint-severity: "warning"` is untouched, and a missing required section still emits `SEVERITY=WARN` with exit 0.
- **Tests pin both polarities:** deleting `## Consequences` still warns, and a three-section ADR now validates `STATUS=OK`, preventing a schema that requires nothing from passing.
- **Measurement is honest and reproducible:** 45 → 5 section warnings across the 22 existing ADRs, and the PR correctly notes that issue #2446’s “53” baseline does not reproduce.
- **Docs land in the same commit** (schema, README, ADR index, ADR-0023) per `docs-currency.md`.
- **ADR-0023 validates clean** against the narrowed schema and carries all five sections itself, modeling the optional-but-encouraged shape.

laurigates:
Verified all six blocking findings independently before acting. **All six were real.** Fixed in `e1d4c98`.

## Blocking

**1. The enforcement ceiling doesn't exist.** Confirmed — `grep -niE "section|Options Considered|Related ADRs" blueprint-adr-validate/SKILL.md` returns zero matches, and there is no five-section authoring template in the repo.

You offered two fixes. I took the second: **withdraw the claim**, don't build the gate. Adding a section check to `blueprint-adr-validate` is a different decision from narrowing the schema, and #2446's triage decision was the narrowing — building a new gate inside this PR would widen it past what was decided. `sections.description` now says which sections it polices and which it deliberately does not; ADR-0023 states the no-ceiling consequence plainly rather than mitigating it with a claim that isn't true, and names building the ceiling as separate work. Happy to file that follow-up if you want it.

**2. `hook-design-decisions.md` "P0 - Blocking".** Confirmed at lines 116–124, and wrong on both counts — two sections are now optional, and sections WARN rather than block. Split into a required table (WARN, never blocking) and a recommended-but-unenforced table.

**3. ADR-0023's stale-README prose.** Confirmed at line 187. Recast: the row now reads as fixed-in-this-commit and enumerates all three surfaces corrected, including the two you found.

**4. Missing from `## Categories`.** Confirmed — added under *Architecture Evolution*, matching `domain: architecture`.

**5. The user-facing warning string.** Confirmed. Was:

> `required `## ` section missing — the ADR/PRP template declares it; add it or run the matching validate skill`

Now: `` required `## ` section missing — add it (Context, Decision and Consequences are the required set) ``. No pointer to enforcers that don't exist.

**6. Dangling `test-schema-field-parity.sh`.** Confirmed in **three** files, not two — `adr.schema.json:5`, `validate-frontmatter.sh:17`, and `check-schema.py:26`. All now name the real guard: `test-check-schema.sh`'s `no field lists:` assertions. `grep -rn test-schema-field-parity` is clean. You were right that this belonged in the diff, not in a "not fixed here" note — it's a one-line comment in the file whose sole-enforcement-point role this change depends on.

## Suggestions

**Fixture fidelity — this was the best catch in the review.** You're right that a no-op `sed` exits 0, and the consequence was worse than cosmetic: a renamed heading produced a fixture *identical to the input*, so the assertion would have passed against a validator checking nothing.

Replaced with `drop_section`, which deletes a heading and its body up to the next `## ` — removing the `+1d` body-length guess in the same change — and fails loudly when the heading is absent. Added section-count guards on all three fixtures, including `good.md` itself, since every other fixture derives from it.

Mutation-verified rather than assumed: renaming one heading to `ConsequencesTYPO` takes the suite from `PASSED=27 FAILED=0` to `PASSED=6 FAILED=22`. Under the old `+1d` form that typo was silent.

**`has_issue` regex-by-default.** Confirmed all five call sites pass literals (`SCHEMA_APPLICABLE=true`, `STATUS=OK`, `frontmatter_missing`, `superseded-by`, `superseded_by`). Now `grep -qF`, matching its documented contract.

**Assertion-count drift.** Confirmed at `regression-testing.md:314`. It is now 27, not the 24 your review predicted — the fixture guards added three more.

**Dormant shellspec specs.** Agreed, and left alone: pre-existing, and shellspec isn't installed here so I can't verify a change to them. Worth its own issue — say the word and I'll file it.

## Verification

| Check | Result |
|---|---|
| `test-check-schema.sh` | `PASSED=27 FAILED=0` |
| ADR corpus section WARNs | 5 (unchanged) |
| ADR-0023 validates | `STATUS=OK ISSUE_COUNT=0` |
| `check-docs-index.sh` | `STATUS=OK ISSUE_COUNT=0` |
| `pre-commit` on all 9 changed files | exit 0 |
| `x-blueprint-severity` | still `warning`, untouched |

One note on the review's framing: I've kept your "Risk: MEDIUM" verdict intact rather than arguing it down. The schema change was correct, but five of six blocking findings were inaccurate claims *about* enforcement, which is exactly the failure mode the original issue warned about — describing a standard that nothing upholds.

---

laurigates:
Both follow-ups from the review are now filed.

**#2702 — build the ceiling.** The section check in `blueprint-adr-validate`, or a real five-section template, or an explicit decision that the convention is enough. Carries the caveat that a check flagging all 20 historical ADRs would reproduce exactly the unreadable-report problem #2446 was filed to fix, so scope (new-ADRs-only vs. corpus-wide) has to be decided rather than defaulted.

**#2703 — the dormant shellspec specs.** Your review called these "pre-existing, but they now contradict the WARN-only design". Verified, and it is worse than that: they describe the validator #2440 deleted.

Measured every assertion in `validate_adr_frontmatter_spec.sh` against the live hook:

| Spec assertion | Actual | |
|---|---|---|
| missing `id` → exit 2, `ERROR: Missing required frontmatter field: id` | exit 2, `SEVERITY=ERROR … 'id' is a required property` | message ❌ |
| bad id format → `ERROR: Invalid ADR id format` | schema-violation line | message ❌ |
| bad status → `ERROR: Invalid ADR status` | `'Pending' is not one of [...]` | message ❌ |
| missing `## Decision` → exit 2 | **exit 0, `STATUS=WARN`** | both ❌ |
| missing `domain` → `WARNING: Missing 'domain' field` | **`STATUS=OK`, no warning** | message ❌ |

Five of five message assertions wrong, two of five exit statuses wrong. `validate_prp_frontmatter_spec.sh:90` has the same missing-section shape and `prp.schema.json` is also `severity: warning`, so it is wrong identically.

And `shellspec` appears nowhere in `.github/workflows/`, the `justfile`, or `.pre-commit-config.yaml` — these four files have never run, which is why nothing ever reported it. A suite that cannot fail cannot regress; its real cost is that the next reader believes the hooks are tested.

#2703 is labelled `question` because delete-vs-revive is a genuine fork. I lean delete — `test-check-schema.sh` covers the same ground with 27 mutation-verified assertions that actually run in pre-commit — but reviving them would recover hook-contract coverage (stdin → stderr → exit code end to end) that `test-check-schema.sh` does not have. I did not delete them unilaterally.

Neither is folded into this PR: both are separate changes, and `shellspec` can't be installed in this sandbox (the GitHub release archive 403s through the proxy), so a rewrite here would have been unverifiable — which is how the specs drifted in the first place.

This PR is unchanged and green: `e1d4c98`, 8/8 checks, `mergeable=true/clean`, no open threads.

---

Issue #2172, opened 2026-07-27
feat(plugins): ship batch-only workflow templates for four skills (Tier B)


laurigates:
Tier B — four skills whose **modal invocation stays prose**, gaining a template only for the large-input path. Filed together because they share one decision: *narrow the harness to the batch case, leave the single-item path alone.*

Source: [`docs/plans/dynamic-workflow-migration.md`](../blob/main/docs/plans/dynamic-workflow-migration.md) rows 6–9 (merged in #2162).

Ship these **after** the Tier A harnesses — they are the weaker half, and one of them (`cold-read-gate`) is explicitly optional.

## 1. `git-plugin/git-pr-feedback` — `--all` path only

Shape: fan-out-and-synthesize. Fan-out unit: **one worktree agent per actionable PR**.

Step 6 already specifies a full JSON return contract; the harness makes it enforced rather than requested. The single-PR path — the modal case — stays prose.

Must carry both worktree clauses: the `#1868` resume warning, and **push/PR-creation only in a single sequential finalise stage**, never inside a fanned-out agent.

## 2. `evaluate-plugin/evaluate-skill` — the only one needing global registration

Shape: fan-out-and-synthesize. Fan-out unit: **one rollout agent per (case × run × config) cell**.

The cartesian product `len(evals) × runs × configs` belongs in a script, not in prose the model expands by hand.

**Distribution is `both`** — this is the sole exception to the bundled-only rule. `evaluate-plugin-batch` (#2169) calls `workflow('evaluate-skill', …)`, and one-level nesting needs a resolvable name, so it must also register in `~/.claude/workflows`. **Register this one and nothing else.**

Ship with the `cellCap: 30` guard the batch harness passes in.

## 3. `git-plugin/git-issue` — `--parallel` path only

Shape: composite. Fan-out unit: **one worktree agent per conflict-free group**.

Grouping → barrier → sequential finalise is a real structure. The single-issue path is the modal case and stays linear.

Same two worktree clauses as `git-pr-feedback`.

## 4. `agent-patterns-plugin/cold-read-gate` — batch only, **optional**

Shape: adversarial-verification. Fan-out unit: **one haiku reader per outward-bound artifact**.

**This is the weakest candidate in the set.** Its original motivation (#2063) is already spent — `SKILL.md:52` pins `run_in_background: false`. Ship this **only if a real N≥3 batch path is actually wanted**; otherwise close this quarter of the issue as not-planned.

[...]

Issue #2169, opened 2026-07-27
feat(evaluate-plugin): ship evaluate-plugin-batch fan-out harness template


laurigates:
Tier A #3 — the textbook fan-out-and-synthesize case, but **it must ship capped**.

Source: [`docs/plans/dynamic-workflow-migration.md`](../blob/main/docs/plans/dynamic-workflow-migration.md) §3 (merged in #2162).

## Why it wins

`parallel()` replaces a hand-rolled `--parallel N`, and `aggregate_benchmark.sh` is a genuine synthesis barrier — the plugin report needs every skill's result at once. Fan-out unit: one nested `evaluate-skill` run per `skills/*/SKILL.md` that has evals.

## Why the caps are non-negotiable

Without them this is a one-command path to **~480 agents on a 48-skill plugin**, against a repo that currently has **one `evals.json`** and a `skill-evaluation.md` Tier-2 cadence of "monthly + on model release."

## Sketch

```js
phase('Discover skills');
const inv = await agent(
  `Inventory ${args.plugin}: bash evaluate-plugin/scripts/inspect_eval.sh --plugin-dir ${args.plugin}.
   Report {path,dir,name,hasEvals} per skills/*/SKILL.md. Evaluate NOTHING.`,
  { label:'inventory', schema: InventorySchema, model:'opus', effort:'low' });

const included = inv.skills.filter(s => s.hasEvals || args.createMissingEvals);
const CAP = args.cap ?? 25;                      // skill-evaluation.md golden set — refuse above it
if (included.length > CAP) return { abort:true, reason:`${included.length} skills exceeds the golden-set cap ${CAP}` };
const WAVE = args.parallel ?? 1;                 // honour --parallel N (default 1); never supersede it
log(`${inv.skills.length} skills: ${included.length} included, ${inv.skills.length - included.length} skipped`);

phase(`Evaluate ${included.length} skills`);
const results = [];
for (let i = 0; i < included.length; i += WAVE) {
  const wave = included.slice(i, i + WAVE);
  results.push(...await parallel(wave.map(s => () => workflow('evaluate-skill', {
    skill: s.path, skillDir: s.dir, runs: args.runs ?? 1,
    baseline: args.baseline ?? false, createEvals: !s.hasEvals, cellCap: 30 }))));
}

phase('Aggregate plugin report');   // SYNTHESIS BARRIER
return await agent(
  `bash evaluate-plugin/scripts/aggregate_benchmark.sh ${args.plugin}
[...]

Issue #2144, opened 2026-07-27
test(evaluate-plugin): ablate the context-engineering findings — turn judgments into measurements


laurigates:
Follow-up to #2139 (context-engineering benchmark). This is the rigor step the benchmark deliberately deferred, and the one that would make its recommendations falsifiable.

## Why

Anthropic's headline claim is an **ablation result**: they removed >80% of Claude Code's system prompt *"with no measurable loss on our coding evaluations"*. Cut, measure, keep the cut.

Ours are **judgments**. Two agents read a rubric and scored an artifact. That is a real signal — 150/150 evidence quotes verified, 1 contested pair of 55 — but it is not the same class of evidence, and #2139 §6 says so plainly:

> **Nothing is ablated.** No finding has been tested by cutting the content and re-measuring.

Every finding in the report is labelled **proven** (a judged, quoted finding) or **proxy-only** (Channel M alone) precisely so this issue has a queue to work.

## The blocker

**1 of 407 skills has an `evals.json`.** The Tier-1/Tier-2 machinery in `evaluate-plugin` exists and works — `grade_deterministic.py`, `/evaluate:matrix`, `render_matrix_report.py`, the baseline-delta 2×2 in `.claude/rules/skill-evaluation.md` — but it has essentially no coverage. That is why #2139 stopped at judgment.

## Shape

1. **Author `evals.json` for ~8 top-ranked units.** Use typed checks (`regex` / `substring` / `absent_regex`) per `evaluate-plugin/references/schemas.md` so most assertions grade for zero judge tokens. The golden set doubles as the ablation set.
2. **Baseline first.** Run `/evaluate:matrix` *before* any cut to get the with-skill − baseline delta per model. Cache the baseline per model version.
3. **Cut, then re-run.** Apply the report's proposed cut on a branch, re-run, compare deltas.
4. **Read the 2×2 honestly** (`skill-evaluation.md` Principle 2). A cut that leaves the delta intact is **proven**. A cut that collapses it was a proxy artifact — revert it and mark the finding refuted in the report.
5. **Promote to a standing gate only if the deltas prove informative.** If they don't discriminate, say so and leave Channel M + Channel J as the durable answer rather than shipping a gate that measures nothing.

`/evaluate:improve --apply --best-of 3` already ranks candidate revisions by re-running evals, so it can host the cut-vs-keep decision rather than needing new machinery.

## Sequencing

[...]

PR #2705, opened 2026-09-20
feat(plugins): ship the last four dynamic-workflow harness templates (#2169, #2172)

laurigates:
Closes the two remaining children of epic #2174. With these four templates the dynamic-workflow migration's shippable set is complete: the epic nominated five Tier A harnesses plus a Tier B batch of four, four Tier A landed earlier, and this PR lands the fifth plus the three Tier B items that were decided in.

Implemented by a 5-agent workflow: one survey agent derived the convention contract from the four already-shipped templates, three implementers worked disjoint file sets in parallel, and a fresh verifier that did not write any of it re-ran every gate against the tree (`loop-integrity.md` Pillar 1). Its three findings are fixed in the branch.

## What lands

| Commit | Scope | Issue |
|---|---|---|
| `docs(rules)` | narrow the `context: fork` hazard to *unbounded* fan-out width | #2172 |
| `feat(evaluate-plugin)` | `evaluate-skill` batch template + the one global registration | #2172 item 2 |
| `feat(evaluate-plugin)` | `evaluate-plugin-batch` capped fan-out | #2169 |
| `feat(git-plugin)` | `git-pr-feedback --all` | #2172 item 1 |
| `feat(git-plugin)` | `git-issue --parallel` | #2172 item 3 |

## The `context: fork` resolution

Per the decision recorded on #2172, the rule is **narrowed, not exempted by name**: the `[1m]` cascade bites a width nobody can name, so a fan-out bounded before dispatch (a fixed agent-type table, a cap constant, a script-enumerated set) keeps `fork`, while a caller-chosen width (`--parallel N`, one agent per query hit) keeps it off.

This turned out to be a **correction as much as an enabling edit**. `testing-plugin/test-analyze` has been CI-pinned to `context: fork` *and* shipping an 8-wide `parallel()` harness since the migration, and the rule still classified it as a single-subagent skill with no carve-out covering it.

The rule, the guard commentary, and the `workflow-vs-skill.md` corollary move in one commit. **No skill's frontmatter changed and the pin list is unchanged** — the `plugin-compliance-check.sh` diff is comments only. Three stale citations of that guard (all three had drifted ~125 lines) now name the `for fork_skill in` loop instead.

## Acceptance

**#2169** — `CAP` (25) aborts with the count and reason rather than truncating; `--parallel N` is honoured verbatim, default 1; `aggregate_benchmark.sh` owns the denominators and disagreement emits `partial-sweep`; no worktree stage. The three deletions from the original design (per-skill worktree normalize, uncapped `pipeline()`, workflow-count-as-authority) stay deleted.

**#2172** — each template at `workflows/<purpose>.workflow.js` beside its SKILL.md, each SKILL.md gaining a `## Workflow harness (template)` section with the *not a script to run verbatim* framing and an explicit **skip the harness when** naming its modal single-item path; both worktree templates carry the #1868 no-resume-to-retry clause and the sequential-finalise clause; `cold-read-gate` ships nothing, per the 2026-09-20 not-planned decision.

## Two deviations, both deliberate

1. **One PR, not three.** The #2172 decision asked for separate PRs per skill. This session is pinned to a single branch, so they land as separate, independently reviewable commits instead.
2. **The `~/.claude/workflows` registration checkbox is documented, not ticked.** Registration is an install-time act against a home directory outside the repo — it cannot be committed and no CI lint can assert it. `docs/dynamic-workflow-registration.md` records the install snippet, the by-hand verification (including a `diff` against the bundled source), and the gate for ever adding a second name. The sub-clause *"it is the only one so registered"* holds: no other bundled template declares a name another harness calls.

`evaluate-plugin-batch` runs near-empty until a golden set of evals exists (#2144). That was accepted in the issue, and the framing says so, so an empty sweep is not mistaken for a bug.

## Verification

All run against the final tree, not the implementers' reports: `check-workflow-js-model.sh` (27 agent calls, every one pinning model + effort), `plugin-compliance-check.sh`, `check-delegation-reachability.sh` (429 files), `check-subagent-types.sh`, `check-skill-references.sh`, `check-docs-index.sh`, `check-loop-integrity.sh`, `lint-shell-scripts.sh` — all pass. Full `pre-commit run --from-ref origin/main --to-ref HEAD`: no hook failed. Every new template parses as an async workflow body, every script and flag its prompts name exists on disk, and a bare invocation of each reaches a named abort rather than throwing.

https://claude.ai/code/session_014hHT67GHWX6G51LsykWxs2

---
_Generated by [Claude Code](https://claude.ai/code/session_014hHT67GHWX6G51LsykWxs2)_


laurigates:
## Compliance review: one item is this PR's, the rest are pre-existing

Every row is ⚠️, no ❌ — `plugin-compliance-check.sh` exits 0 on this head. Of the flagged items, exactly one was crossed *by this PR*:

| Skill | before | after | note |
|---|---|---|---|
| `evaluate-plugin/evaluate-skill` | 7,322 | 12,285 | crossed the 10k advisory band here |
| `evaluate-plugin/evaluate-plugin-batch` | 3,984 | 7,713 | still under |
| `git-plugin/git-issue` | 13,821 | 16,922 → split by the bot, below |
| `git-plugin/git-pr-feedback` | 21,168 | 23,977 | was already over before this PR |

**Not extracting `evaluate-skill` to `REFERENCE.md`.** The 4,963 chars added are the `## Workflow harness (template)` section, which is the epic's own requirement — per `.claude/rules/docs-currency.md` and the migration plan, a `.js` without its framing beside it is exactly the *"script to run verbatim"* the plan warns against, so moving the framing one file away would defeat what it is there for. The resulting size also sits squarely inside the shipped harness cohort — `test-analyze` 11,859, `verify-before-filing` 12,037, `blueprint-story-audit` 14,208 — and is under half the 26,000 ERROR ceiling. An identically-diverging cohort signals intent rather than drift (`.claude/rules/generated-fleet-drift.md`).

**`git-issue` was handled automatically.** The Skill splitter workflow pushed `290a24f` to this branch mid-review, extracting `REFERENCE.md` (13,821 → 16,922 → split). I verified the split rather than assuming it: the `## Workflow harness (template)` section stayed in `SKILL.md` with all four required clauses (the *not a script to run verbatim* framing, **Skip the harness when**, the #1868 no-resume-to-retry clause, the sequential-finalise clause), the two extracted sections left pointer stubs, and `workflows/issue-group-wave.workflow.js` is byte-identical. All gates re-run green on that head.

**One to watch, not this PR's to fix:** `git-pr-feedback` is at 23,977 of the 26,000 hard ceiling. It was already over the advisory band on `main` (21,168); the framing section pushed it closer. The next edit to that skill will likely need the same `REFERENCE.md` extraction the splitter just applied to `git-issue`.

The remaining ⚠️ rows — five over-length `git-plugin` descriptions, six missing `## When to Use This Skill` sections, and the other oversized `git-plugin` skills — are all pre-existing on `main` and untouched here.

---

